PR #2, opened 2018-12-17
Table Component

ahuth:
screen shot 2018-12-17 at 11 48 54 am


src/components/table/table.spec.jsx
```diff
@@ -0,0 +1,29 @@
...
+          <TableCell>Bob Marley</TableCell>
+        </TableRow>
+      </TableBody>
+    </Table>,
+  );
+}
+
+test('rendering', () => {
+  const wrapper = render();
+  expect(wrapper).toMatchSnapshot();
```
ahuth: Went with the snapshot test here, because this produces a lot of DOM and I didn't want to type out a ton of assertions for each thing. The snapshot test gets us comprehensive tests for all of that, with the downside that it's kind of "out of sight, out of mind", and the developer intent isn't super clear from the test itself.

src/components/table/table.story.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { Table, TableHead, TableBody, TableRow, TableCell } from './index';
+
+storiesOf('Table', module)
+  .add('Default', () => (
+    <Table>
+      <TableHead>
+        <TableCell>id</TableCell>
+        <TableCell>name</TableCell>
+        <TableCell>instrument</TableCell>
+      </TableHead>
```
ahuth: Chose to not require a `TableRow` wrapper around the table cells here in the header (it adds it internally). The downside of that is that with this current API we can't really use ColGroups.

I don't think we've done that a lot (if at all) in Mavenlink, and I judged this API to be okay for now.

src/components/table/table-cell.jsx
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+import styles from './table.css';
+
+export default function TableCell({ children, className, header }) {
+  const Element = header ? 'th' : 'td';
+
+  return (
+    <Element className={className}>
+      {children}
+    </Element>
```
ahuth: Note that I made the decision to re-use `TableCell` for both `td` and `th`.

There's certainly an argument to made **_against_** that.

src/components/table/table-head.jsx
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+import styles from './table.css';
+
+export default function TableHead({ children, className, cssRow }) {
+  return (
+    <thead className={className}>
+      <tr className={cssRow}>
+        {React.Children.map(children, (child) => {
+          return React.cloneElement(child, { header: true });
+        })}
```
ahuth: Using `cloneElement` to apply the `header` prop telling TableCell to make it a th instead of a td.

Alternatively, we could have different components for normal cells vs header cells.

.storybook/config.js
```diff
@@ -10,7 +10,7 @@ addDecorator(withOptions({
 }));
 
 function loadStories() {
-  const req = require.context('../src/components', true, /\-story\.jsx$/);
+  const req = require.context('../src/components', true, /\.story\.jsx$/);
```
roblevintennis: So they prefer story as more idiomatic then spec in storybook?

ahuth: We actually have both stories and specs. I changed this to a dot instead of hyphen here, to be consistent between the two. For example:

- input.spec.js
- input.story.js

Instead of one having a dot and one having a hyphen

src/components/table/table-cell.jsx
roblevintennis: Yeah, you could keep `TableCell` and allow it to take TH || TD and then provide those as decorators e.g. you'd also have a `TableHeader` and a `TableColumn`

ahuth:
Thanks Rob!

PR #6, opened 2019-01-07
Add Cypress for integration testing

ahuth:


.circleci/config.yml
```diff
@@ -16,19 +16,27 @@ jobs:
...
+          - v2-dependencies-
 
-      - run: yarn install
+      - run: yarn install --frozen-lockfile
 
       - save_cache:
           paths:
-            - node_modules
-          key: v1-dependencies-{{ checksum "package.json" }}
+            - ~/.cache
```
ahuth: The [Cypress Docs](https://docs.cypress.io/guides/guides/continuous-integration.html#Caching) state that we should be caching `~/.cache` instead of node_modules to ensure that the right version of Cypress is installed. It also indicates that the JS dependencies should still be cached correctly.

However, I'm not seeing that on our circle builds, and it seems like we're still re-downloading everything now. May need your help with this when you're back, @juanca

package.json
```diff
@@ -9,6 +9,8 @@
   "scripts": {
     "build": "styleguidist build",
     "cleanup": "rm -rf build",
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run",
```
ahuth: Perhaps this should've been:

- cypress:browser
- cypress:headless

PR #7, opened 2019-01-10
Overview

ahuth:
Add the Overview section (without hero, for now). Part of https://www.pivotaltracker.com/story/show/163068735.

overview

FYI @roblevintennis


styleguide/content/overview/overview.md
```diff
@@ -0,0 +1,19 @@
+<div class="prologue-para prologue-para-library">
+  <span class="prologue-title">Make useful things</span>
+  <span class="prologue-separator"></span>
+  Our components and styles are pre-architected and rigorously tested, allowing you to focus your efforts on creating purpose-built solutions and workflows for your users. Spend your time solving the hard problems, not building the pieces.
+</div>
+<img alt="" class="prologue-svg prologue-svg-library" src="svgs/library.svg" />
+<div class="clearall"></div>
+<img alt="" class="prologue-svg prologue-svg-reuse" src="svgs/reuse.svg" />
```
ahuth: @roblevintennis , let me know what you think of using an `` tag here, instead of using background-image in the stylesheet.

Doing it this way resolves an issue I was having with the SVGs working locally (webpack dev server), but not in production (static build).

roblevintennis: I don't have any issue with an image vs. background url. I'm pretty sure we'll have to adjust the animation code in terms of the selectors, but I presume that won't be too hard to figure out.

ahuth: I'm going to hold off on the hero. Want to investigate if there's easier ways of showing those things

PR #8, opened 2019-01-10
First stab at adding the google analytics stuff. code needs to be upd…

roblevintennis:
needs the code updated. I basically monkey see'd https://github.com/mavenlink/mavenlink/commit/5ce4d79b6c34324aa0acd3a3ee47d29b03230a41 besides converting link.css from sass

https://www.pivotaltracker.com/story/show/163068777


.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "mavenlint-react",
   "env": {
+    "browser": true,
```
ahuth: I think we don't want to do this. Almost left a note about this, but we need our components to run on the server or the client, and `window` stuff in our components is a big no-no for that.

What we can do is have the `styleguide` directory have it's own eslint config that allows browser stuff

styleguide/content/utils/page-view.js
```diff
@@ -0,0 +1,11 @@
+import ReactGA from 'react-ga';
+
+ReactGA.initialize('UA-7529271-29');
```
ahuth: Per our discussion, I need to generate a new GA site/number.

Also, could perhaps look at setting it based off an ENV.

styleguide/components/styleguide/index.js
```diff
@@ -1,3 +1,4 @@
-import Styleguide from './styleguide';
+import Styleguide from './styleguide-container';
 
 export default Styleguide;
+
```
ahuth: Is this line intended?

roblevintennis: Not sure there

styleguide/components/link/link.css
```diff
@@ -0,0 +1,16 @@
+@import "../../../src/styles/colors.css";
+
+.link {
+  &:link,
```
ahuth: Convert to CSS:

```
.link:link, 
.link:visited {
  ...
}
```

styleguide/content/utils/page-view.js
ahuth: @roblevintennis got a new number for ya:

```
UA-7529271-31
```

ahuth:
```suggestion
ReactGA.initialize('UA-7529271-31');
```

PR #9, opened 2019-01-18
Working on some more section like guidelines and soul. Added some css…

roblevintennis:
Just adds 2 new sections. Dev QA'd it so going to merge

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -23,9 +23,7 @@ Pull requests are the best way to propose changes to the codebase (we use [Githu
 
 ## Running the app locally
 
-```
-yarn start
-```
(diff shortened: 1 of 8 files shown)


docs/contributing.md
```diff
@@ -23,9 +23,7 @@ Pull requests are the best way to propose changes to the codebase (we use [Githu
 
 ## Running the app locally
 
-```
-yarn start
-```
+```yarn start```
```
roblevintennis: When I actually pulled this in from the styleguidist aka design-system site, it complained until I got rid of the newlines. Your syntax is valid but whatever plugin they're using must not like it lol

PR #10, opened 2019-01-21
Mds overview section hero

roblevintennis:
Load heroes in to sections

Looking in to the pulsating stars stuff and, if we want to stick with using only postcss here, we'll probably have to loose those animations. I tried pulling in postcss-for, postcss-random, but then encountered that I'm using `hsl`, sass multiplication, relying on `&:nth-child`, etc. It's just probably not worth it and doesn't look _that much better_ imo cc / @juanca 

e.g. stuff like:

`$t: (1 + 0.01 * random(100)) * 25s;`

and

`&:nth-child($(i) + 1}) {`


roblevintennis:
My thought is that we should probably just create a static SVG with the stars and nuke `stars.jsx` et al

package.json
```diff
@@ -67,6 +67,9 @@
     "webpack": "^4.28.3"
   },
   "dependencies": {
-    "prop-types": "^15.6.2"
+    "file-loader": "^3.0.1",
+    "prop-types": "^15.6.2",
+    "react-transition-group": "^2.5.3",
```
juanca: We probably want to move these new dependencies as a `devDependencies`

styleguide/components/hero/animations.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CSSTransition from 'react-transition-group/CSSTransition';
+import astroStyles from './astronauts.css';
+import Stars from './stars';
+
+function getTransition(klass, appearKlass, appearActiveKlass) {
```
juanca: INFO: I would just avoid defining the variables below in favor of just using the parameters defined here.

styleguide/components/hero/animations.jsx
```diff
@@ -0,0 +1,55 @@
...
+      <div className={astroStyles[baseClassName]} />
+    </CSSTransition>
+  );
+}
+
+export default function Animations(props) {
+  return (
+    <div>
+      <Stars slug={props.slug} />
+      {props.slug === 'overview' && [
```
juanca: INFO: Since we're testing `props.slug` against three known values, I would normally use a `switch` statement for this:

```
{switch(props.slug) {
  case 'overview': [];
  case '...': [];
  case '...': [];
  default: ???;
}}
```

... assuming I have the syntax correct.

styleguide/components/hero/hero.css
```diff
@@ -0,0 +1,59 @@
...
+.hero {
+  position: absolute;
+  left: 200px;
+  top: 65px;
+  background-size: 100% 100%;
+  display: inline-block;
+  width: var(--heroWidth);
+  height: var(--heroHeight);
+
+  @media (min-width: var(--largeScreenBreakPoint)) {
```
juanca: Oh, are we suppose to un-nest this?

PR #11, opened 2019-01-24
Updated link to support passing in classname. Added in components-lis…

roblevintennis:
https://www.pivotaltracker.com/story/show/163408640

### DIFF
--- a/styleguide/components/components-list/components-list.jsx
+++ b/styleguide/components/components-list/components-list.jsx
@@ -1,0 +1,52 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import Link from '../link';
+import styles from './components-list.css';
+
+export default function ComponentsList({ current, items }) {
+  const filteredItems = items.filter(item => item.name);
+
+  if (!filteredItems.length) {
+    return null;
+  }
+
+  return (
+    <ul className={styles.list}>
+      {filteredItems.map(({ name, href, content }) => {
+        if (current === name) {
+          console.log(`current: ${current}`);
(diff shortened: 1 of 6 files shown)


styleguide/components/components-list/components-list.jsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <ul className={styles.list}>
+      {filteredItems.map(({ name, href, content }) => {
+        if (current === name) {
+          console.log(`current: ${current}`);
+          console.log(`content: ${content}`);
+          console.log(`href: ${href}`);
+          console.log(`name: ${name}`);
```
ahuth: Hey @roblevintennis @juanca these console logs caused the build to fail on lint checks. Do we still want these in here?

Also, FYI, one thing I like doing when debugging is grouping all these logs into one thing:

```js
console.table({ current, content, href, name });
```

Makes it pretty print real nicely.

juanca: Good call. I'll set up another PR to fix this. It also seems like we need to tighten the requirements for merging into `master`. 👀

ahuth: Thanks. Also looks like I never set up a GitHub integration with CI 😬 

Could make it easier to spot these issues. I can see if I have access to do that

juanca: I believe only admin members of the owning organization can change those settings. Done and done @ https://github.com/mavenlink/design-system/pull/13

PR #12, opened 2019-01-30
Mds icons new

roblevintennis:
Moving some icon stuff over here.

This belongs with bigmaven and mavenlink-js PRs:
https://github.com/mavenlink/mavenlink-js/pull/551
https://github.com/mavenlink/mavenlink/pull/15147

### DIFF
--- a/src/styles/icons-vars.css
+++ b/src/styles/icons-vars.css
@@ -1,0 +1,13 @@
+:root {
+  --iconSize: 18px;
+  --iconSizeMicro: 9px;
+  --iconSizeMiniscule: 11px;
+  --iconSizeTiny: 14px;
+  --iconSizeSmall: 16px;
+  --iconSizeLarge: 20px;
+  --iconSizeJumbo: 28px;
+  --iconSizeBig: 32px;
+  --iconSizeHuge: 48px;
+  --iconSizeGigantic: 64px;
+  --statusLightGreen: #A4DD74;
+}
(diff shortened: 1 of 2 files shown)


juanca:
We should figure out why this is not running on CI.

roblevintennis:
Yeah, I think we could 1. get them in 2. post mortem with Design 3. make tweaks if needed.
That said, just to be crystal clear, we should _NOT_ expose `src/styles/icons-vars.css` usage, just the decorator classes made available at the interface level. I think we agree there, just being explicit. Let's chat at standup or in person and make another pivotal for that.

juanca:
Hmmm... So it would be hard to make a stylesheet private in this repository. My first thoughts are just to ensure everything is readily available to be consumed publicly. Otherwise, we can perhaps start a pattern of prefixing our private stuff with `__unsafe__` to denote that we could break the API without warning.

juanca:
Actually, it looks like we are only using icon small and large from the variables file. Do we want to use the rest of the sizes in this repository? 

I'm actually okay with exposing the icon sizes (and we should probably cut down on the list) since others can ensure their icons are standardized with our icons (task / milestone / issue / etc). 

Though, nothing wrong with keeping them private for now until we have a clearer picture in mind.

roblevintennis:
Closing this. Gonna port all this stuff back from design system (here) back over and have it all reside in `src/stylesheets/styleguide/components/icons/icons-new.css` in mavenlink-js so we don't have to deal with the sass hoisting problem. We'll investigate some of this again after we've converted from sass all the way over to postcss (see https://github.com/mavenlink/mavenlink/pull/15327 for spike on that)

PR #13, opened 2019-02-06
PR requirements for merging

juanca:
Adding a few things through this PR:

- Green CI suite
- PR review
- Up-to-date `master` commits
- Speed up `yarn install` step via proper caching


juanca:
We still need to figure out the reviewers pool. 

I will bypass that check via an admin override. But we are definitely seeing all the checks required above. 💯

PR #17, opened 2019-03-01
Self-contained space values

juanca:
Ensures we are using our design language in our components. In addition, it ensures the components have deterministic width calculations -- especially now that we are standardizing padding / margin values.

### DIFF
--- a/src/components/input/input.css
+++ b/src/components/input/input.css
@@ -1,4 +1,5 @@
 @import "../../styles/colors.css";
+@import "../../styles/spacing.css";
 @import "../../styles/typography.css";
 
 .input {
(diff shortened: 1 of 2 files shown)


juanca:
I just realized we cannot QA this since we cannot deploy this without clobbering our deploy from `master`. 

Hmmm..

juanca:
I ended up running through a pair QA exercise.

A couple of follow ups: 

- We need the Spacing Soul section in this repository. However, we are collecting feedback for the time being and will port once it is in a better shape. 
- We need a section for QA criteria -- for example: what does it mean to have consistent spacing? 
- We need to have some sort of "instastage" for this repository to facilitate easier peer review (and an PR check since it _needs_ to compile and deploy)

PR #20, opened 2019-04-23
Jest Snapshots

roblevintennis:
Jest FTW!

* Adds jest snapshot spec
* maintains jest as test runner for our enzyme based unit specs that already existed for our components
* Configured to ignore the `styleguide/cypress/integration` which is a cypress intergration spec


src/components/input/input.spec.jsx
```diff
@@ -6,11 +6,6 @@ function render(props) {
   return shallow(<Input {...props} id="hi" />);
 }
 
-test('rendering', () => {
-  const wrapper = render();
-  expect(wrapper.find('input').prop('id')).toEqual('hi');
-});
-
 test('passing props through', () => {
```
roblevintennis: This is actually _also_ covered by the snapshot if you think about it. Ima leave this so we can see what we had before...but after, I'd like to just completely delete this spec if y'all agree!

package.json
```diff
@@ -16,7 +16,8 @@
     "postdeploy": "yarn run cleanup",
     "predeploy": "yarn run build",
     "start": "styleguidist server",
-    "test": "jest"
+    "test": "node node_modules/.bin/jest",
```
juanca: Why do we need this change?

package.json
```diff
@@ -16,7 +16,8 @@
...
+    "test:update": "node node_modules/.bin/jest --updateSnapshot"
```
juanca: Why do we need to specify the full path to the jest script file?

roblevintennis:
Updated to just `jest`

Issue #21, opened 2019-04-24
Deploy every branch


juanca:
I have done something like this in https://github.com/juanca/react-aria-components/pull/56

It would enable us to deploy every branch to Github Pages.

http://mavenlink.github.io/design-system/master
http://mavenlink.github.io/design-system/some-banch

This would be great for QA purposes and iterating a bit faster.

PR #23, opened 2019-05-02
Instastage for each branch

juanca:
I will be replicating the efforts in one of my personal projects: https://github.com/juanca/react-aria-components/pull/56

If we want to increase the amount of efforts towards improving the design-system then we will need to improve our staging workflows. 

- This allows for Github pages to work for every branch -- or PR, depending on CI settings.
- This allows QA to test changes for every PR!
- This allows Developers to preview changes on a "production-like" build instead of the local development environment

This branch is deployed to: https://mavenlink.github.io/design-system/instastage-pages/
Master branch is deployed to: https://mavenlink.github.io/design-system/master/

Related issue: https://github.com/mavenlink/design-system/issues/21

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,8 @@ workflows:
       - deploy:
           filters:
             branches:
-              only:
-                - master
+              ignore:
+                - gh-pages
           requires:
             - build_and_test
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,4 +68,15 @@ jobs:
       - add_ssh_keys:
           fingerprints:
             - "79:26:19:57:4e:1a:59:64:ec:bb:b7:76:e2:5f:7c:ad"
-      - run: yarn deploy
+      - run:
+          name: Compile and deploy Github pages
+          command: |
+            yarn build
+            git checkout gh-pages
+            rm -rf $CIRCLE_BRANCH
+            mv build $CIRCLE_BRANCH
+            git add .
+            git diff-index --quiet HEAD || git commit -m 'Update Github pages for $CIRCLE_BRANCH'
+            git push origin gh-pages
+      - run: echo https://mavenlink.github.io/design-system/$CIRCLE_BRANCH/
+      - run: ls $CIRCLE_BRANCH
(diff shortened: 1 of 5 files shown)


juanca:
Looks like I'm missing a little bit of improvements to our `.gitignore`. It looks like we are committing the `node_modules` -- which we need to remove since our static assets do not need them. 🤦‍♂

.circleci/config.yml
```diff
@@ -68,4 +68,15 @@ jobs:
...
+      - run:
+          name: Compile and deploy Github pages
+          command: |
+            yarn build
+            git checkout gh-pages
+            rm -rf $CIRCLE_BRANCH
+            mv build $CIRCLE_BRANCH
+            git add .
+            git diff-index --quiet HEAD || git commit -m 'Update Github pages for $CIRCLE_BRANCH'
+            git push origin gh-pages
```
roblevintennis: I'd like to see our main site preserved with a branch we deem "stable"

image

juanca: Ah, that is now in the master folder.

juanca: https://mavenlink.github.io/design-system/master/

Does that work for you?

roblevintennis: Yes, so I saw that, but I think maybe we just need to figure out how to alias the root site to `master` too. Is that going to prove difficult?

My reasoning:
* It's a very common thing for folks to manually go to top level of a site and that should not cause 404
* We link to it from our READMEs currently (obviously fixable, but I still think top level should work imo)

juanca: I think I am going to place an `index.html` with a link to the `master/` then. I would prefer to start the new process of always specifying the branch instead of an alias or redirect. We can gather some feedback later and make corrective actions then.

juanca: Done @ https://mavenlink.github.io/design-system/

roblevintennis: why not just do:
`window.location.replace("https://mavenlink.github.io/design-system/master/");`

juanca:
I think the landing page without the redirect has more potential to be useful. If we establish a redirect then it becomes a "surprise" when and if we decide to change it.

However, if we were to keep the index file, then others can just bookmark the `master/` folder.

Does that make sense?

roblevintennis:
Ok, I see your point. I think we should track a task with intent to make the landing page compelling—especially since it's public-facing. For now, let's move forward with this PR as is. Thanks for adding this @juanca ⚡️

PR #24, opened 2019-07-12
<Button />

juanca:
Deployment: https://mavenlink.github.io/design-system/button/index.html

Type of work: component

Deployment URL: https://mavenlink.github.io/design-system/button https://www.pivotaltracker.com/story/show/165361377

## Motivation

This PR focuses on introducing a button to our design system.

It will only introduce two kinds of buttons: submit a form and cancel a form. While we might have use-case for other kinds of button we do not have to over-engineer our MDS button. 

## Acceptance Criteria

- Button can be primary
- Button can be secondary
- Button looks good in our supported browsers
- Button has good enough API

## Change log entry

Implements `<Button />` component. At the moment, we only introduce two kinds of buttons as a means of minimizing the breadth of our supported components.


src/components/button/button.md
```diff
@@ -0,0 +1,11 @@
+```jsx
+<Button>Default Button</Button>
+```
+
+```jsx
+<Button color="primary">Primary Button</Button>
+```
+
+```jsx
+<Button color="secondary">Secondary Button</Button>
```
roblevintennis: Might I recommend that we consider a blank button too? I think we use that a lot and it might be a good one for public consumption. Really a product / design questions so consider this just a "let's think about it" ;)

juanca: I'm using the requirements of the Maven Page for this. 

So you're in luck because it *does* use a cancel button that looks like a link. I am actually going to remove secondary for now and only include Primary and, uh, Blank.

roblevintennis: Makes sense and I think we realized that during the preview the other day 👍

juanca:
Replaces:

Screen Shot 2019-07-15 at 3 02 26 PM

With:

Screen Shot 2019-07-15 at 3 03 34 PM
Screen Shot 2019-07-15 at 3 03 39 PM

src/components/button/button.jsx
```diff
@@ -25,6 +25,7 @@ export default function Button(props) {
 Button.propTypes = {
   className: PropTypes.string,
   color: PropTypes.oneOf([
+    'blank',
```
roblevintennis: Looks like we need to get rid of `secondary` below...friendly reminder as I'm assuming you're planning to do that.

roblevintennis:
in the screenshot from project create, can we ignore the icons and only consider the cancel and primary buttons and disabled state?

src/components/button/button.jsx
juanca: I decided against it. It's part of a commit @ https://github.com/mavenlink/design-system/pull/24/commits/de3f2daf627a048d277d0f635a760d7d468027b9 Essentially, it helps to extract the appropriate class names.

juanca:
What are your thoughts on that?

src/components/button/button.css
```diff
@@ -0,0 +1,67 @@
...
+  padding-right: var(--spacing-x-large);
+  padding-top: 0;
+}
+
+.button:hover {
+  opacity: .7;
+}
+
+.button[disabled] {
+  background: #eee;
```
juanca: Questions for design: are we okay with bypassing our palette?

src/components/button/button.css
```diff
@@ -0,0 +1,67 @@
...
+  text-shadow: 0 1px 1px var(--white);
+  color: #ccc;
+  cursor: default !important;
+}
+
+.button[disabled]:hover {
+  opacity: 1;
+}
+
+.blank {
```
juanca: Question for design: is there a concept for a disabled state on a blank button?

src/components/button/button.css
roblevintennis: This is leftover from unicorn-ui (I pretty sure). I agree, let's replace this with something more from Mavenlink. We have an #ededed; which should be virtually the same visually...just use whatever that palette is. I don't think design needs to corroborate but might not be bad idea

roblevintennis:
@juanca thanks, yes, since you asked for my thoughts...

I've been thinking about this a bit, and an issue I have is that we're essentially copying over (with some tweaks) a lot of our button code that already exists. I'd prefer to see whatever we deem core to buttons be centralized and shared. We could both leave our buttons in bigmaven to make iterating easy, but also add this, but point both to the more centralized shared core button code.

Put differently, I think we should not be adding to the code duplication problem by recreating things that already exist, but rather improve the core parts of those while removing the need for having a _Thing_ in several places. I have ideas on how we could do that we could discuss, but wanted to express the big picture idea.

src/components/button/button.css
juanca: Question for design: is there a "active" state for the blank button? legacy styleguide has some inset styling.

src/components/button/button.css
```diff
@@ -0,0 +1,67 @@
...
+  cursor: pointer;
+}
+
+.typography {
+  font-family: var(--mavenlink-type-font-family);
+  font-size: var(--mavenlink-type-button-font-size);
+  line-height: var(--spacing-x-large);
+}
+
+.button {
```
juanca: Question for design: is there a "active" state for the a normal button? legacy styleguide has some inset styling.

juanca: Question for design: are we ensuring our components have the appropriate `box-sizing: border-box`?
Question for design + engineering: what other "self-contained" styling do we need?

roblevintennis: Global resets like zeroing out margins. Also look for helpers e.g. we had a antialiasing thing and if other stuff it should live besides that. But from what we did in self contained components I'd say yes, we want to use `box-sizing`. To my mind, Design probably doesn't care as that's an implementation detail so long as the component is compliant with the spacing spec and any other similar design specs.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mavenlink/design-system",
-  "version": "0.7.0",
+  "version": "0.7.0-buttons-1",
```
juanca: Looks like we didn't need a pre-release version. The custom forms repository can pull in a github dependency if we use a non-SSH git protocol during development.

roblevintennis:
This might be interesting to compare albeit much larger surface API then we'll need no doubt, but perhaps to see if we've forgotten anything:
https://ant.design/components/button/#API

juanca:
I'll peruse it once we're a bit more complete on this buttons work. I want to make sure I am serving the needs of MavenPages and Mavenlink before thinking about the rest. Though, my intuition for MDS is to avoid being too flexible as a way to ensure we can keep maintenance costs low.

src/styles/colors.css
```diff
@@ -50,6 +51,13 @@
   --palette-neutral-x-light: #f5f5f2;
   --palette-neutral-xx-light: #fbfaf6;
 
+  /* New Neutral */
```
juanca: What are appropriate names for these?

src/styles/colors.css
```diff
@@ -50,6 +51,13 @@
...
+  --palette-03: rgba(0, 0, 0, 0.03);
```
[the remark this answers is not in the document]
juanca: Yes. We need to come up with a good API for these colors. It seems like we want to move forward with these.

src/components/button/button.jsx
```diff
@@ -0,0 +1,49 @@
...
+    'reset',
+    'submit',
+  ]),
+};
+
+Button.defaultProps = {
+  className: undefined,
+  color: 'primary',
+  disabled: undefined,
+  onClick: () => {},
```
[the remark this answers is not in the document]
juanca: Yeah, it will still work! We are doing that in the MavenPage!

src/components/button/button.css
```diff
@@ -0,0 +1,64 @@
+@import '../../styles/colors.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography.css';
+
+.interactivity {
+  cursor: pointer;
+}
+
+.typography {
+  font-family: var(--mavenlink-type-font-family);
```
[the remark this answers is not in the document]
juanca: Since we allow overrides via the CSS API, I think this is fine. 

We are also providing only 1 point of specificity so it can be easily cascaded away or `!important`.

src/components/button/button.test.jsx
```diff
@@ -0,0 +1,116 @@
...
+    it('can be "reset"', () => {
+      const tree = renderer.create((
+        <Button type="reset">
+          Hello world!
+        </Button>
+      )).toJSON();
+      expect(tree.props.type).toEqual('reset');
+    });
+
+    it('can be "submit"', () => {
```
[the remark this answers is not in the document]
juanca: I think this is unnecessary unless we somehow end up with a regression.

The only way to bypass the default behavior is by calling `event.preventDefault()` in the event handler. I think this would be more relevant if we were overriding the default handler every time.

package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "@mavenlink/design-system",
+<<<<<<< HEAD
```
roblevintennis: Might have been inadvertently left?

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
```
roblevintennis: Noice!

.stylelintrc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
roblevintennis: Also noice! 🙌

src/components/button/button.css
```diff
@@ -0,0 +1,64 @@
...
+  color: var(--palette-grey-base);
+}
+
+.secondary {
+  composes: button;
+  background-color: var(--white);
+  border-width: 1px;
+  border-style: solid;
+  border-color: var(--palette-grey-light);
+  color: var(--palette-grey-dark);
```
roblevintennis: yeah, the -grey- reads good to me!

src/styles/colors.css
```diff
@@ -35,12 +35,13 @@
...
   --palette-warning-base: #ffdf8d;
   --palette-warning-light: #ffedc0;
+
   /* Warning interactions */
   --palette-warning-hover: #ffd874;
 
   /* Neutral */
   --palette-neutral-xx-dark: #282624;
-  --palette-neutral-x-dark: #333333;
+  --palette-neutral-x-dark: #333;
```
roblevintennis: I'm not quite sure why we needed to do this but it works so ok ;)

juanca: It is part of the standard Stylelint rules from the "popular" ruleset. I figured we might as well keep the  set of rules until we feel really strongly against the rules.

juanca:
I removed the setting in our Github repository to auto-dismiss reviews after pushing new commits. I realized that we get a LGTM then we push a new commit with the release so it is futile. 

For posterity's sake, Rob already LGTM'd this. I am merging.

PR #32, opened 2019-07-17
More colors

juanca:
Introduce missing colors for our MDS colors.

Related bigmaven PR: https://github.com/mavenlink/mavenlink/pull/16906

- [ ] Design review
- [ ] Remove non-MDS colors and move back into bigmaven

### DIFF
--- a/src/styles/colors.css
+++ b/src/styles/colors.css
@@ -1,23 +1,23 @@
 :root {
   /* Action */
   --palette-action: #33ae10;
-  --palette-action-dark: #000;
-  --palette-action-light: #000;
+  --palette-action-dark: #257f0c;
+  --palette-action-light: #41dd14;
   --palette-action-hover: #328a15;
 
   /* Caution */
   --palette-caution: #e85342;
   --palette-caution-active: #000;
-  --palette-caution-base: #000;
-  --palette-caution-dark: #000;
+  --palette-caution-base: #e85342;
+  --palette-caution-dark: #dc2f1b;
   --palette-caution-hover: #e53e2b;
-  --palette-caution-light: #000;
+  --palette-caution-light: #ee7c6f;
 
   /* Highlight */
   --palette-highlight: #ff8400;
   --palette-highlight-hover: #ffa800;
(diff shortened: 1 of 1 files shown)


src/styles/colors.css
```diff
@@ -1,23 +1,75 @@
 :root {
-  --palette-primary: #448fea;
-  --palette-primary-hover: #447bc6;
+  /* Action */
   --palette-action: #33ae10;
```
juanca: Looks like a color without a modifier is the same as itself with `-base` postfix. 

We should simplify it and choose one or the other. I am voting for `base` since it means we keep some kinda of modifier on it.

src/styles/colors.css
```diff
@@ -1,23 +1,78 @@
...
+  --palette-action-dark: #257f0c;
+  --palette-action-light: #41dd14;
   --palette-action-hover: #328a15;
-  --palette-highlight: #ff8400;
-  --palette-highlight-hover: #ffa800;
+
+  /* Caution */
   --palette-caution: #e85342;
+  --palette-caution-active: #000;
+  --palette-caution-base: #e85342;
```
roblevintennis: Interesting. Do we have `-base` for all variations? In diff I don't see one for highlight for example. Not sure if I added these, but I don't see the advantage of not having just one. I don't care between `--name` or `--name-base` whatever is easier to backfill is my sense

src/styles/colors.css
```diff
@@ -1,23 +1,78 @@
...
   --white: #fff;
   --black: #000;
+
+  /* Warning */
+  --palette-warning: #ffdf8d;
+  --palette-warning-dark: #ffd15a;
+  --palette-warning-light: #ffedc0;
+  --palette-warning-hover: #ffd874;
+
+  /* Unknown -- possibly bogus */
```
roblevintennis: I'd search codebase for the hex values maybe, I recall these were sort of accent variations maybe used by marketing mostly

juanca:
We do not want all of these. This branch / PR was being used as a test on how to deprecate SASS in Bigmaven. These tests are concluded.

PR #35, opened 2019-07-18
Color re-organization

juanca:
Adds the missing colors from our palette. We are explicitly avoiding porting "product" colors since those are application specific. 

The colors will change in the near future as well when we collaborate with the design team.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mavenlink/design-system",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "description": "Mavenlink React Components",
   "main": "src/index.js",
   "files": [
--- a/src/styles/colors.css
+++ b/src/styles/colors.css
@@ -1,13 +1,24 @@
 :root {
+  /* Primary */
   --palette-primary: #448fea;
   --palette-primary-hover: #447bc6;
+
+  /* Action */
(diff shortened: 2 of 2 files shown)


juanca:
I am standardizing with `-base`. Deprecating anything without modifier.

roblevintennis:
@juanca where's the link back to pivotal for this?

Also, if you're going to call something like this dev qa'd when it's a globally used defs file like colors.css, please at minimum at least run `yarn styleguide` locally to ensure no components are affected. Thanks

Here's what I believe is not redirected to the `-base`:

https://github.com/mavenlink/mavenlink-js/blob/0a561896ccba3f680d814fef6043470bc23c560f/src/stylesheets/styleguide/components/icons/icons-new.css#L135

patch for regression coming soon

juanca:
Good point. I am not sure if we're linking pivotal stories in MDS PRs.. but we probably should? 

I updated the PR description with a link to the story for this instance. I am going to add a PR template for this repo.

roblevintennis:
> I am not sure if we're linking pivotal stories in MDS PRs.. but we probably should?

Hrm, good point. I suppose that could throw public onlookers huh?! Maybe we just do pivotal --> gh for design system I agree that would feel very exclusionary to outsiders.

PR #38, opened 2019-07-23
IE11 support on MDS site

juanca:
Type of work: site

Deployment URL: https://mavenlink.github.io/design-system/ie11-support

(Optional for outside contributions) Tracked work in Mavenlink: https://www.pivotaltracker.com/story/show/167441913

## Motivation

Includes `@babel/polyfill` as its own entry in MDS site. This ensures the site loads on IE11. We can now test components and accessibility in IE11!

## Acceptance Criteria

- [x] Site loads on IE11
- [x] Site loads on Edge
- [x] Site loads on Chrome
- [x] Site loads on Firefox
- [x] Site loads on Safari

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -17,6 +17,7 @@ module.exports = {
   },
   pagePerSection: true,
   require: [
+    '@babel/polyfill',
     './styleguide/content/content.css',
   ],
   sections: [
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "devDependencies": {
     "@babel/core": "^7.2.2",
     "@babel/plugin-proposal-class-properties": "^7.2.3",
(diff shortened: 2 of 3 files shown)


juanca:
Huzzah!

MDS-IE11

styleguide.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   },
   pagePerSection: true,
   require: [
+    '@babel/polyfill',
```
roblevintennis: Interesting, yeah, I think this makes sense .. we had this issue with the internal one for same reason until we did added a require in the styleguide config like:

`    './frontend/features/styleguide/polyfills.js',`

which imports the polyfills. Guess we can't reach that from here though, so this makes sense to me!

PR #42, opened 2019-07-24
Icon component in Design System

roblevintennis:
Type of work: component

Deployment URL: https://mavenlink.github.io/design-system/all-svgs-in-internal-design-system-partII

https://www.pivotaltracker.com/story/show/167474360

## Motivation

Maven Forms MVP requires an alert icon in the input for one of the fields, and we will need to compose in to that input field. Additionally, the <Icon /> component will have the API for making SVG icons.

## Acceptance Criteria

* It's documented in the MDS ✅
* We'll see the icon styled properly ✅
* Styles will be self contained and not require any external dependencies ✅
* Has proper test coverage ✅

MDS local rendered Icon component page:

image

__Note: In the spirit of starting with minimal requirements, we are not yet supporting neutral colors, black, white, and other more exotic colors available in bigmaven. We can always add those in future iterations in the spirit of being agile!__

## Change log entry

Adds an `<Icon />` component for rendering SVGs. Currently only adds the alert icon.


styleguide/webpack.config.js
```diff
@@ -22,7 +23,45 @@ module.exports = {
...
+                { removeTitle: true },
+                { removeNonInheritableGroupAttrs: true },
+                { removeUselessStrokeAndFill: false },
+                { collapseGroups: true },
+                { convertColors: { shorthex: false } },
+                { convertPathData: false },
+                {
+                  removeAttrs: {
+                    attrs: '(fill|stroke)',
+                    preserveCurrentColor: true,
```
roblevintennis: These are copied over from what we have in the main application, but without `preserveCurrentColor` for example, our `color: val` based SVGs won't work.

src/components/icon/icon.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+    describe('stroke', () => {
+      it('sets stroke colors turning off fills and color', () => {
+        const colors = ['primary', 'action', 'highlight', 'caution'];
+        colors.forEach((color) => {
+          const tree = renderer.create((
+            <Icon name={'foobar'} stroke={color} />
+          )).toJSON();
+          expect(tree.props.className).toContain('icon-base');
+          expect(tree.props.className).toContain(`stroke-${color}`);
+          expect(tree.props.className).toContain('fill-none');
```
roblevintennis: Fwiw, stroke based svg icons always turn off the fill and color by default, until you ask for those as well (covered by the combination describe later).

juanca:
Let's update the deployment URL in the PR description

juanca:
Let's update the pivotal story ID with a link to the story.

roblevintennis:
@juanca I believe I've made the requested updates .. please have another look

juanca:
For the "type of work" PR description, I think it is better to choose one and delete the rest.

At the moment you have:

> Type of work: bug / component / site / accessibility / other
> component

But I think it is meant to be:

> Type of work: component

Right? Otherwise, we are leaving a bunch of useless information on the description which makes it harder to read. 

It would be the same as leaving the default deployment URL _and_ appending the branch URL:

> Deployment URL: https://mavenlink.github.io/design-system/$BRANCH
> https://mavenlink.github.io/design-system/all-svgs-in-internal-design-system-partII

juanca:
Also, (my office internet is really flaky at the moment), but you never updated with a link to the pivotal story. Can you add that?

roblevintennis:
I left
> (Optional for outside contributions) Tracked work in Mavenlink:
> #167474360

which is the pivotal task number

juanca:
I understand that but for posterity's sake, let's switch it to a URL instead of just a number. This has been common feedback in our normal Mavenlink deployment process as well.

src/components/icon/icon.css
```diff
@@ -0,0 +1,85 @@
+@import "../../../src/styles/colors.css";
+
+:root {
```
juanca: I just realized that this is a great pattern:

Others can import this stylesheet for both variables and classes. 

This is way different than the SASS counterpart where we had to split the variables, the mixins, etc, into different files.

src/components/icon/icon.md
```diff
@@ -0,0 +1,14 @@
+```jsx
+const icon = require('../../svgs/icon-caution-fill.svg').default;
+<Icon name={icon.id} size="large" currentColor="caution" />
+```
+Sizes
+
+```jsx
+const icon = require('../../svgs/icon-caution-fill.svg').default;
+<div>
```
juanca: TBH, the icons next to each other do not look great / appealing.

Can we provide some margin within this example? Maybe we can structure the documentation a little bit different to make it more pleasant on the eyes?

src/components/icon/icon.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+describe('Icon', () => {
+  it('has defaults', () => {
+    const tree = renderer.create((
+      <Icon name={'foobar'} />
+    )).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  describe('Size API', () => {
```
juanca: Typo: `Size` => `size`

src/components/icon/icon.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+      sizes.forEach((size) => {
+        const tree = renderer.create((
+          <Icon name={'foobar'} size={size} />
+        )).toJSON();
+        expect(tree.props.className).toContain(`size-${size}`);
+      });
+    });
+  });
+
+  describe('focusable API', () => {
```
juanca: What does the `focusable` attribute on the SVG element do?

I tried setting it and tabbing into it but I can't in our deployment.

juanca:
Oh, something I don't think we covered when creating the story: for AC, let's ensure the component looks / works well in our support browsers.

src/components/icon/icon.test.jsx
roblevintennis: TLDR; Ima remove it from here. Backstory...

I think I see what happened, my guess is Huth referenced:

https://www.w3.org/TR/SVGTiny12/interact.html#FocusableAttribute

But SVGTiny12 is not in browsers and I believe SVG 2.0 which may add something similar never finalized yet (I think). I think even `tabIndex` is flaky and/or will require some more research.

Nice catch, and I think I'll just remove this. Note, it's in our bigmaven codebase as well, so we may just want to leave that since it probably doesn't mess anything up but maybe we do.

src/components/icon/icon.md
roblevintennis: Yeah, I can just one-off the layout from the component's specific CSS

juanca:
I want to play with the new Icon code in my input component before approving. But besides that existing comments, it looks pretty good!

styleguide/webpack.config.js
juanca: What's the reasoning for not bringing that over? Is that not needed for the current SVG in this repo?

src/components/icon/icon.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+          expect(tree.props.className).toContain('icon-base');
+          expect(tree.props.className).toContain(`fill-${color}`);
+          expect(tree.props.className).toContain(`color-${color}`);
+          expect(tree.props.className).toContain(`stroke-${color}`);
+        });
+      });
+    });
+  });
+
+  describe('title API', () => {
```
juanca: It looks like this does not work. Most of the internet is saying that an SVG does not support the title attribute.

src/components/icon/icon.test.jsx
roblevintennis: fixed

src/components/icon/icon.md
roblevintennis: image

src/components/icon/icon.test.jsx
roblevintennis: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements

src/components/icon/icon.md
juanca: Thanks!

src/components/icon/icon.test.jsx
juanca: Neither of those links support our implementation in this PR. 

If you read closely, those are talking about a `<title>` element instead of a `title=` attribute. 

Am I misreading it? Does the attribute translate to the element under the hood?

styleguide/webpack.config.js
roblevintennis: >What's the reasoning for not bringing that over? Is that not needed for the current SVG in this repo?

I did bring it over and I'm setting it true here. Not sure the confusion.

src/components/icon/__snapshots__/icon.test.jsx.snap
```diff
@@ -3,7 +3,6 @@
 exports[`Icon has defaults 1`] = `
 <svg
   className="icon-base"
-  focusable={false}
```
juanca: So I read into this...

This _is visually supported_ in Edge. It is also supported in IE (but missing styles). I think perhaps it is also supported in Chrome / Firefox / Safari but it might be the same thing as IE (basically missing styles).

I don't have a use-case for now. But it was worthwhile to explore the focusability problem of SVGs for a bit. 

We might never have to handle this if we wrap focusable SVGs in a <Button /> or something. 

Just spreading context between us.

styleguide/webpack.config.js
roblevintennis: Are you talking about the specific branch for svgo? If so, they merged my PR a long time ago and it was never deemed a priority in bigmaven to point back to `#master` for svgo, but it's in the version we're pulling in here. Fwiw, the alert icon itself uses this so it wouldn't be rendering the caution red if it wasn't being applied ;)

src/components/icon/icon.test.jsx
juanca: I agree that removing it is best for now. I also looked into it -- findings @ https://github.com/mavenlink/design-system/pull/42#discussion_r307064724

src/components/icon/icon.test.jsx
juanca: Oh, @roblevintennis , I think this is actionable. I do not see the title working in the deployment page.

roblevintennis: @juanca yeah, that was bogus to put as an attribute 🤦‍♂ 


13ee9b2 should fix it...I tried it locally and see the tooltip on hover (I just typed it in the live edit of styleguidist)

Ima add another commit with it like that way in one of the .md example too...but as of ^ commit it should at least be available for your input explorations

juanca:
Hey @roblevintennis, have you looked into: 

> for AC, let's ensure the component looks / works well in our support browser

juanca:
Some feedback that I am unsure whether it is actionable.

The distinction between `fill` and `stroke` on the API seems obvious. However, once you throw `currentColor` then it actually becomes a bit confusing. I had this problem when using the alert icon... I was just trying all the things and perhaps that is not the right way to do it. 

Should we document the difference between each? How would we effectively document the differences? 

IMO, I think we should make a dummy SVG that really showcases the differences as part of the markdown file.

juanca:
Looks like CI is red and did not deploy the latest code.

juanca:
Another question: 

Do we need SVGO optimizations? Are those only relevant on our Mavenlink application? Do we want to start the practice of only committing well-crafted SVGs -- basically avoid just committing the output of Sketch (or whatever tool we use)?

roblevintennis:
I'm capturing any delta feedback and answering question here so I can consolidate my last round of work here and easily find TODO items left...

- [ ] CI is red and did not deploy the latest code.
- [ ] Explain differences between `fill`, `currentColor` and `stroke` in the markdown file
- [ ] ensure the component looks / works well in our support browser

> Do we need SVGO optimizations? Are those only relevant on our Mavenlink application? Do we want to start the practice of only committing well-crafted SVGs -- basically avoid just committing the output of Sketch (or whatever tool we use)?


> Are those only relevant on our Mavenlink application?

No, the whole Icon API assumes such optimization and approach is being used e.g. 1. strip fillls and strokes generally, but leave ones with `currentColor`, and 2. use CSS to style the SVG (not presentational attributes). So, cleansing the SVGs as a safeguard makes sense. It doesn't preclude well crafted SVGs too—if it doesn't find a fill or stroke nothing might get applied for example. But it's much more fault tolerant to have this in place.

More details on this:

Our whole API is based off the assumption that you're wishing to use SVG Sprited `<symbol>`'s that must get dropped on the page so that the `<use xlink:href="ID-OF-SYMBOL"...` works. I've used several techniques/workflows to generate such sprites and they always include a transpile step from tooling such as grunt-svgstore, svgo and svg-loader in this case, and others. Besides general cleanup, the fills and strokes are removed so that the styling can come from your CSS Api instead (in this case, our CSS modules is ultimately taking care of that).

Once the SVG has been altered as described above, it becomes only useful in generating the symbol for sprite inclusion. Put differently, viewing them standalone can become difficult; and it's useful to be able to view them directly from within the svgs directory with something like Finder.

Also, I think enforcing well-crafted SVGs will add too much burden in placing these in the system. As it turns out, the `currentColor` must be added manually and so, for example, I DID create a well crafted `icon-caution-fill.svg`:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 12">
  <path fill="currentColor" d="M11.9 10.6c.1.3.1.6 0 .9-.2.3-.4.5-.7.5H1c-.3 0-.8-.2-.9-.5-.2-.3-.1-.9.1-1.2L5.3.5c.1-.3.4-.5.7-.5s.6.2.7.5l5.2 10.1zm-5-7.4v-.1C6.8 3 6.8 3 6.7 3H5.3c-.2 0-.2 0-.3.1v.2l.1 3.5c0 .2.2.2.3.2h1.2c.1 0 .2-.1.2-.2l.1-3.6zm.1 5c0-.1-.1-.2-.2-.2H5.2c-.1 0-.2.1-.2.2v1.5c0 .2.1.3.2.3h1.6c.1 0 .2-.1.2-.2V8.2z"/>
</svg>
```
But I think requiring that is burdensome and I don't see any problems with having svg-loader in the MDS pipeline. Put differently, I'd be surprised if many others could create a working well-crafted SVG that gels with this system easily. I only can because I sort of geeked out on it for a period, but being able to play with the SVG, and determine exactly what sub-element should have the `fill="currentColor"` or similar takes some practice. The most useful way to figure this out at the moment, is to look at the IconsAll in bigmaven which has every variation. 

> Should we document the difference between each? How would we effectively document the differences?

I think this first iteration could certainly explain these from the top of the corresponding markdown file. I'll add that to my checklist above.

> IMO, I think we should make a dummy SVG that really showcases the differences as part of the markdown file.

That increases scope for this task which should have an MVP target. I've captured it as another investigatory follow up task: https://www.pivotaltracker.com/story/show/167761758

roblevintennis:
Verification in IE screen grab:

VirtualBox_Custom Forms IDE Clone_07_08_2019_09_35_19

roblevintennis:
For checklist item:

> Explain differences between fill, currentColor and stroke in the markdown file

I've added that:

image

src/components/icon/icon.md
```diff
@@ -0,0 +1,39 @@
...
+Basic Usage
+
+```jsx
+const icon = require('../../svgs/icon-caution-fill.svg').default;
+<Icon name={icon.id} size="large" currentColor="caution" title="alert icon" />
+```
+Sizes
+
+```jsx
+const styles = require('./example.css'); /* For example purposes only! */
```
Bloodyaugust: Should we perhaps add a short `css` example to show how `currentColor` can be used?

roblevintennis: Gotcha, thanks!

Yeah, so, you cannot set this with CSS from the outside per se as the API will do it on your behalf from the icon component itself (via it's own CSS module in icon.css of course).

That said, I've added a small example to hopefully drill this in:

image

roblevintennis:
Verified on branch's staging. Merging!

PR #46, opened 2019-07-26
List of colors in MDS design system

roblevintennis:
Type of work: colors section

Deployment URL: https://mavenlink.github.io/design-system/colors

https://www.pivotaltracker.com/story/show/167479026

## Motivation

Developers (and anyone interested in the offerings of MDS) have a need to see which colors we are officially supporting (and even deprecating). At the moment, there seems to be no way to see all the colors we are offering from our MDS site. If we had a page with all the colors, it would help others see how they can customize their components and other HTML things.

## Acceptance Criteria

We have a list of colors in MDS

image

image

### DIFF
--- a/styleguide/content/soul/colors.css
+++ b/styleguide/content/soul/colors.css
@@ -22,13 +22,11 @@
   display: inline-block;
 }
 
-:global(.swatch-primary) {
-  background-color: var(--palette-primary-base);
+:global(.swatch-title) {
+  font-size: 24px;
 }
 
-:global(.swatch-brand) {
-  background-color: #001c33;
-}
+
 
 :global(.swatch-black),
 :global(.swatch-brand),
@@ -41,6 +39,9 @@
 }
 
 /* Actions */
+:global(.swatch-primary) {
+  background-color: var(--palette-primary-base);
+}
 :global(.swatch-action) {
   background-color: var(--palette-action-base);
 }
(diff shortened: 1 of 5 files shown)


styleguide/content/soul/colors.md
```diff
@@ -0,0 +1,92 @@
...
+  <div class="rgba">`rgba(0, 0, 0, .87)`</div>
+  <div class="hex">`#000000`</div>
+</div>
+<div class="color-container grays">
+  <div class="rgba">`rgba(0, 0, 0, .54)`</div>
+  <div class="hex">`#757575`</div>
+</div>
+<div class="color-container grays">
+  <div class="swatch-grey-base">
+    <span class="text">Md Gray</span>
```
juanca: Can we unify the label with its name? It's a bit weird to call it grey base but also medium grey. 

Also, it looks like we are diverging on spelling as well. Gray vs grey.

juanca:
Do we want to also expose the variable name?

styleguide/content/soul/colors.css
```diff
@@ -0,0 +1,126 @@
+@import "../../../src/styles/colors.css";
+@import "../../../src/styles/typography.css";
```
juanca: It looks like we aren't using this import. Let's clean this up

juanca:
I don't understand why but it looks like some text is not being rendered well.

Screen Shot 2019-07-29 at 12 55 39 PM

juanca:
Can you update the PR description as well? It looks like it says this is still a WIP.

roblevintennis:
Regarding the fixes requested I've done:
✅We should unify gray vs grey.
✅We should fix the weird text glitches.
✅We should not import a stylesheet but not use it.

For

>I am not sure but perhaps we should also expose the variable names.

Let's figure that out and do it later. I'm not sure if the example at the top is enough and/or if adding code snippets clutters things. Since we can iterate, I'd prefer to consider that for a later rev.

roblevintennis:
Yes, regarding the light greys I've fixed in latest commit:

image

Issue #34, opened 2019-07-17
Documentation: how to set up in another repository


juanca:
I was a little rusty on setting up a new Webpack repository and ended up re-learning a few things. I think this can be alleviated by having some documentation about how to set up the library in another repository.

A few things to ensure we include:

- Babel + React processing
- Style + CSS modules processing

It might make sense to _have_ an example repository. Hopefully it can have integration tests as well. Good way to ensure a release candidate is releasable.

juanca:
It looks like this problem is exacerbated with the introduction of SVGs. 

I will triage this now to establish a pattern and do better at avoiding future conflicts.

juanca:
I think our installation section is good enough for now: https://github.com/mavenlink/design-system/tree/1c8b990202a7769b718fe0dfd02e8baac1b851e5#installation

PR #47, opened 2019-07-26
<Input /> v2

juanca:
Type of work: component

Deployment URL: https://mavenlink.github.io/design-system/input

(Optional for outside contributions) Tracked work in Mavenlink: https://www.pivotaltracker.com/story/show/165360149

## Motivation

In order to better support MavenPages, we need an input component that is already accessible and forward-looking with design. 

Version 2 of the input will bake in better accessibility by enforcing a label. It will also remove the pass-through properties interface to avoid consumer applications from misusing the input -- for example, we have a case where a developer used the input as a checkbox. 

Misuses of components will create a maintenance burden in the future and misdirect vital demand information about MDS. 

Useful resources:

- http://web-accessibility.carnegiemuseums.org/code/forms/
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

## Acceptance Criteria

- [x] It looks great in our support browsers list
- [x] It is backwards compatible with existing usages in Mavenlink (or usages are changed for more appropriate inputs)

## Change log entry

Implement `<Input>` version 2. More accessible and better defined property interface. 
Improve `<Button>`: allow `name` and `value` properties.


src/components/input/input.jsx
```diff
@@ -3,35 +3,34 @@ import React from 'react';
...
 
-/**
- * Base input component. Accepts any prop that an input can take.
- * @see [RFC](https://github.com/mavenlink/rfc/pull/75)
- * @see [Design spec](https://zpl.io/295Yp7p)
- */
-export default function Input({ 'aria-labelledby': ariaLabelledBy, className, id, ...rest }) {
+export default function Input(props) {
   return (
     <input
```
juanca: It looks like I missed a few other usages of this existing `<Input />`. Hmmm...

juanca:
Closes https://github.com/mavenlink/design-system/issues/34

src/components/input/input.css
```diff
@@ -23,3 +28,20 @@
...
+
+.invalid-icon {
+  box-sizing: border-box;
+  height: 100%;
+  pointer-events: none;
+  position: absolute;
+  right: 0;
+}
+
+.invalid {
```
juanca: I just realized. I might not have to rely on a property for an invalid state. 

TODO: Try out psuedo state for invalid. A required input that is invalid will somehow style itself differently -- so can I utilize whatever is happening under the hood? Are there any cons to using this technique?

src/components/input/input.jsx
```diff
@@ -1,37 +1,54 @@
...
-    <input
-      {...rest}
-      aria-labelledby={ariaLabelledBy}
-      className={className}
-      id={id}
-    />
+    <div className={styles.container}>
+      <label htmlFor={props.id} className={styles.label}>
+        {props.label}
+        {props.required && <span>(Required)</span>}
```
juanca: Looks like `(Required)` and `(Optional)` do not have any consistent logic due to the nature of our application.

Sometimes a form has all inputs as required inputs which means we do not want to clutter the UI with `(Required)`, opting for something else.

Same thing with `(Optional)`. 

Thus, I think we should not bake any logic about appending to the label and let the context of the `<Input>` decide on how the label looks.

src/components/input/input.jsx
```diff
@@ -1,37 +1,65 @@
...
+export default function Input(props) {
   return (
-    <input
-      {...rest}
-      aria-labelledby={ariaLabelledBy}
-      className={className}
-      id={id}
-    />
+    <React.Fragment>
+      <label htmlFor={props.id} className={getLabelClassName(props.cssLabel, props.invalid)}>
```
roblevintennis: How are ID collisions being handled? It looks like this API places the burden entirely on the consumer with no recommendation. Realistically, are they going to use a uuid generator and prepend something like `<GENERATED-UUID>-meaingful-id-name`? I think should think about this and address it to make this as frictionless as possible for MDS consumers.

src/components/input/input.css
```diff
@@ -2,24 +2,56 @@
...
 @import "../../styles/typography.css";
 
+.container {
+  display: inline-block;
+  position: relative;
+}
+
+.label {
+  display: block;
+  font-family: var(--mavenlink-type-font-family);
```
roblevintennis: What happens if I'm a consumer and I want to use my own font stack while using MDS components? I think we have this same issue in Button. We're enforcing ourselves. I think we should have a discussion with Design and Product to ensure this is the approach we want to take. There's certainly an argument for both sides (forcing them to be more consistent, not forcing them to allow some flexibility and white labelling opportunities)

src/components/input/input.jsx
```diff
@@ -1,37 +1,65 @@
...
+
+export default function Input(props) {
   return (
-    <input
-      {...rest}
-      aria-labelledby={ariaLabelledBy}
-      className={className}
-      id={id}
-    />
+    <React.Fragment>
```
roblevintennis: Why haven't we set up:

```jsx
<>
  <...>
</>
```

shorthand? I think it would be nice as `<React.Fragment>...</React.Fragment>` is super verbose

src/components/input/input.jsx
```diff
@@ -1,37 +1,65 @@
...
+      <label htmlFor={props.id} className={getLabelClassName(props.cssLabel, props.invalid)}>
+        {props.label}
+      </label>
+      <div className={styles.container}>
+        <input
+          className={getClassName(props.className, props.invalid)}
+          disabled={props.disabled}
+          id={props.id}
+          name={props.name}
+          onChange={props.onChange}
```
roblevintennis: Do we plan to support something like `onKeyDown`? This allows for things like handling ENTER:

```js
onKeyDown={handleKeyDown}
...
  const handleKeyDown = (e) => {
    if (e.key === 'Enter') {
      e.preventDefault();
      if (inputValue) {
        props.onAdd(inputValue);
        setInputValue('');
      }
    }
  }
```

Are there other events we should allow? Obviously, we should keep surface API lean, but I think the key down use case is compelling enough to consider seriously.

src/components/input/input.jsx
```diff
@@ -1,37 +1,65 @@
...
+          placeholder={props.placeholder}
+          required={props.required}
+          type="text"
+          value={props.value}
+        />
+        {props.invalid && <Icon className={styles['invalid-icon']} currentColor="caution" name={cautionSvg.id} />}
```
roblevintennis: Do we want to standardize on this specific ordering of invalid messages? Or, allow the consumer of MDS to handle positioning such things themselves? The later seems consistent with our approach of almost always supporting `{ children }` where it makes sense.

src/components/input/input.md
```diff
@@ -1,3 +1,11 @@
 ```jsx
-<Input id="greeting" placeholder="hello world" />
+<Input id="greeting" label="Greeting" />
+```
+
+```jsx
+<Input id="greeting" label="Greeting" invalid />
```
roblevintennis: what's `invalid` without a value mean here exactly?

src/components/input/input.css
roblevintennis: __UPDATE: Nathan was able to confirm preference to go ahead and "force" use of Mavenlink font stack.__

So this is no longer an issue and we're good as we have here!

src/components/input/input.jsx
roblevintennis: UPDATE: I don't think this is yet fully finalized, but again, after speaking with Nathan at standup today, I think he's fine with having some _opinion_ and forcing matters when we want to. So, I think we still have some questions to answer here, but nothing blocking per se.

juanca: Yup, we're going to stick with the caution icon being on the right side.

src/components/input/input.md
juanca: It's standard JSX. When you specify a value like this, it implies setting it as `true`. 

```
<Input id="greeting" label="Greeting" invalid={true} />
```

src/components/input/input.jsx
juanca: I am going to not support it until we have a real use-case. I might be useful in some cases but perhaps that is just another component.

src/components/input/input.jsx
juanca: I have the opposite opinion. I think `<>...</>` is unnecessary short-hand.

src/components/input/input.css
juanca: Yup. It is also easy to use CSS to override our defaults.

src/components/input/input.jsx
juanca: At the moment, I have a use-case to have a consumer-provided ID API. Kinda weird, huh? But that is the current real world. 

I think it's easy for others to provide their own unique ID that is meaningful to them. For example, if you are building a feature in Mavenlink, you should probably prefix your IDs with your feature name. 

It might make sense to break this API in the future though. I think then it would be ideal to provide a "migration" guide or something.

roblevintennis: Yeah, as I started on the file picker and thought about this more, I agree that while it's burdensome for consumer (it's burdensome for whoever though), it really should be an injected dependency. On the component level it could be argued the component should generate it's own uuid or similar, but I'm fine with the container or consumer doing so. Bigger stuff can leverage server data ids of course, but at the component level I think `feature-generateduuid` or similar is how folks will have to go perhaps.

roblevintennis:
I think juanca's answered my questions, and after speaking with Nathan required my biggest concerns, the big picture take away for posterity here, is we're fine with locking in certain conventions e.g. where the alert icon will show up upon a validation error, forcing the use of Mavenlink's font within the widget, etc., etc., and so I think what we have here is good to go.

src/components/input/input.jsx
roblevintennis: What happens when you don't have this available and you upgrade a library if you don't keep up with the times? Well, you'll be temporarily stuck at `"react-styleguidist": "9.0.4"` for example:

https://github.com/styleguidist/react-styleguidist/issues/1321

I got same error as the result trying to use latest I believe because the library uses shorthand.

juanca: If we do a bit more research, I think it comes with React. So the better question, why do we not have it already by just using React? Are we out-of-date with React? 

If we do have to do additional set up, it probably means the library is adding additional peer dependencies without documentation. Or their own dependency tree is broken.

juanca:
If we have a bunch of inputs in a couple of rows, the label spacing does not make it easy to discern the label-input pairs.

Proposed label changes:

margin-top: 4px
margin-bottom: 2px

setup-enzyme.js
```diff
@@ -2,3 +2,7 @@ import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 configure({ adapter: new Adapter() });
+
+console.error = (message) => { // eslint-disable-line
+  throw new Error(message);
```
Bloodyaugust: Does this make all errors thrown in enzyme tests actually log? What's the goal?

src/components/input/input.jsx
```diff
@@ -1,37 +1,91 @@
...
-      className={className}
-      id={id}
-    />
+    <React.Fragment>
+      <label htmlFor={props.id} className={getLabelClassName(props.cssLabel, props.invalid)}>
+        {props.label}
+      </label>
+      <div className={props.cssContainer}>
+        <input
+          className={getClassName(props.className, props.invalid)}
```
Bloodyaugust: I'm concerned about the number of configurations here. Are there things we could do, like advocate for the usage of native handlers for some of these?

src/components/input/input.jsx
```diff
@@ -1,37 +1,91 @@
...
+  name: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  onInput: PropTypes.func,
+  onKeyDown: PropTypes.func,
+  placeholder: PropTypes.string,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  type: PropTypes.oneOf([
```
Bloodyaugust: We should consider making these `type`s separate components.

.eslintrc.js
```diff
@@ -1,6 +1,11 @@
 module.exports = {
   extends: 'mavenlint-react',
   parser: 'babel-eslint',
+  rules: {
+    'import/extensions': ['error', 'always', {
```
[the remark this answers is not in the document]
juanca: It's an ESLint rule! See docs @ https://github.com/benmosher/eslint-plugin-import/blob/112a0bf442e52b25cd7029a9905df2508e191ac1/docs/rules/extensions.md

We are always specifying extensions in our `import` statements (except for packages).

Basically, this means we _should_ not be relying on Webpack magic in our source files. The reasoning for this: we do not want to impose _too_ much Webpack configuration magic on a host repository.

README.md
```diff
@@ -14,29 +14,48 @@ A set of [React](https://reactjs.org/) components created by, and for, [Mavenlin
   yarn add @mavenlink/design-system
   ```
 
-- Setup [CSS modules](https://github.com/css-modules/css-modules). One way of doing that is with [css-loader for Webpack](https://github.com/webpack-contrib/css-loader#modules)
+- Setup [React JSX processing](https://reactjs.org/docs/jsx-in-depth.html). One way of doing that is with [`babel-loader` for Webpack](https://webpack.js.org/loaders/babel-loader/).
```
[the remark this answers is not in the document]
juanca: Yup!

README.md
```diff
@@ -14,29 +14,48 @@ A set of [React](https://reactjs.org/) components created by, and for, [Mavenlin
...
   ```js
   // Webpack configuration
+  const SpriteLoaderPlugin = require('svg-sprite-loader/plugin');
+
   module.exports = {
     module: {
-      rules: [
-        {
-          test: /\.css$/,
+      rules: [{
```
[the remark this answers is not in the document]
juanca: We have a Webpack file for our styleguidist configuration. It, however, has some problems because of styleguidist. This snippet of Webpack configuration makes it easier to pick up MDS as a dependency in a Webpack project without any mess.

jest.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  moduleNameMapper: {
+    css$: 'identity-obj-proxy',
```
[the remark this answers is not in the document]
juanca: This is a Jest tesing thing. Jest runs in a Node environment. We do not have a concept of CSS (or SVGs) in a Node environment -- only straight JavaScript. This is normal practice to stub out file types in a Node environment when running tests.

setup-enzyme.js
juanca: Jasmine does _not_ fail when `console.error` is called since it is not an actual _thrown_ error. This just makes sure we do not have any `console.error`s in our tests.

src/components/input/input.jsx
juanca: This is an opinionated implementation. 

As you can see from the code removed section, the MDS Input v1 was just that -- it passed through all props onto the `<input>` element. We thought that would be the best for MDS library but we saw incorrect usages of our MDS Input. Since we can just pass any kind of `type` into the MDS input, we ended up in a situation where a feature team used our MDS input as a checkbox -- which basically gutted the whole thing. 

Now, I think I understand the whole battle between configuration vs composition. While we do see many configuration options, I do not think this is a "configurable" component -- I think that is usually applied to components that change their behavior based on their properties. In this case, we are whitelisting the configuration options to the `<input>` -- which means we _want_ to control how much configuration is allowed for native elements.

src/components/input/input.jsx
juanca: That's an interesting point! 

I think that is 100% the right thing when dealing with Input types that are drastically different than a text field.

I am not convinced we need to do it for all types though. I think we might want to get more data on this but for now this is a reasonably small list!

src/components/input/input.jsx
roblevintennis: Time to have another look at typescript :) We're having same sorts of nuanced issues at my new place and have even gone so far as to throw a fatal Error if unrecognized props are passed in—this starts to make the code less readable and it's not fault tolerant per se. Typescript lets you know at compile time, you can just declare the interface you expect, and all is good. It IS a PITA to configure though.

Something to maybe consider longer term I  dunno :)

Bloodyaugust: I'm definitely not comfortable with the idea that implementation issues should be solved by just switching languages, there's a wealth of baggage for any language. 😄 I do like the opinionated approach, but perhaps we could forward native event handlers specifically? That way our interface maintains at least part of the underlying HTML native interface, and the part that we need for proper usage of the component while also supporting wrappers for more advanced behavior.

src/components/input/input.jsx
Bloodyaugust: Fair enough! We can always iterate if we see this bloating the component down the line, I'm mainly worried about "god" components, and I think this is still a long way from that.

src/components/input/input.jsx
juanca: There are no native event handles in React! React uses its own event system (and, sure, it ties it into native event handles but that is abstracted away from us React developers).

juanca: @roblevintennis Hello! 👋 

Yeah, I'm on the same boat as always: I don't want to adopt yet another non-standard thing if I can post-pone it as long as possible. We adopted React and CSS Modules but we did wait a good 4 years before doing any of that. Hah, maybe in 4 years I'll consider TS. ;)

roblevintennis: I feel ya kinda ;) In this spirit of coding towards standards, I'd be very curious if esnext et al introduce some sort of notion of interfaces; although it's sort of useless as a runtime only thing imo (main thing is to get compile / transpile knowledge of bad inputs).

If you think about it, really, Webpack is a very specific way to set things up and many folks consider it extremely painful (folks at my new gig are in favor of using create react app or gatsby or something because they've tired of fighting Webpack). React leaves itself around in your code (e.g. read the philosophy of Svelte which compiles itself away). CSS Modules is really just a nifty spec which ala css-loader and Webpack does some name munging tricks, and as we know does not work perfectly (although I'm still using it as it's the least of all unfortunate options).

All to say, our jobs and lives in web-land are still a mess and not really so standards-based 😋 I am sad that my new gig is still quite rooted in Sass but maybe we'll get on the PostCSS train in 2020.

roblevintennis: > there's a wealth of baggage for any language

Typescript is not really a language but more of a superset of JavaScript. Yes, there's some baggage, but also some benefits. Especially if you like to use inversion of control (which I know you loathe from a particular meeting I recall where we brought up DI and you were a bigtime detractor).

PR #59, opened 2019-12-04
Add a new Custom Field centric Text Input component

Bloodyaugust:
- Tweak the sidebar to use a dynamic width to account for long component names

Type of work: component

Deployment URL: https://mavenlink.github.io/design-system/custom-field-input-text

(Optional for outside contributions) Tracked work in Mavenlink: https://www.pivotaltracker.com/story/show/169509661

## Motivation
New "custom field" inputs of various types are needed for 3rd party custom forms as part of the MBridge work.

## Acceptance Criteria

1. The component is in the MDS site
2. It looks like an MDS input
3. It has a label
4. It can be disabled
5. It has placeholder text
6. It has an optional "unique value" constraint
7. It follows the Zeplin design

### DIFF
--- a/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
+++ b/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
@@ -1,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CustomFieldInputText has defaults 1`] = `
+<div
+  className="custom-field-input-text"
+>
+  <div
+    className="heading-container"
+  >
+    <label
+      className="label"
+    >
+      Input Descriptor
+    </label>
+    <span
+      className="additional-information-icon"
+    >
+      ?
+    </span>
+  </div>
+  <div
+    className="input-container"
+  >
+    <input
+      disabled={false}
+      onClick={[Function]}
+      type="text"
+    />
+  </div>
+  <span
+    className="help"
+  />
+</div>
+`;
--- a/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
+++ b/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
@@ -17,6 +17,7 @@ exports[`CustomFieldInputText has defaults 1`] = `
     className="input-container"
   >
     <input
+      className="input"
       disabled={false}
       onClick={[Function]}
       type="text"
--- a/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
+++ b/src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
@@ -12,11 +12,6 @@ exports[`CustomFieldInputText has defaults 1`] = `
     >
       Input Descriptor
     </label>
-    <span
-      className="additional-information-icon"
-    >
-      ?
-    </span>
   </div>
   <div
     className="input-container"
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -24,12 +24,7 @@
   color: var(--palette-caution-base);
 }
 
-.input-container {
-  height: 34px;
-  margin: var(--spacing-small) 0;
-}
-
-input {
+.input {
   border: 1px solid var(--palette-grey-light);
   border-radius: var(--spacing-x-small);
   color: var(--palette-grey-x-dark);
@@ -37,17 +32,22 @@ input {
   padding: var(--spacing-medium);
 }
 
-.input-container input::placeholder {
-  color: var(--palette-grey-dark);
-}
-
-.disabled input {
+.disabled .input {
   background-color: var(--palette-grey-light);
 }
 
-.error input {
+.error .input {
   border-color: var(--palette-caution-base);
   padding-right: var(--spacing-x-large);
+}
+
+.input-container {
+  height: 34px;
+  margin: var(--spacing-small) 0;
+}
+
+.input-container input::placeholder {
+  color: var(--palette-grey-dark);
 }
 
 .input-icon-container {
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -8,7 +8,7 @@
 
 .heading-container {
   color: var(--palette-grey-dark);
-  font-size: var(--mavenlink-type-body-font-size);
+  font-size: var(--mavenlink-type-small-size);
 }
 
 .error .heading-container {
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -9,9 +9,7 @@
   cursor: pointer;
   display: inline-block;
   height: var(--spacing-large);
-  left: var(--spacing-medium);
-  padding-left: 1px;
-  position: relative;
+  margin-left: var(--spacing-medium);
   text-align: center;
   width: var(--spacing-large);
 }
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -2,18 +2,6 @@
 @import '../../styles/spacing.css';
 @import '../../styles/typography.css';
 
-.additional-information-icon {
-  border: 1px solid var(--palette-grey-light);
-  border-radius: 50%;
-  color: var(--palette-grey-dark);
-  cursor: pointer;
-  display: inline-block;
-  height: var(--spacing-large);
-  margin-left: var(--spacing-medium);
-  text-align: center;
-  width: var(--spacing-large);
-}
-
 .custom-field-input-text {
   font-family: var(--mavenlink-type-font-family);
 }
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -1,0 +1,84 @@
+@import '../../styles/colors.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography.css';
+
+.additional-information-icon {
+  border: 1px solid var(--palette-grey-light);
+  border-radius: 50%;
+  color: var(--palette-grey-dark);
+  cursor: pointer;
+  display: inline-block;
+  height: var(--spacing-large);
+  left: var(--spacing-medium);
+  padding-left: 1px;
+  position: relative;
+  text-align: center;
+  width: var(--spacing-large);
+}
+
+.custom-field-input-text {
+  font-family: var(--mavenlink-type-font-family);
+}
+
+.heading-container {
+  color: var(--palette-grey-dark);
+  font-size: var(--mavenlink-type-body-font-size);
+}
+
+.error .heading-container {
+  color: var(--palette-caution-base);
+}
+
+.help {
+  color: var(--palette-grey-dark);
+  font-size: var(--mavenlink-type-base-size);
+}
+
+.error .help {
+  color: var(--palette-caution-base);
+}
+
+.input-container {
+  height: 34px;
+  margin: var(--spacing-small) 0;
+}
+
+.input-container input {
+  border: 1px solid var(--palette-grey-light);
+  border-radius: var(--spacing-x-small);
+  color: var(--palette-grey-x-dark);
+  font-size: var(--mavenlink-type-base-size);
+  padding: var(--spacing-medium);
+}
+
+.input-container input::placeholder {
+  color: var(--palette-grey-dark);
+}
+
+.disabled .input-container input {
+  background-color: var(--palette-grey-light);
+}
+
+.error .input-container input {
+  border-color: var(--palette-caution-base);
+}
+
+.input-icon-container {
+  background-color: var(--white);
+  display: inline-block;
+  height: 20px;
+  position: absolute;
+  margin-left: -28px;
+  margin-top: var(--spacing-small);
+  width: 20px;
+}
+
+.label {
+  margin-right: var(--spacing-small);
+}
+
+.optional {
+  display: inline-block;
+  margin-right: var(--spacing-small);
+  min-width: 30px;
+}
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -47,6 +47,7 @@
   color: var(--palette-grey-x-dark);
   font-size: var(--mavenlink-type-base-size);
   padding: var(--spacing-medium);
+  padding-right: var(--spacing-x-large);
 }
 
 .input-container input::placeholder {
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -41,7 +41,7 @@
   margin: var(--spacing-small) 0;
 }
 
-.input-container input {
+input {
   border: 1px solid var(--palette-grey-light);
   border-radius: var(--spacing-x-small);
   color: var(--palette-grey-x-dark);
@@ -54,11 +54,11 @@
   color: var(--palette-grey-dark);
 }
 
-.disabled .input-container input {
+.disabled input {
   background-color: var(--palette-grey-light);
 }
 
-.error .input-container input {
+.error input {
   border-color: var(--palette-caution-base);
 }
 
--- a/src/components/custom-field-input-text/custom-field-input-text.css
+++ b/src/components/custom-field-input-text/custom-field-input-text.css
@@ -47,7 +47,6 @@ input {
   color: var(--palette-grey-x-dark);
   font-size: var(--mavenlink-type-base-size);
   padding: var(--spacing-medium);
-  padding-right: var(--spacing-x-large);
 }
 
 .input-container input::placeholder {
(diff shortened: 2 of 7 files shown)


Bloodyaugust:
@juanca I think we should talk about the validity of AC 6. Should we do that here or internally?

Bloodyaugust:
> @juanca I think we should talk about the validity of AC 6. Should we do that here or internally?

Turns out that AC was making sure that the product constraint of "errors from the backend can be displayed", which is supported with the `error` state and the configurable help text, so we're good there.

juanca:
There is a weird overlap into the left navigation menu that is not present on `master`:

Screen Shot 2019-12-05 at 5 15 02 PM

juanca:
We might want a design review on this. It looks like there is missing spacing around the icon when there is enough text to fill the input:

MDS Input

Screen Shot 2019-12-05 at 5 23 19 PM

versus 

MDS Custom Field Text Field

Screen Shot 2019-12-05 at 5 24 08 PM

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,84 @@
...
+
+.additional-information-icon {
+  border: 1px solid var(--palette-grey-light);
+  border-radius: 50%;
+  color: var(--palette-grey-dark);
+  cursor: pointer;
+  display: inline-block;
+  height: var(--spacing-large);
+  left: var(--spacing-medium);
+  padding-left: 1px;
```
juanca: It seems this is causing the circle to be more oval-y than a value of `0px`:

1px 

Screen Shot 2019-12-05 at 5 32 04 PM

0px

Screen Shot 2019-12-05 at 5 32 26 PM

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,84 @@
+@import '../../styles/colors.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography.css';
+
+.additional-information-icon {
+  border: 1px solid var(--palette-grey-light);
+  border-radius: 50%;
+  color: var(--palette-grey-dark);
+  cursor: pointer;
```
juanca: I'm not sure this is a valid application of `pointer`. From the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#Values):

> The cursor is a pointer that indicates a link. Typically an image of a pointing hand.

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,84 @@
...
+  display: inline-block;
+  height: var(--spacing-large);
+  left: var(--spacing-medium);
+  padding-left: 1px;
+  position: relative;
```
juanca: We want to avoid using relative positioning. While there are valid cases of relative positioning, we normally avoid it when we want to pixel-push around elements. 

position: relative && left: ?px

Screen Shot 2019-12-05 at 5 46 09 PM

position: static && margin-left: ?px

Screen Shot 2019-12-05 at 5 46 29 PM

juanca: It's pretty hard to find this in the documentation but an answer in Quora does a good job at explaining the functional difference: 

https://www.quora.com/What-are-the-differences-between-margin-left-and-left-in-CSS

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,67 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import cautionSvg from '../../svgs/icon-caution-fill.svg';
+import Icon from '../icon/icon.jsx';
+import styles from './custom-field-input-text.css';
+
+function getRootClassName(className, error, disabled) {
+  const resolvedClassName = className || styles['custom-field-input-text'];
```
juanca: This is a good time to explore the benefits of using a default property! Instead of using in-band logic we can leverage:

```
CustomFieldInputText.defaultProps = {
  className: styles['custom-field-input-text'],
};
```

Same thing but slicker! It _even_ is documented in our MDS Site!

Bloodyaugust:
> There is a weird overlap into the left navigation menu that is not present on `master`:
> 
> Screen Shot 2019-12-05 at 5 15 02 PM

Yeah, looks like the hero images are statically sized. I'll back out the width change, and we can address this in a separate PR.

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,84 @@
+@import '../../styles/colors.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography.css';
+
+.additional-information-icon {
```
juanca: Question:

- Could we postpone shipping this additional info icon until we get a tooltip functionality?

juanca: From our 1<>1: 

- Let's remove this icon as a commit and create a pivotal story to follow up with tooltip functionality (not for this version since we might be waiting on Mavengers for their tooltip work)

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (disabled) {
+    return `${resolvedClassName} ${styles.disabled}`;
```
juanca: RE: Do we want to fully utilize CSS Composition through CSS Modules or concatenate our default styles?

I think we need to look at a few different examples:

SOON TO COME

juanca: Possibly, we might want design to give us their opinion on how a CSS API should interact with the defaults.

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,84 @@
...
+  color: var(--palette-grey-x-dark);
+  font-size: var(--mavenlink-type-base-size);
+  padding: var(--spacing-medium);
+}
+
+.input-container input::placeholder {
+  color: var(--palette-grey-dark);
+}
+
+.disabled .input-container input {
```
juanca: Soon to come: example on using nesting versus using flat selectors.

Bloodyaugust: As an example of how overriding would currently work:

```
<CustomFieldInputText className="my_new_classname" />

.my_new_classname.disabled input {
  background-color: green;
}
```

juanca:
```
.my_new_classname input {
  composes: disabled from 'custom-field-text-input.css'
  background-color: green;
}
```

src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
juanca: Perhaps we need documentation in `contributing/testing.md` (another PR or make a new issue for it)

src/components/custom-field-input-text/custom-field-input-text.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+    });
+  });
+
+  describe('onClick API', () => {
+    it('sets the onclick handler', () => {
+      const onClickSpy = jest.fn();
+      const tree = renderer.create((
+        <CustomFieldInputText onClick={onClickSpy} />
+      )).root;
+      tree.findByType('input').props.onClick();
```
juanca: An idea: possible new issue on having a document for test practices for this repository -- with links and stuff. Nothing too heavy.

src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
juanca: Done @ https://github.com/mavenlink/design-system/issues/60

src/components/custom-field-input-text/custom-field-input-text.css
Bloodyaugust: Re 1:1 discussion, we should review this with design.

src/components/custom-field-input-text/custom-field-input-text.css
Bloodyaugust: https://www.pivotaltracker.com/story/show/170224608

src/components/custom-field-input-text/custom-field-input-text.css
juanca: I'm working through improving our Engineering Wiki's CSS guides with reasons for using non-nested selectors and class names. 

I think we're okay with not doing so in *this* component because:

1. This is the first release, we can iterate to improve it when we need to.
2. This is a very specific kind of component: it is just for custom fields
3. It does not have a React node as part of its properties interfaces

Though, when I do end up polishing up the new documentation, I'll try to link it here. But not a blocker!

src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
```diff
@@ -0,0 +1,34 @@
...
+  <div
+    className="heading-container"
+  >
+    <label
+      className="label"
+    >
+      Input Descriptor
+    </label>
+    <span
+      className="additional-information-icon"
```
juanca: As previously discussed, we should be removing this.

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,71 @@
...
+.error .help {
+  color: var(--palette-caution-base);
+}
+
+.input-container {
+  height: 34px;
+  margin: var(--spacing-small) 0;
+}
+
+input {
```
juanca: We *cannot* be styling tag names. This will literally style all other inputs on the page as well since it is a generic selector. Instead, use a class name.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  id: PropTypes.string,
+  name: PropTypes.string,
+  onClick: PropTypes.func,
+  placeholder: PropTypes.string,
+  required: PropTypes.bool,
+  value: PropTypes.string,
+};
+
+CustomFieldInputText.defaultProps = {
+  className: styles['custom-field-input-text'],
```
juanca: Oh, this is another question I wanted to ask but forgot:

What are your thoughts around using `styles['custom-field-input-text']` versus `styles.customFieldInputText`? 

For context, we haven't really decided on either. As an engineering org, we have historically used the hyphenated case. But the recent introduction for CSS modules has had a few people wondering whether we should switch.

I'm indifferent TBH. But if we do develop some reasoning for either then we could probably add a linter rule to ensure consistency in our style sheet files.

src/components/custom-field-input-text/custom-field-input-text.test.jsx
juanca: https://github.com/mavenlink/design-system/issues/60

src/components/custom-field-input-text/custom-field-input-text.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+      expect(tree.props.className).toEqual('prioritize-me');
+    });
+  });
+
+  describe('disabled API', () => {
+    it('can be disabled', () => {
+      const tree = renderer.create((
+        <CustomFieldInputText disabled={true} />
+      )).root;
+      expect(tree.findByProps({ disabled: true })).toBeTruthy();
```
juanca: What's the difference between using a truthy assertion and a specific assertion on the disabled property? Is one better than the other or are they the same?

We have historically preferred simple (and informational) error messages on test failures. When a truthy statement like this fails, it is either "expected false to be true" or "expected <large object serialization> to be true".  In this case, it is the latter.

juanca: This is also interesting: if I were to move the `disabled={props.disabled}` form the `<input>` to the `<label>`, this test still passes. Is that intended or a reasonable pattern?

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  disabled: false,
+  error: false,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  helpText: undefined,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+      </div>
+      <span className={styles.help}>{props.helpText}</span>
+    </div>
+  );
+}
+
+CustomFieldInputText.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  helpText: PropTypes.string,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  id: PropTypes.string,
+  name: PropTypes.string,
+  onClick: PropTypes.func,
+  placeholder: PropTypes.string,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+  required: PropTypes.bool,
```
juanca: This seems to be missing in tests.

src/components/custom-field-input-text/__snapshots__/custom-field-input-text.test.jsx.snap
Bloodyaugust: My bad, forgot to update the snapshot!

src/components/custom-field-input-text/custom-field-input-text.css
Bloodyaugust: Ack, nuked the wrong thing! I'll correct this.

src/components/custom-field-input-text/custom-field-input-text.jsx
Bloodyaugust: I am also indifferent as to which we choose, there are plenty of arguments for both... but I do think that we should choose, and implement linting rules, so... I'm very helpful, I know. 😄

src/components/custom-field-input-text/custom-field-input-text.test.jsx
Bloodyaugust: Turns out there were a few issues with these tests! Truthiness expectations will be removed,  and all tests now explicitly `findByType` their target, turns out that `tree` contains `CustomFieldInputText` as the top level element, so all those expectations found the props on the root element, right where we put them a line before in the `renderer.create` call. 😆

Bloodyaugust:
@juanca Fixes for the various feedback items have been pushed! The tests have been mostly re-written, as a heads up.

The icon spacing should have been fixed with yesterday's push, perhaps a caching artifact made it still broken? 😬 

As for thought leadership, I do like the idea of sitting down to decide which is ultimately better and going with it, I just think the difference in the approaches leaves little room for real pros and cons, while the advantages of making a choice one way or another seem pretty clear. 😄

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.input-container input::placeholder {
+  color: var(--palette-grey-dark);
+}
+
+.input-icon-container {
+  background-color: var(--white);
+  display: inline-block;
+  height: 20px;
```
juanca: This isn't really a blocker and more of nit-picking: 

Usually, we want to avoid setting a height + width on an absolutely position element in favor of letting the browser do its own thing.

In this case, we are not correctly calculating the boundaries of the actual icon with these rules:

Screen Shot 2019-12-13 at 2 58 07 PM

But if we remove height + width then the document flows naturally:

Screen Shot 2019-12-13 at 2 58 19 PM

This is not blocking since it doesn't affect anything but it's a habit that is not optimal. For sake of maintenance, default browser behavior is usually preferred -- from my own experience.

src/components/custom-field-input-text/custom-field-input-text.jsx
juanca: Well, let's just keep it in mind in case the engineering org decides to move in a direction or another... or both! Why not both...

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -0,0 +1,64 @@
...
+      </div>
+      <span className={styles.help}>{props.helpText}</span>
+    </div>
+  );
+}
+
+CustomFieldInputText.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
```
juanca: I just found out a cool thing (but I am not sure if it works for us):

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation

We can set the validity of an input via JavaScript. This means we could set the input to be invalid and allow for styling to be based on its pseudo state: `.input:invalid { ... }` type of thing! The method is `setCustomValidity`.

PR #61, opened 2019-12-13
Custom Field: Number Component

Bloodyaugust:
Type of work: component / site / other

Deployment URL: https://mavenlink.github.io/design-system/custom-field-input-number

(Optional for outside contributions) Tracked work in Mavenlink: https://www.pivotaltracker.com/story/show/169510683

## Motivation

Identical to the previous text input, which this component composes.

## Acceptance Criteria

Also identical to the text input, with one additional AC
1. Properly validates `Number` input strings
2. No breaking changes to the design site are introduced due to additional libraries being added and old ones being updated


juanca:
Heads up: I changed the base branch to make the diff a little less daunting 😅

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (error) {
+    return `${className} ${CustomFieldInputTextStyles.error}`;
+  }
+
+  return className;
+}
+
+export default function CustomFieldInputNumber(props) {
+  const [input, setInput] = useState('');
```
juanca: We might want to make the default value of to be `props.value`?

src/hooks/number-validator/number-validator.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+// TODO: Figure out how to write tests for this, write them
+export default function useNumberValidator(input) {
+  const [valid, setValid] = useState(true);
+
+  useEffect(() => {
+    if (input === '') {
+      setValid(true);
+    } else {
+      setValid(!!parseFloat(input));
```
juanca: This doesn't work with `123123asda1` as an input.

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,67 @@
...
+  className: CustomFieldInputTextStyles['custom-field-input-text'],
+  disabled: false,
+  error: false,
+  helpText: undefined,
+  id: undefined,
+  name: undefined,
+  onClick: () => {},
+  placeholder: undefined,
+  required: false,
+  useValidator: useNumberValidator,
```
juanca: This is pretty cool! I wonder how others can use the existing hook through composition.

juanca:
What do we think about using `<input type="number" />`?

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,67 @@
...
+export default function CustomFieldInputNumber(props) {
+  const [input, setInput] = useState('');
+  const valid = props.useValidator(input);
+
+  return (
+    <CustomFieldInputText
+      className={getRootClassName(props.className, props.error, props.disabled)}
+      disabled={props.disabled}
+      error={!valid}
+      helpText={props.helpText}
```
juanca: I wonder if there is a default error message for non-numeric values. But if we just hook up the backend responses then we probably don't need to worry about it.

src/hooks/number-validator/number-validator.jsx
Bloodyaugust: Yeah, `parseFloat` wasn't cutting it. There's a whole new regex in place now.

src/components/custom-field-input-number/custom-field-input-number.jsx
Bloodyaugust: Agreed! I'm sure it will be relevant, and even if it isn't used, now we can test the hook in isolation.

src/components/custom-field-input-number/custom-field-input-number.jsx
Bloodyaugust: I worry about having a default error message because we currently have no `i18n` in MDS. I wonder if that's a discussion we should have? Maybe it does have a place here.

juanca:
Some interesting output from `yarn start`: 

> Warning: src/hooks/number-validator/number-validator.jsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.
> It usually happens when using third-party libraries, see possible solutions here:
https://react-styleguidist.js.org/docs/thirdparties.html

We might want to get ahead of this message and squash it! It might also direct us in how we're dealing with hooks as a library in our MDS site.

Bloodyaugust:
> Some interesting output from `yarn start`:
> 
> > Warning: src/hooks/number-validator/number-validator.jsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.
> > It usually happens when using third-party libraries, see possible solutions here:
> > https://react-styleguidist.js.org/docs/thirdparties.html
> 
> We might want to get ahead of this message and squash it! It might also direct us in how we're dealing with hooks as a library in our MDS site.

I was thinking of squashing it as well. It complains because it expects a fully-qualified react component function to be exported, which is not a custom hook. I couldn't find any hints in their documentation as to how this use case can be more explicitly supported. 😭

juanca:
It might just mean we need to find a better library than Styleguidist for MDS Site v2 ;)

src/components/custom-field-input-number/custom-field-input-number.jsx
juanca: Good point. I would prefer to just use backend error messaging for now.

package.json
```diff
@@ -55,7 +58,8 @@
     "react": "^16.7.0",
     "react-dom": "^16.7.0",
     "react-ga": "^2.5.6",
-    "react-styleguidist": "^8.0.6",
+    "react-styleguidist": "9.0.0",
```
juanca: We'll want to include AC about our MDS site in general since we're upgrading Styleguidist a major version.

juanca: Also, unless we have a specific reason for doing so: we want to allow for any MINOR and PATCH upgrades to our dependencies.

Something like: `^9` or `^9.0.0`

package.json
```diff
@@ -36,12 +36,15 @@
...
+    "@testing-library/jest-dom": "^4.2.4",
+    "@testing-library/react": "^9.4.0",
+    "@testing-library/react-hooks": "^3.2.1",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^23.6.0",
     "babel-loader": "^8.0.5",
     "css-loader": "^2.1.0",
-    "cypress": "^3.1.4",
+    "cypress": "3.8.0",
```
juanca: Also, unless we have a specific reason for doing so: we want to allow for any MINOR and PATCH upgrades to our dependencies.

Something like: `^3.8`. This would set a lower bound to what we think we need now and an upper bound to future "backwards compatible" versions so our dependencies tree is a bit more loose.

package.json
```diff
@@ -36,12 +36,15 @@
     "@babel/polyfill": "^7.4.4",
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
+    "@testing-library/jest-dom": "^4.2.4",
```
juanca: We might want to consider:

> We also recommend installing the jest-dom eslint plugin which provides auto-fixable lint rules that prevent false positive tests and improve test readability by ensuring you are using the right matchers in your tests

As a separate PR (or same, up to you!). I think it could be valuable enough if it actually spots false positive tests.

setup-enzyme.js
```diff
@@ -1,5 +1,6 @@
 import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import '@testing-library/jest-dom/extend-expect.js';
```
juanca: I currently do not have a preference but we can always whitelist which matchers we want:

```
import {toBeInTheDocument, toHaveClass} from '@testing-library/jest-dom'

expect.extend({toBeInTheDocument, toHaveClass})
```

package.json
juanca: One of the cool things about this library in particular is that incorporates ARIA attributes in its matchers. For example: `toBeRequired()` check `required` or `aria-required` attributes. Pretty cool.

src/hooks/number-validator/number-validator.test.jsx
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  it('passes validation for natural numbers', () => {
+    const { result } = renderHook(() => useNumberValidator('123'));
+
+    expect(result.current).toBe(true);
+  });
+
+  it('passes validation for rational numbers', () => {
+    const { result } = renderHook(() => useNumberValidator('0'));
```
juanca: I think we're thinking of numbers like 1.5 or 3/2. But this might just be redundant since we already have a test for floating numbers below.

src/hooks/number-validator/number-validator.test.jsx
```diff
@@ -0,0 +1,94 @@
...
+    expect(result.current).toBe(false);
+  });
+
+  it('fails validation for multiple . characters', () => {
+    const { result } = renderHook(() => useNumberValidator('1.2.3'));
+
+    expect(result.current).toBe(false);
+  });
+
+  it('fails validation for incomplete decmals', () => {
```
juanca:
```suggestion
  it('fails validation for incomplete decimals', () => {
```

src/hooks/number-validator/number-validator.test.jsx
```diff
@@ -0,0 +1,94 @@
...
+    const { result } = renderHook(() => useNumberValidator('1.'));
```
juanca: Is this normal? I would think that if `.1` is valid then `1.` should be valid. It would just assume a 0 in the incomplete part? I'm not sure what is the correct thing to do but this is most likely a design question since we are talking about "what does a user expect"

src/hooks/number-validator/number-validator.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect, useState } from 'react';
+
+const numberRegex = RegExp(/^-?(0|[1-9]\d*)?(\.\d+)?(?<=\d)$/);
```
juanca: I might be overthinking it but there are places in the world that use `,` as a delimiter in their numbers. Do we support that?

Examples:

```
Many European countries use a comma in place of the decimal point and use periods or blank spaces to separate every third digit.

United States: 2,367.48 francs

France: 2.367,48 francs or 2 367,48 francs
```

juanca: This also plays into brower's locale functionality. I think `type="number"` might support locales but we do have problem with invalid input and the empty string for `onChange`. 🤔 Though, I just had a thought: do we care about `onChange` if we are using browser built-in validations for `type=number`?

juanca: Oh, I just found something cool that we might be able to leverage: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

I think if we leverage either `type="number"` or `window.Intl` then we might be able to accomplish a desirable user experience. We can play with both and ask design for their recommendation.

src/hooks/number-validator/number-validator.test.jsx
Bloodyaugust: @annemnichols you wanna weigh in ^here?

src/hooks/number-validator/number-validator.test.jsx
Bloodyaugust: Yeah, the tests are pretty thorough, so if you think they're good, I do too. :D

setup-enzyme.js
Bloodyaugust: If we did it this way, I'd prefer to include it at a per-test level.

package.json
Bloodyaugust: Created a new issue to track this effort: https://www.pivotaltracker.com/story/show/170393787

package.json
Bloodyaugust: @juanca where should that new AC go, here in the PR, in the pivotal issue, or both?

src/hooks/number-validator/number-validator.jsx
Bloodyaugust: @juanca this regex does not support the European delimiters. Is that something we need to support?

The problem with the built-in browser validations is that they're not very strict, it'd be pretty easy to have bad input that could be caught on the frontend before going to the backend at all.

I'll have a look at that link on the morrow.

juanca: I'm not sure if we need to support it. What does the existing custom fields do? Is our bigmaven backend smart enough to parse out different formats? This might also play into it. But my intuition is to support varying formats since our application is used around the globe.

Bloodyaugust: Fair enough, seems like some good product/design questions for @annemnichols. 😉

package.json
```diff
@@ -55,7 +58,8 @@
     "react": "^16.7.0",
     "react-dom": "^16.7.0",
     "react-ga": "^2.5.6",
-    "react-styleguidist": "^8.0.6",
+    "react-styleguidist": "^9.0.0",
+    "react-test-renderer": "^16.12.0",
```
juanca: We need to merge `master` into this branch. Looks like this dependency version is wrong. `react`, `react-dom`, and `react-test-renderer` need to match on their versions. I think this is fixed in `master` and I found I didn't need to mess with other dependencies. I think the only thing that is missing from `master` is `@testing-library/react-hooks`, so I am not sure if that messes with these other testing libraries.

setup-enzyme.js
```diff
@@ -1,5 +1,6 @@
 import { configure } from 'enzyme';
```
juanca: I ended up deprecating enzyme in `master`.

setup-enzyme.js
juanca: This should now live in a different in `master`.

src/components/custom-field-input-number/custom-field-input-number.md
```diff
@@ -0,0 +1,28 @@
+This component enters an error state if the input is not parse-able as `float`.
```
juanca:
```suggestion
This component enters an error state if the input is an invalid `float`.
```

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,45 @@
...
+describe('CustomFieldInputNumber', () => {
+  it('has defaults', () => {
+    const tree = renderer.create((
+      <CustomFieldInputNumber />
+    )).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  describe('prop-forward API', () => {
+    it('forwards all props accepted by CustomFieldInputText', () => {
```
juanca: Looking over this test a second time... I'm not entirely confident it covers the intention that the props are forwarded. 

I think for _us_, it makes sense that the props on each interface. But if we fast-forward a year or two into the future, a new engineer might not know that this is suppose to cover the intention of forwarding the props. 

I am not sure if is possible to have a simple test for this -- especially with `@testing-library/react`. Usually, we can just expect the props from `CustomFieldInputNumber` to be on the instance of `CustomFieldInputText ` but the testing library wants tests written from the perspective of user interactions. We might want to use `react-test-renderer` to see the props being forwarded?

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+CustomFieldInputNumber.defaultProps = {
+  className: CustomFieldInputTextStyles['custom-field-input-text'],
+  disabled: false,
+  error: false,
+  helpText: undefined,
+  id: undefined,
+  name: undefined,
+  onChange: () => {},
+  onClick: () => {},
```
juanca: `onClick` can be `undefined` since we're not using it in our function component.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -57,8 +69,10 @@ CustomFieldInputText.defaultProps = {
   helpText: undefined,
   id: undefined,
   name: undefined,
+  onChange: () => {},
   onClick: () => {},
   placeholder: undefined,
   required: false,
+  type: 'text',
```
juanca: We might want to drop this if we're going to stick with `text`

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -44,9 +54,11 @@ CustomFieldInputText.propTypes = {
   helpText: PropTypes.string,
   id: PropTypes.string,
   name: PropTypes.string,
+  onChange: PropTypes.func,
   onClick: PropTypes.func,
   placeholder: PropTypes.string,
   required: PropTypes.bool,
+  type: PropTypes.string,
```
juanca: If we don't drop this, we will want to restrict the values to a limited set. I've seen misuses of `type` when we were intending a text-like field as a checkbox :(

package.json
Bloodyaugust: It is messing with it, `renderer` was too far behind to work with `hooks` included. I believe I've fixed the deps and lockfile after quite a bit of faffing about. 😅

src/components/custom-field-input-text/custom-field-input-text.jsx
Bloodyaugust: I think we probably still want to give users the option to customize this, especially if for whatever reason the text being entered is sensitive and needs `password`.

src/components/custom-field-input-text/custom-field-input-text.jsx
Bloodyaugust: Do we want to work with @annemnichols to define the valid set of text?

Bloodyaugust:
> A couple of revisions. A major test question. And I think we need to ping the product team IRL instead of Github. Let me know if I should coordinate a session -- or we could discuss these things in a team meeting.

I think Anne and I can at least start the conversations needed here on Monday, when we start in on https://www.pivotaltracker.com/story/show/170522065. Perhaps we'll need another story/meeting after that, and we can tackle that then?

src/hooks/number-validator/number-validator.test.jsx
Bloodyaugust: Anne: `1.` should be valid as well.

src/components/custom-field-input-text/custom-field-input-text.jsx
Bloodyaugust: @juanca I'm just now realizing I mis-interpreted this statement, and I definitely agree. For these inputs, we can probably just do `date||number||text`, does that sound okay?

src/hooks/number-validator/number-validator.jsx
Bloodyaugust: After talking this over with Anne and asking some questions of the fine folks over at M-Who, it seems like we could, do, and probably should support European-style delimiters. I'll put that in.

package.json
juanca: Looks good. I thought `react` and any `react-` libraries were supposed to have their versions the exact same. But I couldn't find much to back that up. I just have a recollection of React devs encouraging or fixing bugs by doing that. I guess we should just let dependency validation do its own magic.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    it('renders all forwarded props except event handlers', () => {
+      const tree = renderer.create((
+        <CustomFieldInputNumber
+          className={'test'}
+          disabled={false}
+          error={true}
+          helpText={'help'}
+          id={'id'}
```
juanca: Let's avoid having non-unique strings in our values -- especially when we are stringifying the whole JSON tree.

The current `stringTree` has `... id='id' ...` in it and the test asserts the the characters `id` is somewhere in there. If the implementation were to use some other property value than the provided ID then this would be a false positive.

We will want to set the value to something unique and assert that we find it. Or we want to traverse the JSON tree with keys and assert values -- in that case, it is okay to be generic since we will not incorrectly match against a key instead of a value.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+          name={'name'}
```
juanca: See concern with `id`

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+          placeholder={'placeholder'}
```
juanca: See concern with `id`

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+      expect(tree.props.className).toContain('test error');
+      expect(tree.props.className).not.toContain('disabled');
+
+      expect(stringTree).toContain('help');
+      expect(stringTree).toContain('id');
+      expect(stringTree).toContain('name');
+      expect(stringTree).toContain('placeholder');
+      expect(stringTree).toContain('Required');
+      expect(stringTree).toContain('type');
```
juanca: Looks like we want to assert the value of `type` instead of the key being present.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+          className={'test'}
+          disabled={false}
+          error={true}
+          helpText={'help'}
+          id={'id'}
+          name={'name'}
+          placeholder={'placeholder'}
+          required={true}
+          type={'text'}
+          value={'value'}
```
juanca: See concern with `id`

src/components/custom-field-input-text/custom-field-input-text.jsx
juanca: Yeah, at the very least those are text-like. :+1:

src/hooks/number-validator/number-validator.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect, useState } from 'react';
+
+const numberRegex = RegExp(/^-?(0|[1-9]\d*)?[,.]?(\d+)?$/);
```
juanca: I'm not sure this is a blocker. But I was playing around with this and found this odd:

| number | valid? |
| --- | --- |
| 0 | true |
| 00 | true |
| 001 | true |
| 001. | false |
| 1. | true |

Very unexpected. But also unlikely we care. I think this is a kind of thing to throw to QA and product.

juanca: I also found it odd that adding leading or trailing space breaks the input as well.

package.json
juanca: Both! We want it here to lead by example (as an open source library). We want it in pivotal for our QA fellows.

src/hooks/number-validator/number-validator.jsx
juanca: I'm not sure I see this bits of code. Let me know where to look.

package.json
Bloodyaugust: Excellent, that's what has been done. :D

src/components/custom-field-input-number/custom-field-input-number.test.jsx
Bloodyaugust: Yeah, I was taking a shortcut, but I'll make the properties unique and search for the whole property. 😄

src/hooks/number-validator/number-validator.jsx
Bloodyaugust: Makes sense that space characters are invalid, perhaps we could trim them or explicitly allow them in the regex. Not sure why 0 padded numbers with an incomplete decimal are triggering a false alarm tho, I'll look into it.

juanca:
@Bloodyaugust Can you review this? I did my own pass as well.

jest.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   },
   setupFilesAfterEnv: [
     '@testing-library/jest-dom/extend-expect.js',
+    '<rootDir>/jest.config.setup.files.after.env.js',
```
Bloodyaugust: `<rootDir>` is a really cool pattern, news to me! 😄

jest.config.setup.files.after.env.js
```diff
@@ -0,0 +1,7 @@
+console.error = (message) => { // eslint-disable-line
+  // During test development, it might be useful to comment out
```
Bloodyaugust: It's probably a good idea to leave this note in one of our README or doc files, to make it more discoverable for new contributors. Or maybe a reference to this file?

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <CustomFieldInputText
+      className={getRootClassName(props.className, invalid, props.disabled)}
+      disabled={props.disabled}
+      error={invalid}
+      id={props.id}
+      inputRef={inputRef}
+      label={props.label}
+      max={2 ** 31} // This value is sourced from the API
```
Bloodyaugust: Could we make these named constants at the top of the file, with names that imply their source/purpose?

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -0,0 +1,72 @@
...
+      min={-(2 ** 31)} // This value is sourced from the API
+      name={props.name}
+      onKeyUp={handleOnKeyUp}
+      placeholder={props.placeholder}
+      required={props.required}
+      step={1}
```
Bloodyaugust: Great addition 👍

src/components/custom-field-input-number/custom-field-input-number.md
```diff
@@ -0,0 +1,31 @@
...
+<CustomFieldInputNumber
+  disabled
+  id="test-id-2"
+  label="Example 2"
+  value={1337}
+/>
+```
+----
+##### Error:
+```jsx
```
Bloodyaugust: Maybe a note here about why this is an error, to bring to the fore the `step` config?

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    it('is valid on a decimal integer', () => {
+      render(<TestComponent value={1.00} />);
+      expect(screen.getByTestId('custom-field-input')).not.toHaveClass('error');
+    });
+
+    it('is invalid on a decimal number', () => {
+      render(<TestComponent value={1.01} />);
+      expect(screen.getByTestId('custom-field-input')).toHaveClass('error');
+    });
```
Bloodyaugust: We should probably have some additional tests the cover incorrect usage of radix and signum. I'm also curious about how we could test the correct detection and allowance of alternate radix based on localization settings.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -39,7 +39,7 @@ export default function CustomFieldInputText(props) {
   });
 
   return (
-    <div className={getRootClassName(props.className, props.error, props.disabled)}>
+    <div className={getRootClassName(props.className, props.error, props.disabled)} data-testid="custom-field-input" >
```
Bloodyaugust: Should the `testid` here by dynamic to support testing of multiple input elements? Maybe combine with `id`, or some auto-generated UUID?

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -73,20 +77,38 @@ CustomFieldInputText.propTypes = {
...
   label: PropTypes.string.isRequired,
+  max: PropTypes.number,
+  min: PropTypes.number,
   name: PropTypes.string,
+  onKeyUp: PropTypes.func,
   placeholder: PropTypes.string,
   required: PropTypes.bool,
-  value: PropTypes.string,
+  step: PropTypes.number,
+  type: PropTypes.oneOf([
```
Bloodyaugust: 👍 Nice and explicit about what kinds we support.

src/components/custom-field-input-text/custom-field-input-text.test.jsx
```diff
@@ -99,6 +130,25 @@ describe('CustomFieldInputText', () => {
...
+    it('can be set to `number`', () => {
+      render(<TestComponent type="number" />);
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('type', 'number');
+    });
+
+    it('can be set to `text`', () => {
+      render(<TestComponent type="text" />);
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('type', 'text');
+    });
+  });
```
Bloodyaugust: Seems like we should have a fail case tested for here as well.

jest.config.setup.files.after.env.js
juanca: Done!

src/components/custom-field-input-number/custom-field-input-number.jsx
juanca: Done!

src/components/custom-field-input-number/custom-field-input-number.md
juanca: The error state explanation is in the file, a bit above. I am not sure whether to keep the description in the same section of the page or separate it out.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
juanca: I looked into this and I think we're covered by the prop-type and the number element type.

- Anything string is going to be an invalid prop-type. I am not sure about testing it since it is already expected to use the component with a real number value. 
- `--1` is invalid because it is not a real number and thus breaks in the JS token parsing (cannot test this)
- Any radix besides `.` is invalid in the number element type and it not interpreted. I think I can do a bit more googling on this but it might be best to delegate all number validations and stuff to the native browser behavior.

src/components/custom-field-input-text/custom-field-input-text.jsx
juanca: I am not sure. I didn't have a use-case for that. I am not sure if tests with multiple custom field inputs should be using IDs to assert on the containers. I think instead those tests should be asserting on the state of their forms.

src/components/custom-field-input-text/custom-field-input-text.test.jsx
juanca: I'll repeat what I said in the other similar feedback for the sake of posterity: I don't think we should be testing invalid prop-types. Otherwise, we will be adding a lot of tests for the sake of making sure prop-types are working. Though, I do like that we are testing every (reasonable) valid prop-type since we care about ensuring those behaviors are working.

juanca:
> Another potential blocker is that the styleguidist version (or perhaps another dependency) has changed since the hook implementation was nuked in favor of an example hook, and now doesn't work with hooks again. If we plan to use hooks in the currency implementation, we should fix that (which can happen as part of this PR or currency). If we're not planning on using hooks yet, maybe we should just tear them out completely for now?

Oh! I forgot to check that! I had a bad state for my node_modules while I was cleaning things up -- and probably checked it when I had the later version of styleguidist. Good catch :+1:

I am going to remove them and introduce the custom hooks in another PR to keep things more tidy.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
Bloodyaugust: I agree that delegating control of behavior to the browser makes sense... if the browser functionality is in line with our own AC. I know it wasn't ever officially a part of the AC, but it seems like if we do need internationalization in the form of locale-specific radix, then we should probably implement some sort of workaround.

src/components/custom-field-input-number/custom-field-input-number.md
Bloodyaugust: Oh, I missed that bit! Could we change the wording to something like "a positive or negative number evenly divisible by `step`"?

src/components/custom-field-input-number/custom-field-input-number.test.jsx
juanca: I found this snippet interesting:

> This specification does not define what user interface user agents are to use; user agent vendors are encouraged to consider what would best serve their users' needs. For example, a user agent in Persian or Arabic markets might support Persian and Arabic numeric input (converting it to the format required for submission as described above). Similarly, a user agent designed for Romans might display the value in Roman numerals rather than in decimal; or (more realistically) a user agent designed for the French market might display the value with apostrophes between thousands and commas before the decimals, and allow the user to enter a value in that manner, internally converting it to the submission format described above.

I think we're good there!

src/components/custom-field-input-number/custom-field-input-number.md
juanca: It would be "evenly divisible by 1". Or a multiple of 1. I think both sound kinda weird? It feels like "integer" or "whole integer" is more descriptive.

src/components/custom-field-input-number/custom-field-input-number.test.jsx
Bloodyaugust: Awesome, so it's up to the browser to implement the spec here. I wonder if Chrome does... 🤔 Also interesting was the snippet a bit lower on when it is appropriate to use a number input, especially regarding input that uses numbers but is not _actually_ a number, like credit cards.

src/components/custom-field-input-number/custom-field-input-number.md
Bloodyaugust: With the default `step` of 1, yes, but we allow `step` to be configured. Either way, this definitely isn't a blocker, so I'll leave it up to you.

Bloodyaugust: Oh wow, except I'm totally wrong! Maybe we should allow `step` to be configured, but I see now that it currently isn't. 😅 Again, I'll leave that decision up to you, I'm not sure if we _want_ to support a use case for this input where `step !== 1`.

juanca:
@Bloodyaugust I think I hit every single comment. Lemme know how it is holding up.

src/components/custom-field-input-number/custom-field-input-number.md
juanca: For now, I'm going to leave it as a _no_ for supporting more than our Custom Field API requirements. 

But that might change if we go down the path of a generic _Number_ component -- which I am liking more and more er' day.

Bloodyaugust:
@juanca Looks good to me! If QA is happy, let's merge this.

PR #63, opened 2019-12-27
New testing patterns: Button (and installing libraries)

juanca:
Type of work: other

Deployment URL: https://mavenlink.github.io/design-system/new_testing_patterns

## Motivation

## Acceptance Criteria

CI tests are passing.

### DIFF
--- a/src/components/button/button.test.jsx
+++ b/src/components/button/button.test.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
+import { render, fireEvent, screen } from '@testing-library/react';
 import Button from './button.jsx';
 
 describe('Button', () => {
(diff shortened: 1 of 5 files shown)


JustinRyanH:
Why screen over `const { getByText } = render(...)`?

juanca:
The examples on the [README](https://github.com/testing-library/react-testing-library) use `screen`. 

I'm trying to look for materials discussing pros and cons of `screen` versus generated bounded query functions. Intuitively, I prefer `screen` since it would be how a user interacts with a UI.

JustinRyanH:
They should both be functionally the same, I think `screen` is something they've added within the last couple of weeks

src/components/button/button.test.jsx
```diff
@@ -14,136 +15,136 @@ describe('Button', () => {
...
     it('can be "reset"', () => {
-      const tree = renderer.create((
+      render((
         <Button type="reset">
           Hello world!
         </Button>
-      )).toJSON();
-      expect(tree.props.type).toEqual('reset');
+      ));
+      expect(screen.getByText('Hello world!')).toHaveAttribute('type', 'reset');
```
cbacon91: I think generally speaking the `testing-library` wants to promote things a user can observe.

A user can't observe the type of the button, but they can observe its behavior. What does a `reset` button actually do? Does it have a specific class? Does it do something different on click?

src/components/button/button.test.jsx
```diff
@@ -14,136 +15,136 @@ describe('Button', () => {
...
     it('prioritizes className prop', () => {
-      const tree = renderer.create((
+      render((
         <Button className="prioritize-me" color="primary">
           Hello world!
         </Button>
-      )).toJSON();
-      expect(tree.props.className).toEqual('prioritize-me');
+      ));
+      expect(screen.getByText('Hello world!')).toHaveClass('prioritize-me');
```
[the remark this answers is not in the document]
cbacon91: I see this was already asked and I was too late, disregard

src/components/button/button.test.jsx
juanca: The `submit` and `reset` types are related to `<form>`s. If we were to add a form then we could test the side effects. I think that would be more of an integration test and it would be more applicable when using the component in a host application. 

In our world, we would have a test in our Mavenlink application repository which asserts that the actual form is being reset when clicking the button. 

I think this might be a subtle difference between `@testing-library/react` and `react-test-renderer`. We want as a library to test the direct API changes so we're perhaps going against the grain in our test files.

PR #65, opened 2019-12-30
New testing patterns: SVG 

juanca:
Type of work: other

Deployment URL: https://mavenlink.github.io/design-system/new_testing_patterns_2

## Motivation

## Acceptance Criteria

- CI tests are passing
- MDS Site SVGs look good (slight logic simplification)

### DIFF
--- a/src/components/icon/icon.test.jsx
+++ b/src/components/icon/icon.test.jsx
@@ -39,7 +39,6 @@ describe('Icon', () => {
           )).toJSON();
           expect(tree.props.className).toContain('icon-base');
           expect(tree.props.className).toContain(`fill-${color}`);
-          expect(tree.props.className).not.toContain('stroke');
         });
       });
     });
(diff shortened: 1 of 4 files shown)


src/components/icon/icon.test.jsx
```diff
@@ -11,80 +12,144 @@ describe('Icon', () => {
   });
 
   it('has title', () => {
-    const tree = renderer.create((
+    render((
       <Icon name="foobar" title="yo" />
-    )).toJSON();
-    expect(tree).toMatchSnapshot();
+    ));
+    expect(screen.getByRole('img')).toContainHTML('<title>yo</title>');
```
cbacon91:
```suggestion
    expect(screen.getByTitle('yo')).toExist();
```

juanca: ❗️ 

Let me see if it will work. I forgot about that particular query during development.

PR #66, opened 2019-12-30
Add "autoFocus" and "ref" Attributes To Input Component

ade713:
Type of work: component / accessibility

Deployment URL: https://mavenlink.github.io/design-system/ds-add-ref-attribute-to-input-component

## Motivation

Adding the `ref` attribute, in order to enable auto-focusing on the input element of the Input component.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,9 @@ module.exports = {
       ignorePackages: true,
     }],
   },
+  env: {
+    browser: true,
+  },
   overrides: [{
     files: [
       '**/*.test.js',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,9 +6,6 @@ module.exports = {
       ignorePackages: true,
     }],
   },
-  env: {
-    browser: true,
-  },
   overrides: [{
     files: [
       '**/*.test.js',
--- a/src/components/input/input.jsx
+++ b/src/components/input/input.jsx
@@ -22,6 +22,7 @@ export default function Input(props) {
       </label>
       <div className={props.cssContainer}>
         <input
+          autoFocus={props.autoFocus} // eslint-disable-line jsx-a11y/no-autofocus
           className={getClassName(props.className, props.invalid)}
           disabled={props.disabled}
           id={props.id}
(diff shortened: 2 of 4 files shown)


src/components/input/input.jsx
```diff
@@ -50,6 +51,10 @@ Input.propTypes = {
   cssLabel: PropTypes.string,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
+  inputRef: PropTypes.oneOfType([
```
juanca: I think our React version in our Mavenlink application supports the latest ref types. If that is the case, I would like to avoid supporting the function-based refs to have more reason to move our codebase forward:

In addition, a "ref" is actually meant to just be a simple pointer -- the way we use it implies HTML elements already. This is actually a bit more universal (think React Native in the future) and acceptable:

```
inputRef: PropTypes.shape({ current: PropTypes.any }),
```

.eslintrc.js
```diff
@@ -6,6 +6,9 @@ module.exports = {
       ignorePackages: true,
     }],
   },
+  env: {
```
juanca: If we refactor our `inputRef` type then we should drop this.

src/components/input/input.test.jsx
```diff
@@ -296,4 +297,20 @@ describe('Input', () => {
       expect(tree.findByType('input').props.value).toEqual('test-value');
     });
   });
+
+  describe('ref API', () => {
+    it('is set to the inputRef property', () => {
```
juanca: This is just a note for future devs: we're moving away from `react-test-renderer` and `enzyme`. How would we test this with `@testing-library/react`?

juanca:
I think exposing a ref on the API is good. However, I have a thought:

> Adding the ref attribute, in order to enable auto-focusing on the input element of the Input component.

Does adding a `autofocus` attribute on the `<input>` element work for the intended case? (I actually don't know if it would do the right thing but it might be worth checking that out -- and we might want to also expose it as an API (if it does work)).

src/components/input/input.jsx
ade713: Change made. 
We were thinking of being specific towards the input html element since that's what it was being used for, and not expecting a reference to another element type.

src/components/input/input.test.jsx
ade713: React testing-library uses deep rendering by default, so no special additions would be required to check/test the prop values.

ade713:
@juanca So we tested the `autoFocus` and it works right now. We'll leave the `ref` addition and tag on with the `autofocus` since both are relevant.

PR #68, opened 2019-12-31
New testing patterns: Input

juanca:
Type of work: other

Deployment URL: https://mavenlink.github.io/design-system/new_testing_patterns_3

## Motivation

## Acceptance Criteria

- CI tests are passing

### DIFF
--- a/src/components/input/input.test.jsx
+++ b/src/components/input/input.test.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import { mount } from 'enzyme';
 import renderer from 'react-test-renderer';
+import { fireEvent, render, screen } from '@testing-library/react';
 import Input from './input.jsx';
 
 describe('Input', () => {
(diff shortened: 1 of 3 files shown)


src/components/input/input.test.jsx
```diff
@@ -16,314 +16,311 @@ describe('Input', () => {
...
           id="foo"
-          label="I am a label for accessibility"
+          label="the label"
           onFocus={onFocusSpy}
         />
-      )).root;
-      tree.findByType('input').props.onFocus();
+      ));
+      fireEvent.focus(screen.getByLabelText('the label'));
       expect(onFocusSpy.mock.calls.length).toEqual(1);
```
JustinRyanH: Not at all related, but I figured this is a good moment to mention it @juanca but I think it will speed up our development if we use test double spys over native jest spys. 

https://github.com/testdouble/testdouble-jest

Not block, nothing you gotta address. Just ping me about it if you have questions

juanca: Our philosophy is to avoid bringing in too many external dependencies. As of this writing, I haven't seen any benefit from td over spies. 

But if things change, we could do the change in another PR in the future.

PR #69, opened 2020-01-02
New testing patterns: Table

juanca:
Type of work: other

Deployment URL: https://mavenlink.github.io/design-system/new_testing_patterns_3

## Motivation

## Acceptance Criteria

CI tests are passing


juanca:
We've found success with two things:

- Testing default attributes (and lack of default attributes) on DOM elements
- Testing DOM structure

Otherwise, we need a test for every possible attribute on the DOM elements -- or we can get fancy and pluck all the applied attributes, but it is a test for every DOM element nonetheless.

PR #71, opened 2020-01-16
Custom Field: Currency

Bloodyaugust:
Type of work: component / hook

Deployment URL: https://mavenlink.github.io/design-system/custom-field-input-currency

https://www.pivotaltracker.com/story/show/169510702

## Motivation

As the other custom field input components, so goes the story of the currency input.

## Acceptance Criteria

Acceptance criteria in the tracker story linked above. 

## Change log entry

Add currency custom field input and commensurate custom hook for validation


src/components/custom-field-input-currency/currencies.js
```diff
@@ -0,0 +1,1978 @@
+/* Downloaded XML file from https://www.currency-iso.org/en/home/tables/table-a1.html and converted to JSON */
```
cincospenguinos:
```suggestion
/* Downloaded XML file from https://www.currency-iso.org/en/home/tables/table-a1.html and converted to javascript */
```

src/components/custom-field-input-currency/custom-field-input-currency.jsx
```diff
@@ -0,0 +1,148 @@
...
+  );
+}
+
+CustomFieldInputCurrency.propTypes = {
+  className: PropTypes.string,
+  currencyCode: CurrencyCodeType,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
+  helpText: PropTypes.string,
+  id: PropTypes.string,
```
cincospenguinos:
```suggestion
  id: PropTypes.string.isRequired,
```

src/components/custom-field-input-currency/custom-field-input-currency.md
```diff
@@ -0,0 +1,62 @@
...
+----
+##### Disabled:
+```jsx
+<CustomFieldInputCurrency
+  disabled={true}
+  helpText="This input is disabled."
+  id="test-id-2"
+  name="test-name-2"
+  value={123}
+  label="money"
```
juanca:
```suggestion
  label="Disabled Currency"
```

src/components/custom-field-input-currency/custom-field-input-currency.md
```diff
@@ -0,0 +1,62 @@
...
+Errors can be classified as technical or contextual. The first example below is a technical error--cFa cannot be
+separated below a single cFa. The second, on the other hand, is contextual--perhaps the backend denied the value
+presented to it, responding with a specific message for the context of that specific currency value.
+```jsx
+<CustomFieldInputCurrency
+  id="test-id-3"
+  name="test-name-3"
+  value={1.12}
+  currencyCode="XAF"
+  label="money"
```
juanca:
```suggestion
  label="Internal Error State"
```

src/components/custom-field-input-currency/custom-field-input-currency.md
```diff
@@ -0,0 +1,62 @@
...
+/>
+
+<CustomFieldInputCurrency
```
juanca: This one is missing a label.
```suggestion
<CustomFieldInputCurrency
  label="External Error State"
```

juanca: In this note, let's open the dev tools on this site's page and check for errors -- it should have been a prop validation error.

src/components/custom-field-input-currency/custom-field-input-currency.md
```diff
@@ -0,0 +1,62 @@
+This component may be used to express currency. Provide a numeric value to the component along with a currency code, and
```
juanca: We need a link to our list of supported currencies:

https://github.com/mavenlink/design-system/blob/master/src/components/custom-field-input-currency/currencies.js

juanca:
@cincospenguinos I added a few more things as comments.

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -35,6 +35,8 @@ export default function CustomFieldInputNumber(props) {
     setInvalid(!inputRef.current.validity.valid);
   });
 
+  const value = props.value || 0;
```
juanca: Just a note: we should prefer default props! 

Though, I'm looking into this behavior and I think I would prefer another solution. The reason being: While the UX is near identical, the component now will always have a value. We cannot determine whether a user typed the value `0` or left it `undefined` -- and thus we will always be storing a value in the "backend." I'm gonna timebox some time and see if we can figure out a way to have it as undefined as well.

juanca:
> Would it be possible for us to expose it for other components and still avoid its usage here?

I would think it would have the same problems as currency. Whenever there is an invalid value in the number element, it will expose an empty string as a value even though the user typed something in. We don't have to parse anything on the change event, we only have to parse when on the blur event -- as far as I can tell -- otherwise we are still editing.

PR #74, opened 2020-02-05
Upgrade Jest (and a lot of other things)

juanca:
Type of work: other

Deployment URL: https://mavenlink.github.io/design-system/upgrade-jest

## Motivation

In order to have a more reliable test environment, we want to upgrade our `jsdom` version that is used in Jest. In another PR, we are getting invalid tests due to a bad validation algorithm in their `HTMLInputElement-impl.js#stepMismatch`.

In order to complete this upgrade, we need to upgrade a bunch of other things. 

1. Upgrade Node to latest in major version 10 (we are using 8.11.1 in Bigmaven)
2. Upgrade Jest to latest

In order to minimize the friction between developing on MDS and Bigmaven, I encourage other contributors to install the `n` package globally. I am not sure if we can set it up so `n` automatically switches to the right version while working within this repository. 

## Acceptance Criteria

- Tests run and pass.
- MDS Site is functional

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ commands:
 jobs:
   build_and_test:
     docker:
-      - image: cypress/base:8
+      - image: cypress/base:8.16.0
     working_directory: ~/repo
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
       - checkout
       - install_dependencies:
           image: "cypress"
-          version: "10"
+          version: "10.18.0"
       - run: yarn lint
       - run: yarn test
       - run:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -55,13 +55,13 @@ jobs:
...
     docker:
-      - image: circleci/node:8.11.1
+      - image: circleci/node:10.18.0
     working_directory: ~/repo
     steps:
       - checkout
       - install_dependencies:
           image: "node"
-          version: "1"
+          version: "10.18.0"
```
Bloodyaugust: Holy wow, were we really using node v1?! 😮

Bloodyaugust: Oh wait, it seems like this is redundant, considering the image itself has node already installed? Can we remove the dependency completely?

jest.config.js
```diff
@@ -2,8 +2,9 @@ module.exports = {
   moduleNameMapper: {
     css$: 'identity-obj-proxy',
   },
-  // Key is deprecated in future version in favor of setupFilesAfterEnv
-  setupTestFrameworkScriptFile: '<rootDir>/setup-test-framework.js',
+  setupFilesAfterEnv: [
+    '@testing-library/jest-dom/extend-expect.js',
```
Bloodyaugust: 👍 nice move!

.circleci/config.yml
juanca: We were using Node 8! The version value here is just for the cache key. I decided to just keep it the same as the actual Node version for the sake of consistency (and easier understanding in the future of when we need to bust the key).

PR #81, opened 2020-03-16
Read-only Date Component

cincospenguinos:
## Motivation

In an effort to offer a well rounded set of front-end interactions with our application, a date component would be a valuable addition. This component is designed to be able to provide such an interaction.

## Acceptance Criteria

It is read-only and thus disabled
It has a datepicker icon on the right side
See screenshot below and take note of the comment!
The SVG icon is in the internal MDS -- we need to copy it over (and possibly delete it in the internal design system)
It should format the date correctly
At the moment, we want the same format as dates in bigmaven: Apr 01, 2020

## Change log entry

<details>
<summary>PR upkeep checklist</summary>
<br />

- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-dates-you-can-pick-some-can-t-be-made
- [x] (Optional) Pivotal tracker URL: [...](https://www.pivotaltracker.com/story/show/170268458)
- [x] (Mandatory) Song link: ["Things You Can Do" by Deltron the Funky Homosapien](https://www.youtube.com/watch?v=V-Easr4-qJk)

</details>


src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+    const isInputValid = inputRef.current.validity.valid;
+    const valid = isValueValid(props.value, props.error, isInputValid);
+    setIsValid(valid);
+
+    if (!valid) {
+      setIsEditing(true);
+    } else {
+      setIsEditing(false);
+    }
+  }, [inputRef.current]);
```
Bloodyaugust: We should also probably look at `props.value` and `props.error`.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+
+    props.onChange(event);
+  };
+
+  const helpText = () => {
+    if (!isValid && !isValidInput(props.value)) {
+      return `"${props.value}" is an invalid date`;
+    }
+
+    return props.helpText;
```
Bloodyaugust: Do we want to show the help text when our input is valid? My understanding is that help text is only supposed to be shown when we're in an invalid state, so `props.helpText` should be used as an override for our default, if it exists. We could give `props.helpText` a default value in fact, and treat it as a saved template string function, allowing the frontend to override our default string, but still have contextualization for their error.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  };
+
+  const sharedProps = {
```
Bloodyaugust: Is `sharedProps` around to define the props we forward to `CustomFieldInputText`?

src/components/custom-field-input-date/custom-field-input-date.md
```diff
@@ -0,0 +1,61 @@
...
+    label="Juanca's Birthday"
+    value="03/03/1990" 
+    disabled
+/>
+```
+----
+
+##### Errored state:
+
+Below are two examples of an invalid custom field input: a syntactic error and a semantic error. The first below is
```
Bloodyaugust: We should probably just let the browser wrap these lines instead of newlining them, unless we want them separated by whitespace for clarity.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -38,6 +40,27 @@ export default function CustomFieldInputText(props) {
...
+
+    if (props.icon) {
+      return props.icon;
+    }
+
+    return undefined;
+  };
+
+  const showIcon = () => {
+    return props.error || !!props.icon;
```
Bloodyaugust: I'm not sure we need to coerce `props.icon` here, considering our default is `undefined`, and type is `PropTypes.node`.

src/svgs/icon-calendar-fill.svg
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Bloodyaugust: Did we verify with @annemnichols that this icon is the correct one? Do we want to first strip it of metadata, and obfuscate? Not sure what our process is/should be here.

juanca:
@cincospenguinos Can you update the PR checklist with the deployment URL?

src/svgs/icon-calendar-fill.svg
juanca: This is the correct one (for now). Looks like the design team made designs for a new calendar icon but has not vetted it -- we might switch to the new one in the future. 

As for stripping, it's okay, in bigmaven we are using SVGo to optimize our SVGs and strip away most irrelevant data -- most of the metadata is from the Sketch library anyway and adding another step in the design process might be more headaches than SVGo. For our open-source aspect, we might just recommend using SVGo as well (or similar).

yarn.lock
```diff
@@ -11019,6 +11019,11 @@ ws@^7.0.0, ws@^7.2.1:
   resolved "https://registry.yarnpkg.com/ws/-/ws-7.2.1.tgz#03ed52423cd744084b2cf42ed197c8b65a936b8e"
   integrity sha512-sucePNSafamSKoOqoNfBd8V0StlkzJKL2ZAhGQinCfNQ+oacw+Pk7lcdAElecBF2VkLNZRiIb5Oi1Q5lVUVt2A==
 
+ws@^7.2.1:
```
juanca: This is a weird addition since nothing else changed. Did we update `ws` explicitly?

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -18,9 +18,11 @@ function getRootClassName(className, error, disabled) {
 }
 
 export default function CustomFieldInputText(props) {
-  const inputRef = props.inputRef || useRef(null);
+  const defaultRef = useRef(null);
```
juanca: For the curious: this was a bug. The `||` is a conditional and would run the hooks in different order depending on a re-render of the component (and some `key` shenanigans). Very similar to "do not run hooks conditional" then we shouldn't invoke `useRef` conditionally. Instead, we always invoke it and then conditionally choose to use its value.

src/components/custom-field-input-date/custom-field-input-date.jsx
juanca: We might want to also scope it down to: `inputRef.current.validity.valid` but I am not sure. Is the idea that we don't want to re-run this method when the element reference does not change? Or when the validity of the element does not change?

Bloodyaugust: This array is things to watch and re-run when they _do_ change, so we definitely want to know if the `inputRef.current` changes.

yarn.lock
cincospenguinos: I don't remember updating `ws` directly. Is it possible that it was included in another installation? `jest-jsdom-environment-sixteen` perhaps?

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  }, [isEditing, isFocused]);
+
+  const handleOnFocus = () => {
+    setIsFocused(true);
+    setIsEditing(true);
+  };
+
+  const handleOnBlur = () => {
+    setIsFocused(false);
+    setIsEditing(false);
```
juanca: If the value is invalid, do we want to stay in editing mode?

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  };
+
+  const handleOnChange = (event) => {
+    if (inputRef && inputRef.current) {
```
juanca: When is `inputRef` not defined? I think this might be redundant.

src/components/custom-field-input-date/format/format-date.js
```diff
@@ -0,0 +1,92 @@
...
+      return {
+        day: pieces[1].padStart(2, '0'),
+        month: month.padStart(2, '0'),
+        year: pieces[2],
+      };
+    },
+  },
+};
+
+export function formatMatching(dateString) {
```
juanca: We might not need to export this. I can't see a usage of it.

juanca:
We will want to do something about the SVG styling (let's get a design review to understand how it is suppose to look like):

Screen Shot 2020-03-17 at 3 39 26 PM

juanca:
Okay, I realized that perhaps the initial AC was a tiny bit wrong:

Let's remove the "readonly implies disabled" criteria. Let's instead use the HTML5 readonly attribute to create a "focusable but not editable" input. This is more accessible for a "read only page"

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+      return `"${props.value}" is an invalid date`;
+    }
+
+    return props.helpText;
+  };
+
+  const sharedProps = {
+    className: props.className,
+    disabled: true,
+    icon: <Icon name={calendarSvg.id} title={props.label} stroke="primary" />,
```
juanca: I am not sure if the calendar icon requires a title or if the title should be the label. This might need to be revisited in the future once we do an accessibility pass on the component.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+CustomFieldInputDate.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
+  helpText: PropTypes.string,
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  min: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
juanca: Why is there a min of type number? It looks like we're using it with only a string formatted as a date.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  max: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
juanca: Why is there a max of type number? It looks like we're using it with only a string formatted as a date.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  if (isEditing || !isValid) {
+    const value = validDate(props.value) ? convertToFormat(props.value, 'yyyy-mm-dd') : props.value;
+
+    return (<CustomFieldInputText
+      {...sharedProps}
+      error={!isValid}
+      helpText={helpText()}
+      id={props.id}
+      key={`${props.id}-editing`}
+      min={convertToFormat(props.min, 'yyyy-mm-dd')}
```
juanca: What do we think about simplifying the logic by making the `min` property to have a format `'yyyy-mm-dd'`? Same thing with `max`.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+      max={convertToFormat(props.max, 'yyyy-mm-dd')}
+      onBlur={handleOnBlur}
+      onChange={e => handleOnChange(e)}
```
juanca: As far as I know, we shouldn't be creating inline anonymous functions for event handlers. Since the `handleOnChange` function already implies handling an event, its signature should just match the event handling. 

```suggestion
      onChange={handleOnChange}
```

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+  const handleOnBlur = () => {
+    setIsFocused(false);
+    setIsEditing(false);
+  };
+
+  const handleOnChange = (event) => {
+    if (inputRef && inputRef.current) {
+      const isInputValid = inputRef.current.validity.valid;
+      const newDate = convertToFormat(event.target.value, 'yyyy-mm-dd');
+      setIsValid(isValueValid(newDate, props.error, isInputValid) && validDate(newDate));
```
juanca: Looks like `validDate` on this line is redundant since `isValueValid` already invokes it. Do we need it here?

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -0,0 +1,186 @@
...
+
+  describe('onChange API', () => {
+    it('accepts an onChange listener', () => {
+      const onChange = jest.fn();
+      const { getByLabelText } = renderComponent({ value: '', onChange });
+      changeValue(() => getByLabelText('Field Date'), '2016-07-18');
+      expect(onChange.mock.calls.length).toBe(1);
+    });
+
+    xit('changes from invalid to valid state', () => {
```
juanca: This test seems out of place.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+    type="text"
+    value={convertToFormat(props.value, 'Month dd, yyyy')}
+  />);
+}
+
+CustomFieldInputDate.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
+  helpText: PropTypes.string,
```
juanca: We're missing tests on this.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -0,0 +1,142 @@
...
+    onFocus={handleOnFocus}
+    type="text"
+    value={convertToFormat(props.value, 'Month dd, yyyy')}
+  />);
+}
+
+CustomFieldInputDate.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
```
juanca: We are missing tests on this.

src/components/custom-field-input-date/custom-field-input-date.md
```diff
@@ -0,0 +1,61 @@
+A simple date field for custom fields. Permits a variety of formats and allows custom validation.
+
+##### Basic usage:
+
+```jsx
+<CustomFieldInputDate
+    id="my-birthday-1"
```
juanca: Looks like we're diverging from the other markdown files: indentation is 1-tab which is 2-spaces. This codebase is small enough that I want to keep up some resemblance of spacing consistency.

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: This function only selects the proper help text--either "value is an invalid date" when the input itself is invalid, or the help text handed to the component. This only sets the prop--the actual logic that handles displaying help text is hidden inside of the `CustomFieldInputText`

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: `sharedProps` are props that are shared between the two components we swap between. They need the same className, disabled prop, icon, etc. This was to ensure that we aren't repeating ourselves.

src/components/custom-field-input-text/custom-field-input-text.jsx
cincospenguinos: We show an error icon if we have an error, or if we have an icon handed to us we present it. This is how we show the calendar icon

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: That will be handled in the next PR, given that this one enforces a read-only component

src/components/custom-field-input-date/format/format-date.js
cincospenguinos: I'm exporting it for testing purposes. Should I remove those specs?

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: I'd be down, but how do we enforce a format in Prop types?

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: I have tests on each of those three props now in the date component

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: See below

src/components/custom-field-input-date/format/format-date.js
juanca: Yeah, we can remove the tests if there is sufficient test coverage on usages of it -- no spying!

src/components/custom-field-input-date/custom-field-input-date.jsx
juanca: Ping!

yarn.lock
juanca: Let's debug this: Let's try:

1. Removing this then run `yarn install` and see if there is a diff
2. Also, removing this, removing `node_modules`, and then run `yarn install` and see if there is a diff

src/components/custom-field-input-date/custom-field-input-date.jsx
juanca: One of the things about prop-types is that we can provide a custom prop-type as a function:

```
  // You can also specify a custom validator. It should return an Error
  // object if the validation fails. Don't `console.warn` or throw, as this
  // won't work inside `oneOfType`.
  customProp: function(props, propName, componentName) {
    if (!/matchme/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  },
```

See the documentation for "more" information: https://reactjs.org/docs/typechecking-with-proptypes.html

yarn.lock
cincospenguinos: When asking yarn why ws:

```
➜  design-system git:(webinf-dates-you-can-pick-some-can-t-be-made) ✗ yarn why ws
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "ws"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "ws@7.2.1"
info Has been hoisted to "ws"
info Reasons this module exists
   - Hoisted from "jest-environment-jsdom-sixteen#jsdom#ws"
   - Hoisted from "jest#@jest#core#jest-config#jest-environment-jsdom#jsdom#ws"
info Disk size without dependencies: "144KB"
info Disk size with unique dependencies: "144KB"
info Disk size with transitive dependencies: "144KB"
info Number of shared dependencies: 0
=> Found "webpack-dev-server#ws@6.2.1"
info This module exists because "react-styleguidist#webpack-dev-server" depends on it.
info Disk size without dependencies: "136KB"
info Disk size with unique dependencies: "240KB"
info Disk size with transitive dependencies: "240KB"
info Number of shared dependencies: 1
✨  Done in 0.67s.
```

cincospenguinos: It looks like it's included with `jest-environment-jsdom-sixteen`

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: I'll include this in the next PR since this is a readonly component and `min` and `max` are not handled here.

cincospenguinos: See #84 for the implementation of this suggestion

src/components/custom-field-input-date/custom-field-input-date.jsx
cincospenguinos: Azur said that keeping the label as title would be ideal

juanca:
I will make a new PR for the changelog and stuff

PR #92, opened 2020-04-17
Modify the CustomFieldInputCurrency value API to only accept subunits

juanca:
## Motivation

This allows our component to align with the Mavenlink API. We normally are dealing with money through subunits. Let us also keep the component with similar behavior to simplify reasoning (and thus simplify code). This is a major change but thankfully we are in beta still.

For example: if we want to represent $13.37 then we will pass in 1337 as a value and a currency code of USD. Our component already knows how to manipulate the cents because we are using some ISO standard and we are pulling that specification into our codebase.

## Acceptance Criteria

We can use the component by passing in cents and getting dollars (with decimals) -- depending on the currency code. 

<details>
<summary>PR upkeep checklist</summary>
<br />

- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/currency-subunit-to-unit
- [x] (Optional) Pivotal tracker URL: https://www.pivotaltracker.com/story/show/172343217

</details>

### DIFF
--- a/src/components/custom-field-input-currency/custom-field-input-currency.jsx
+++ b/src/components/custom-field-input-currency/custom-field-input-currency.jsx
@@ -27,35 +27,36 @@ function getLocale() {
   return 'en-IN';
 }
 
-function initialInputValid(inputValue, currencyCode) {
-  const maximumFractionDigits = currencyMetaData[currencyCode].maximumFractionDigits;
-
+function initialInputValid(inputValue) {
   if (!inputValue) {
     return true;
   }
 
-  const splitInputValue = inputValue.toString().split('.');
-  if (splitInputValue.length === 1) {
+  if (inputValue.toString().split('.').length === 1) {
     return true;
   }
 
-  return splitInputValue[1].length <= maximumFractionDigits;
+  return false;
 }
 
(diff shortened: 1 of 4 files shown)


src/components/custom-field-input-currency/custom-field-input-currency.jsx
```diff
@@ -27,35 +27,36 @@ function getLocale() {
...
-  const maximumFractionDigits = currencyMetaData[currencyCode].maximumFractionDigits;
-
+function initialInputValid(inputValue) {
   if (!inputValue) {
     return true;
   }
 
-  const splitInputValue = inputValue.toString().split('.');
-  if (splitInputValue.length === 1) {
+  if (inputValue.toString().split('.').length === 1) {
```
cincospenguinos, line 35: This only splits on the American cents delineator `.` but do we need to worry about any others? I believe the pound uses `,` to separate parts from whole and the euro does as well.

juanca: We don't care! Since we are expecting `value` to be cents, we never get a float. This is just here to ensure we actually don't pass it a float.

PR #109, opened 2020-05-19
Implement Section component

Bloodyaugust:
## Motivation
We want to standardize our sections in a form (in our case, a custom app custom field form). 

## Acceptance Criteria 
There is a Section component in external MDS
The section must have a title (to support semantic usage of section)
The section can have a description
The section can have a fields in a row (horizontally spaced out but left aligned)
The section can have many rows
There is spacing between rows
There is spacing between fields in a row
There is spacing around (top, left, right, bottom) the section
Section component documentation:
States to only be used in custom apps (for now)
Uses google doc examples as actual examples (so we can verify it does the thing!)

[Deployment URL](https://mavenlink.github.io/design-system/webinf-get-that-section-goodness-172781818)
[Pivotal Story](https://www.pivotaltracker.com/story/show/172781818)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
     Click to see more.
   </summary>
 
+  - Patch: Fix MDS colors with incorrect transparency
 </details>
 
 ## 0.20.2 (May 15, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
     Changes that have landed in master but are not yet released.
     Click to see more.
   </summary>
+
+  Minor: Add Section component
 </details>
 
 ## 0.20.3 (May 15, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
     Click to see more.
   </summary>
 
-  Minor: Add Section component
+  - Minor: Add Section component
 </details>
 
 ## 0.20.3 (May 15, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,12 @@
     Click to see more.
   </summary>
 
-  - Minor: Create linter for new MDS colors
-  - Minor: Update documentation concerning new linter
-  - Patch: Fix linter errors in MDS repo
 </details>
+
+## 0.20.2 (May 15, 2020)
+- Minor: Create linter for new MDS colors
+- Minor: Update documentation concerning new linter
+- Patch: Fix linter errors in MDS repo
 
 ## 0.20.1 (May 13, 2020)
 - Minor: Update color documentation to newest version
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
   - Minor: Add Section component
   - Minor: Add SectionRow component
+  - Minor: Implement PageHeader component
 </details>
 
 ## 0.20.3 (May 15, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   </summary>
 
   - Minor: Add Section component
+  - Minor: Add SectionRow component
 </details>
 
 ## 0.20.3 (May 15, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@
     Changes that have landed in master but are not yet released.
     Click to see more.
   </summary>
-
-  - Patch: Fix MDS colors with incorrect transparency
 </details>
+
+## 0.20.3 (May 15, 2020)
+- Patch: Fix MDS colors with incorrect transparency
 
 ## 0.20.2 (May 15, 2020)
 - Minor: Create linter for new MDS colors
--- a/src/components/section/section.css
+++ b/src/components/section/section.css
@@ -1,0 +1,18 @@
+@import '../../styles/spacing.css';
+
+.description {
+  margin: 0;
+  margin-bottom: var(--spacing-large);
+}
+
+.section {
+  margin-bottom: var(--spacing-x-large);
+}
+.section > * {
+  margin-bottom: var(--spacing-large);
+}
+
+.subheading {
+  margin: 0;
+  margin-bottom: var(--spacing-large);
+}
--- a/src/components/section/section.css
+++ b/src/components/section/section.css
@@ -8,6 +8,7 @@
 .section {
   margin-bottom: var(--spacing-x-large);
 }
+
 .section > * {
   margin-bottom: var(--spacing-large);
 }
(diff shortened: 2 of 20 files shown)


src/components/section/section.jsx
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from './section.css';
+
+export default function Section(props) {
+  return (
+    <section className={props.className || styles.section}>
```
juanca, line 7: You need to use default props.

src/components/section/section.jsx
```diff
@@ -0,0 +1,26 @@
...
+      <h2 className={styles.subheading}>{props.title}</h2>
+      {props.description && <p className={styles.description}>{props.description}</p>}
+      {props.children}
+    </section>
+  );
+}
+
+Section.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
```
juanca, line 17: What is a use-case for overriding the styling on a section?

src/components/section/section.jsx
```diff
@@ -0,0 +1,26 @@
...
+  description: PropTypes.string,
+  title: PropTypes.string.isRequired,
+};
+
+Section.defaultProps = {
+  children: [],
```
juanca, line 23: I haven't seen this pattern before. Normally, children default to `undefined` since it is kinda weird to assume either a list of child element or a single child element (which is not an array):

- 0 children is `[]`
- 1 child is `element`
- Many children is `[...]`

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
     Changes that have landed in master but are not yet released.
     Click to see more.
   </summary>
+
+  Minor: Add Section component
```
juanca, line 8: For copy pasta purposes, we have been prefixing these changelog items with `-`. But it's been hard to maintain this particular section and I'm. thinking maybe it does not matter. 🤔

src/components/section/section.md
```diff
@@ -0,0 +1,41 @@
+Sections can be used to semantically group controls, information, and inputs. Sections get the `region` ARIA role by default, and the `title` prop is required to support accessibility with a contained heading. For now, the section component should only be used in Custom Apps, while the "page layout" spec is still in flux.
```
juanca, line 1: 🔥 Nice! Wow, now I want to go update the docs for PageHeader.

src/components/section/section.md
```diff
@@ -0,0 +1,41 @@
...
+/>
+```
+
+Sections can contain child elements, like any other element:
+```
+<Section
+  title="Project Details"
+  description="These fields allow control of custom project fields."
+>
+<CustomFieldInputText
```
juanca, line 16: Shouldn't nested elements be indented?

src/components/section/section.md
```diff
@@ -0,0 +1,41 @@
+Sections can be used to semantically group controls, information, and inputs. Sections get the `region` ARIA role by default, and the `title` prop is required to support accessibility with a contained heading. For now, the section component should only be used in Custom Apps, while the "page layout" spec is still in flux.
+
+```
+<Section
+  title="I am a section"
+  description="Test description"
+/>
+```
+
+Sections can contain child elements, like any other element:
```
juanca, line 10: Do we have an example of a Section with my rows of fields?

src/components/section/section.css
```diff
@@ -0,0 +1,19 @@
...
+.description {
+  margin: 0;
+  margin-bottom: var(--spacing-large);
+}
+
+.section {
+  margin-bottom: var(--spacing-x-large);
+}
+
+.section > * {
```
juanca, line 12: This rule is being applied to `subheading` as well! I'm also interested in seeing how this applies to a Section with many rows of children.

src/components/section/section.jsx
Bloodyaugust: Oh man, that's frustrating. I assumed it would be `[element]`. I'll change to `undefined`.

CHANGELOG.md
Bloodyaugust: It's all good, I'll make the change and adjust in the future.

src/components/section/section.md
Bloodyaugust: Yupp, I'll do this. Damn examples for not failing linting! 😉

src/components/section/section.md
Bloodyaugust: Every direct child is a row, so in this case each input is a row. Tables can be achieved by using a container with the proper CSS. Is that something we want to exemplify here?

src/components/section/section.css
Bloodyaugust: Oooooh, good catch! I'll make sure it doesn't apply to subheading, even though the rule is identical. Better for semantics, I think.

My goal was to support `n` rows of children with this wildcard selector, and this way the rows don't have to be a particular type of element.

src/components/section/section.md
juanca: Right, I didn't realized it was a row per field. I think what I want to see as well is a row with many fields.

src/components/section/section.jsx
Bloodyaugust: If you want different behavior for the child rows, I imagine. There are a few different ways to slice that problem, with one of them being an override for the styles. Is that something we want to discourage? Are we confident that we can achieve all possible design specs without such an override?

src/components/section/section.css
Bloodyaugust: On attempting to fix this, I had some more realizations:
- Both the subheading and description have the same `margin-bottom` as any other child element (like row containers).
- Any selector more precise would need to account for the description being optional (IE `*:nth-child(n+3)` would only work with a description present).
- Keeping the classes for a margin reset and moving down the section child selector dries up the css.

All that in mind, I opted to allow the section child selector to apply the `margin-bottom` rules. Lemme know if you have another preferred approach.

juanca: I think `*` is fine since we are using CSS modules _and_ we are only affecting direct children (`>`).

src/components/section/section.jsx
juanca: >  Is that something we want to discourage?

Hmmm... That is a good question: we are already implementing the design spec for a section component and it is pretty simple. Perhaps it is a good idea to avoid letting other change the styling for the section component.

> Are we confident that we can achieve all possible design specs without such an override?

I think a reasonable guess to customizing the styling for the section component would be to customize the rows of the sections. However, the rows are the children of the section which means any dev can just pass down their own fully styled rows if they need overrides. Those rows can even be composing from the future section-row component. 

I think overall, I'm still not sure what is the best outcome. I will let you decide now that we talked about implementing a SectionRow component.

PR #116, opened 2020-05-28
Tag component

juanca:
https://www.pivotaltracker.com/story/show/173020169

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-tag-me-baby-one-more-time-173020169


juanca:
We need to add labels for the clear icon. From the WAI ARIA example:

```
aria-label="Remove"
aria-labelledby="rb1 r1"
```

src/components/tag-skill/tag-skill.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import iconTick from '../../svgs/icon-tick.svg';
+import Icon from '../icon/icon.jsx';
+import Tag from '../tag/tag.jsx';
+import styles from './tag-skill.css';
+
+export default function TagSkill(props) {
+  return (
+    <Tag readOnly>
```
cincospenguinos, line 10: Do we want to assert that it's `readOnly` here? Traditionally we've set read-only in big maven and left these components as they are. I worry about having to deal with removing the `readOnly` when we make custom apps writable.

src/components/tag/tag.jsx
```diff
@@ -0,0 +1,97 @@
...
+import Icon from '../icon/index.js';
+
+export default function Tag(props) {
+  const [tabActiveStates, setTabActiveStates] = useState([true, false]);
+  const [inputHandled, setInputHandled] = useState(true);
+  const iconElement = useRef(null);
+  const titleElement = useRef(null);
+  const refElements = [titleElement, iconElement];
+
+  function handleGridCellKeyDown(keyEvent) {
```
cincospenguinos, line 14: It looks like we are handling active states--that is, focus--according to an array here. What if we switched that piece of state over to a string instead, and then had a helper function indicate whether we were here or there? Something like

```jsx
const [currentTab, setCurrentTab] = useState('tag')
// ...
const isIconActive = () => currentTab === 'icon';
// ...

// Then instead of tabActiveStates[1] we have
if (isIconActive()) {}
```

We could then extract that all to a separate hook and configure it for other tag-like components.

I won't demand such a change--this seems to work alright--I suppose it just smells a bit [primitive obsessive](https://refactoring.guru/smells/primitive-obsession) to me. As always though, take my thoughts with a grain of salt.

src/components/tag-skill/tag-skill.jsx
juanca: As per design, TagSkill are always going to be readOnly. The editing UX is a table of sorts.

src/components/tag/tag.jsx
juanca: That's good food for thought! I know we're going to be revisiting this pattern when we implement the list in a dropdown (since list items can also be keyboard navigable) -- and again when we want to DRY up accessibility functionality. 

For future us: I do want to revisit the data structure. I think we're going to want a dynamic array but perhaps not an array of booleans but maybe an array of refs? It expresses better intent (like using strings for static list of items).

Bloodyaugust: I'm not sure this could really be called primitive obsession though, because the only state we need to represent (for now) is the `focused` state, which is a boolean value.

cincospenguinos: That's true--but it's for the focused state across two elements, and one has to mentally remember that the first thing is the text and the second thing is the icon. Perhaps instead of a string an object with keys as labels for what is/isn't focused may be a nice refactor:

```jsx
const [currentTab, setCurrentTab] = useState({ tag: false, icon: false });
```

Or better yet, an object that has a string as its interface:

```jsx
const [currentTab, setCurrentTab] = useState('tag');
// currentTab = { tag: false, icon: false }
```

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -60,7 +60,7 @@
 }
 
 .input-icon {
-  composes: icon-base from '../icon/icon.css';
+  composes: icon-base from '../icon/icon.css'; /* stylelint-disable-line css-modules/composed-class-names */
```
juanca, line 63: This is captured @ https://github.com/juanca/stylelint-css-modules/issues/15

src/components/icon/icon.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styles from './icon.css';
 
-export default function Icon({ className, name, size, stroke, fill, currentColor, title }) {
+export default function Icon({ className, name, size, stroke, fill, currentColor, title, role, ariaLabel, ariaLabelledBy, a11yKey }) {
```
juanca, line 5: We will want a change log entry with these new props.

juanca: We will want to add documentation with a more detailed explanation of when to use these. Add an example as well so QA can test it.

src/components/icon/icon.jsx
```diff
@@ -12,14 +12,17 @@ export default function Icon({ className, name, size, stroke, fill, currentColor
   ].filter(Boolean);
 
   return (
-    <svg className={classes.join(' ')} role="img">
+    <svg id={`${a11yKey}-icon`} aria-label={ariaLabel || name.split('-')[1]} className={classes.join(' ')} aria-labelledby={ariaLabelledBy} role={role} >
```
juanca, line 15: Sometimes, the SVG is purely presentational. Think calendar icon and icon-caret-down. These don't do anything since focusing and activating (e.g. clicking) an input will have the desired action. Perhaps we can make the label optional and avoid adding a bad label through the SVG name?

> 2.1 No ARIA is better than Bad ARIA

https://www.w3.org/TR/wai-aria-practices-1.1/#no_aria_better_bad_aria

juanca: If the SVG is purely presentational, we will not want to have a ID associated with it as well.

juanca: While we're here, let's expand the JSX attributes into multi-line attributes for better git-ability.

src/components/icon/icon.jsx
```diff
@@ -58,9 +65,13 @@ Icon.propTypes = {
 };
 
 Icon.defaultProps = {
+  a11yKey: undefined,
```
juanca, line 68: We need tests for this.

src/components/icon/icon.jsx
```diff
@@ -58,9 +65,13 @@ Icon.propTypes = {
...
+  ariaLabel: undefined,
+  ariaLabelledBy: undefined,
```
juanca, line 70: We need tests for this prop.

src/components/icon/icon.jsx
```diff
@@ -12,14 +12,17 @@ export default function Icon({ className, name, size, stroke, fill, currentColor
...
-    <svg className={classes.join(' ')} role="img">
+    <svg id={`${a11yKey}-icon`} aria-label={ariaLabel || name.split('-')[1]} className={classes.join(' ')} aria-labelledby={ariaLabelledBy} role={role} >
       { title && <title>{title}</title> }
       <use xlinkHref={`#${name}`} />
     </svg>
   );
 }
 
 Icon.propTypes = {
+  a11yKey: PropTypes.string,
```
juanca, line 23: We need tests for this prop

src/components/icon/icon.jsx
```diff
@@ -12,14 +12,17 @@ export default function Icon({ className, name, size, stroke, fill, currentColor
...
+  ariaLabel: PropTypes.string,
+  ariaLabelledBy: PropTypes.string,
```
juanca, line 25: We need tests for this prop

src/components/tag-skill/tag-skill.jsx
```diff
@@ -0,0 +1,32 @@
...
+      <span className={styles.row}>
+        {props.name}
+        <span className={styles.level + (props.level ? '' : ` ${styles['level-tick']}`)}>
```
juanca, line 13: If the `level-tick` class composes from `level` then we don't need string concatenation. 

```suggestion
        <span className={props.level ? styles.level : styles['level-tick']}>
```

Way easier to read!

It can also help us identify how to reduce some overhead. Perhaps this is even easiest to read: 

```
{props.level ? (
  <span className={styles.level}>{props.level}</span>
) : (
  <span className={styles['level-tick']}>
    <Icon name={iconTick.id} size="small" stroke="skip" fill="skip" currentColor="skip" />
  </span>
)}
```

src/components/tag-skill/tag-skill.jsx
juanca, line 10: I believe we now need `a11yKey`

juanca: This should be captured in _some_ failing test because we have a prop-type violation. Does a test currently fail? Otherwise, I think it would be beneficial to look into getting it to fail -- I thought I did set it up.

src/components/tag/tag.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+describe('Tag', () => {
+  it('renders a tag component', () => {
+    const tree = renderer
+      .create((
+        <Tag>Test Title</Tag>
+      )).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('uses a title', () => {
```
juanca, line 16: This might be better represented as `children API`.

src/components/tag/tag.test.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { render, screen, fireEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import Tag from './tag.jsx';
+
+describe('Tag', () => {
+  it('renders a tag component', () => {
```
juanca, line 8:
```suggestion
  it('renders a tag component with defaults', () => {
```

src/components/tag/tag.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+      userEvent.click(screen.getAllByRole('gridcell')[1]);
+
+      expect(onClearSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('readOnly API', () => {
+    it('does not render a clear button if readOnly is true', () => {
+      render(<Tag readOnly={true}>Test Title</Tag>);
```
juanca, line 59: Just an FYI, this is the JSX shorthand (which is encouraged to use):

```suggestion
      render(<Tag readOnly>Test Title</Tag>);
```

src/components/tag/tag.test.jsx
```diff
@@ -0,0 +1,64 @@
...
+      render(<Tag onClear={onClearSpy}>Test Title</Tag>);
+
+      userEvent.click(screen.getAllByRole('gridcell')[1]);
+
+      expect(onClearSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('readOnly API', () => {
+    it('does not render a clear button if readOnly is true', () => {
```
juanca, line 58: Since this is testing a boolean prop API, we should be testing `true` and `false`. 

The fact that the defaults test tests the defaults, it does not make any assertion about how setting `readOnly={false}` (or lack of `readyOnly`) renders the clear button.

src/components/icon/icon.jsx
```diff
@@ -2,24 +2,33 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styles from './icon.css';
 
-export default function Icon({ className, name, size, stroke, fill, currentColor, title }) {
+export default function Icon(props) {
```
Bloodyaugust, line 5: 🎉

src/components/icon/icon.md
```diff
@@ -1,44 +1,64 @@
...
 ```jsx
 const icon = require('../../svgs/icon-caution-fill.svg').default;
-<Icon name={icon.id} size="large" currentColor="caution" title="alert icon" />
+<Icon name={icon.id} size="large" currentColor="skip" fill="skip" stroke="skip" title="alert icon" />
 ```
-Sizes
+
+### Sizes
+
+For the most part, the `size` API is the main configurable property.
```
Bloodyaugust, line 20: We should probably mention what those sizes are (in pixel units), and how the container size is not the icon size. Encourage the double checking of design mocks to ensure that the correct spacing is used, that kind of thing.

src/components/icon/icon.md
```diff
@@ -1,44 +1,64 @@
...
 ```
 
-Current Color Example
+```jsx
+const icon = require('../../svgs/icon-caution-fill.svg').default;
+<Icon name={icon.id} size="large" currentColor="skip" fill="skip" stroke="skip" />
+```
 
-Can be one of 'primary', 'action', 'highlight', or 'caution';
+### Accessibility
```
Bloodyaugust, line 37: Loving all the new Accessibility documentation! 🥳

src/components/tag-skill/tag-skill.md
```diff
@@ -0,0 +1,15 @@
+`TagSkill` is the variant of `Tag` for Mavenlink skills.
```
Bloodyaugust, line 1: We should probably mention that it is similarly intended for use in `TagList`.

src/components/tag/tag.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Accessibility
+
+This component's accessibility was built using the [WAI ARIA Examples](https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/LayoutGrids.html#ex2_label).
+
+#### Keyboard navigation
+
+| Key | State | Function |
+| --- | --- | --- |
+| Left arrow | | Moves focus to text content |
```
Bloodyaugust, line 11:
```suggestion
| Left arrow/Up arrow | | Moves focus to text content |
```

src/components/tag/tag.md
```diff
@@ -0,0 +1,82 @@
...
+| Right arrow | | Moves focus to the clear button |
```
Bloodyaugust, line 12:
```suggestion
| Right arrow/Down arrow | | Moves focus to the clear button |
```

src/svgs/icon-clear-small.svg
```diff
@@ -0,0 +1,7 @@
...
+    <title>Icons/Zz Working/X/Clear small</title>
+    <g id="Icons/Zz-Working/X/Clear-small" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" fill-opacity="0.38">
+        <path d="M5.61289944,4.20970461 L5.70710678,4.29289322 L8,6.585 L10.2928932,4.29289322 C10.6834175,3.90236893 11.3165825,3.90236893 11.7071068,4.29289322 C12.0675907,4.65337718 12.0953203,5.22060824 11.7902954,5.61289944 L11.7071068,5.70710678 L9.415,8 L11.7071068,10.2928932 C12.0976311,10.6834175 12.0976311,11.3165825 11.7071068,11.7071068 C11.3466228,12.0675907 10.7793918,12.0953203 10.3871006,11.7902954 L10.2928932,11.7071068 L8,9.415 L5.70710678,11.7071068 C5.31658249,12.0976311 4.68341751,12.0976311 4.29289322,11.7071068 C3.93240926,11.3466228 3.90467972,10.7793918 4.20970461,10.3871006 L4.29289322,10.2928932 L6.585,8 L4.29289322,5.70710678 C3.90236893,5.31658249 3.90236893,4.68341751 4.29289322,4.29289322 C4.62333685,3.96244959 5.12750547,3.91161211 5.51140295,4.14038077 L5.61289944,4.20970461 Z" id="Path" fill="#000000" fill-rule="nonzero"></path>
+    </g>
+</svg>
```
Bloodyaugust, line 7: Do we need newlines at the end of svg files? 🤔

juanca: I gave the good fight. I think we want a linter if we want it to be required.

src/components/tag-skill/tag-skill.md
juanca: I can link back to Tag. Lemme figure that out.

src/components/icon/icon.md
juanca: Yeah, after looking at our component doc page, I think accessibility should be the first section of every documentation page:

- highlights how to interact with the component (even for basic examples!)
- highlights accessibility features on top of that 
- makes sure we are doing the research on this

src/components/icon/icon.md
juanca: Good point. Lemme think about that phrasing.

PR #117, opened 2020-05-28
TagList component

juanca:
https://www.pivotaltracker.com/story/show/173020597

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/tag-list-tag-tag-tag-tag-tag


src/components/tag-list/tag-list.jsx
```diff
@@ -0,0 +1,67 @@
...
+  useState,
+} from 'react';
+import PropTypes from 'prop-types';
+import styles from './tag-list.css';
+import { useRef } from 'react';
+import { useEffect } from 'react';
+
+export default function TagList(props) {
+  const [active, setActive] = useState(false);
+  const [focusIndex, setFocusIndex] = useState(0);
```
juanca, line 11: Rename to `activeIndex`

We want to keep terminology consistent with accessibility:

- Something can be default focusable:
  - It could be focused
  - One of its children has a tabIndex of 0
- Something can be active:
  - One of its children is currently focused (or was focused)
  - One of its children has a tab index of 0
  - The moment a widget is active, it is no longer default focusable because its active state is now managed!
- Something can be inactive:
  - All of its children have a tab index of -1

src/components/tag-list/tag-list.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    if (active) {
+      props.refs.forEach((ref, index) => {
+        ref.current && ref.current.setIsParentActive(true);
```
juanca, line 16: Since a `Tag` now has a `ref API`, we want to document how `TagList` uses the `ref` API.

juanca: So maybe we only document that we want our child regs to implement the "parent active" and "self active" interface. This is gonna be reused in the future for all complex widgets but we don't need to get too ahead of ourselves with abstractions.

src/components/tag-list/tag-list.jsx
```diff
@@ -0,0 +1,67 @@
...
+      });
+    }
+
+  }, [active, ...props.refs.map(ref => ref.current), focusIndex]);
+
+  function handleOnKeyDown(keyEvent) {
+    switch (keyEvent.key) {
+      case 'ArrowDown':
+      case 'ArrowRight':
+        setFocusIndex(focusIndex + 1);
```
juanca, line 27: We might want to preventDefault here as well?

PR #119, opened 2020-06-05
TagList component

juanca:
https://www.pivotaltracker.com/story/show/173020597

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-a-sleeve-for-tags-173020597

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11247,8 +11247,9 @@ websocket-driver@>=0.5.1:
     websocket-extensions ">=0.1.1"
 
 websocket-extensions@>=0.1.1:
-  version "0.1.3"
-  resolved "https://registry.yarnpkg.com/websocket-extensions/-/websocket-extensions-0.1.3.tgz#5d2ff22977003ec687a4b87073dfbbac146ccf29"
+  version "0.1.4"
+  resolved "https://registry.yarnpkg.com/websocket-extensions/-/websocket-extensions-0.1.4.tgz#7f8473bc839dfd87608adb95d7eb075211578a42"
+  integrity sha512-OqedPIGOfsDlo31UNwYbCFMSaO9m9G/0faIHj5/dZFDMFqPTcx6UwqyOy3COEaEOg/9VsGIpdqn62W5KhoKSpg==
 
 whatwg-encoding@^1.0.1, whatwg-encoding@^1.0.5:
   version "1.0.5"


src/components/tag/tag.jsx
```diff
@@ -27,15 +30,21 @@ export default function Tag(props) {
...
-        setInputHandled(false);
-        setTabActiveStates(props.readOnly ? [true] : [false, true]);
+        if (!props.readOnly && tabActiveStates[0] === true && tabActiveStates[1] === false) {
+          keyEvent.stopPropagation();
+          setInputHandled(false);
+          setTabActiveStates([false, true]);
+        }
         break;
       case 'ArrowUp':
         keyEvent.preventDefault();
```
juanca, line 40: We shouldn't always prevent the default. We should only prevent the default when the Tag component is consuming the event. Otherwise, we can have a Tag in the middle of a page and the user wants to scroll right but needs to tab out of the Tag. That seems unintuitive. I wonder if there is a rule of thumb for this kind of thing. My intuition is pointing to: "if we use it then we prevent default, if we don't use it then we do the default (because we don't care about the event)"

src/components/tag-list/tag-list.css
```diff
@@ -0,0 +1,5 @@
+@import '../../styles/spacing.css';
+
+.tag-list > [role="row"] {
```
Bloodyaugust, line 3: Good choice of selector here. Nice way of enforcing the accessibility roles thru styles!

src/components/tag-list/tag-list.jsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  return (
+    <div
+      className={styles['tag-list']}
+      onClick={onClick}
+      onFocus={onFocus}
+      onKeyDown={onKeyDown}
+      role="grid"
+      tabIndex="-1"
```
Bloodyaugust, line 60: For clarification after reading the commit message this was introduced with: are we actually required to have the element be focusable? Or is having a `tabIndex` value enough?

src/components/tag-list/tag-list.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Mouse interactions
+
+Clicking on a tag will focus the tag.
+
+### Example
+
+```jsx
+const data = [
+  'Molotov Solution',
```
Bloodyaugust, line 45: We should probably change this datum to be more representative of a Mavenlink use case.

src/components/tag-list/tag-list.jsx
juanca: I think `tabIndex` is good enough.

src/components/tag-list/tag-list.md
juanca: I'm okay with this TBH.

PR #121, opened 2020-06-08
Leverage CI workspace instead of cache

juanca:
## Motivation

Restoring cache in `deploy` is about a minute. It is weird since we already have the dependencies from the test job. 

## Acceptance Criteria 

- Attaching workspace is faster than 20 seconds (thus faster than restoring cache)
- CI is green

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Deployment URL: https://mavenlink.github.io/design-system/faster_ci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,7 @@ jobs:
       - persist_to_workspace:
           root: ~/repo
           paths:
+            - ./
             - node_modules
       - run: yarn lint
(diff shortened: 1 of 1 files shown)


juanca:
Attaching the workspace can be as fast as 17 seconds or as slow as 30 seconds. This isn't worth it.

PR #124, opened 2020-06-10
Update Contributing.md for unused-files webpack failure

nhayesroth:
## Motivation

Update the wiki to include a "gotchas" section with a procedure to follow for new MDS components that are unused.

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -37,3 +37,77 @@ We use GitHub issues to track public bugs. Report a bug by [opening a new issue]
 ## Tips
 
 - `yarn test -o` runs only changed test files
+
+## Gotchas
+### CircleCI fails to Compile Webpack Bundles
+#### Description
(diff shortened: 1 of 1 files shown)


juanca:
Thanks for contributing! Unfortunately, this is for internal MDS usages. I ported the documentation to bigmaven @ https://github.com/mavenlink/mavenlink/pull/20327

As for "what is internal MDS": as of this writing, we have two MDSs called "internal MDS" and "external MDS". The external MDS is this repository which is publicly available for consumption. The internal MDS is the components in bigmaven which are only used in bigmaven -- this allows us to produce components without having to ensure they are ready for public consumption. I apologize for the misunderstanding! We're still trying to figure out what is the best way to distinguish both and how to educate existing and new contributors without a lengthy presentation / what-have-yous.

PR #125, opened 2020-06-11
Add linter for Typography

Bloodyaugust:
## Motivation

We want to avoid encouraging others in writing more one-offs. We want to encourage others to use MDS design tokens.

## Acceptance Criteria 

- Linter fails CI when there is a font-size that is not from MDS
- Linter fails CI when there is a font-type that is not from MDS
- Linter fails on usage of deprecated typography rules
- Linter's error message has a link to the MDS typography rules to engineers can understand why their build is failing
- Existing failures are ignored in situ with comments

https://www.pivotaltracker.com/story/show/172248339

Deployment URL: https://mavenlink.github.io/design-system/webinf-this-linter-this-linter-172248339


src/linters/typography.js
```diff
@@ -0,0 +1,54 @@
...
+const valueTokenIndices = {
+  font: 0,
+  'font-family': 0,
+  'font-size': 0,
+  'font-style': 0,
+  'font-weight': 0,
+  'line-height': 0,
+};
+
+const properties = Object.keys(valueTokenIndices);
```
cincospenguinos, line 27: Given that we are only grabbing the keys of the above variable, why don't we convert this into just an array of the various things we need:

```suggestion
const properties = ['font', 'font-family', 'font-size', 'font-style', 'font-weight', 'line-height'];
```

src/linters/typography.js
```diff
@@ -0,0 +1,54 @@
...
+
+module.exports = stylelint.createPlugin(ruleName, () => {
+  return (root, result) => {
+    root.walkDecls((declaration) => {
+      const property = declaration.prop;
+
+      if (properties.includes(property)) {
```
cincospenguinos, line 34: I'd extract this variable:

```suggestion
      const isFontProperty = properties.includes(property);
      if (isFontProperty) {
```

src/linters/typography.js
```diff
@@ -0,0 +1,54 @@
...
+        const valueTokens = declaration.value.split(' ');
+        const propertyIndex = valueTokenIndices[property];
```
cincospenguinos, line 36: Ah, we use the index here. Given that they're all zero, why not just replace this with 0?

src/linters/typography.js
```diff
@@ -0,0 +1,54 @@
...
+
+        if (!valueTokens[propertyIndex]) {
+          return;
+        }
+
+        const cssVariable = (valueTokens[propertyIndex].match(cssVarRegex) || [])[0];
```
cincospenguinos, line 42: This feels a little obfuscated. Could we extract this into a function that performs a similar operation? Something like

```suggestion
const extractCssVariable = () => {
  const cssVariable = valueTokens[0].match(cssVarRegex);
  
  if (!cssVariable) {
    return [];
  }
  
  return cssVariable;
}
```

src/linters/typography.js
Bloodyaugust: I liked the pattern from our previous linter that allows for index based checking of shorthand attributes, but I could be persuaded considering we don't use that here (yet). If we'd rather keep it simple instead of future-proof, I'll make the change.

src/linters/typography.js
Bloodyaugust: We could do that as well, but I think I'd rather either pull this all out or move to a twin-definition approach.

Bloodyaugust:
@cincospenguinos I refactored linting to be simpler, will you check it out again?

PR #126, opened 2020-06-11
Export the list of linted properties for static analysis capabilities

juanca:
https://www.pivotaltracker.com/story/show/173299530

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)


juanca:
I think we should wait on that. We might not need it. We will most likely iterate on the bigmaven static analysis scripts. Maybe at some point we can house the scripts here. 🤔

PR #132, opened 2020-06-17
Webinf i can still feel the box 173341864

juanca:
https://www.pivotaltracker.com/story/show/173341864

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-i-can-still-feel-the-box-173341864

### DIFF
--- a/src/components/list-option/list-option.test.jsx
+++ b/src/components/list-option/list-option.test.jsx
@@ -51,6 +51,24 @@ describe('src/components/list-option/list-option', () => {
   });
 
   describe('ref API', () => {
+    describe('contains', () => {
+      it('finds the container element', () => {
+        const ref = createRef();
+        render(<ListOption {...requiredProps} ref={ref} />);
+        expect(ref.current.contains(screen.getByText('Test option'))).toBe(true);
+      });
+
+      it('finds a child element', () => {
+        const ref = createRef();
+        render((
+          <ListOption {...requiredProps} ref={ref}>
+            <div>Nested child</div>
+          </ListOption>
+        ));
+        expect(ref.current.contains(screen.getByText('Nested child'))).toBe(true);
+      });
+    });
+
     describe('setActive', () => {
       it('sets active', () => {
         const ref = createRef();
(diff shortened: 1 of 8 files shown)


src/components/list-option/list-option.test.jsx
```diff
@@ -1,39 +1,96 @@
...
+      render(<ListOption {...requiredProps} defaultActive={false} />);
+      expect(screen.getByText('Test option')).toHaveAttribute('tabindex', '-1');
+    });
+  });
+
   describe('selected API', () => {
     it('is selected when selected is true', () => {
-      render(<ListOption selected>SelectMe!</ListOption>);
-      expect(screen.getByText('SelectMe!')).toHaveAttribute('aria-selected', 'true');
+      render(<ListOption {...requiredProps} selected={true} />);
```
cincospenguinos, line 43: If I'm not mistaken, boolean props don't need to have `true` be assigned to them.

```suggestion
      render(<ListOption {...requiredProps} selected />);
```

src/components/list-option/list-option.test.jsx
```diff
@@ -1,39 +1,96 @@
...
+      render(<ListOption {...requiredProps} selected={false} />);
+      expect(screen.getByText('Test option')).toHaveAttribute('aria-selected', 'false');
     });
   });
 
-  describe('title API', () => {
-    it('accepts a title', () => {
-      const title = "Hello. Is it me you're looking for?";
-      render(<ListOption title={title} />);
+  describe('ref API', () => {
```
cincospenguinos, line 53: 🎉

src/components/list-option/list-option.test.jsx
juanca: Yeah, I've been wavering on this but this is my latest train of thought:

Since we're testing `true` and `false`, I kinda want it to be explicit in both cases.

For example, in the `false` case, the "JSX way" of doing things is by not specifying the prop. But I want the reader to understand that the test is actually setting `selected` to `false.

In the same vein, I want to show the reader that I am testing `selected` set to `true`.

PR #140, opened 2020-06-25
[webinf-we-gonna-be-alright-173455369] Selection works in CustomFieldInputSingleChoice

cincospenguinos:
https://www.pivotaltracker.com/story/show/173455369

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-we-gonna-be-alright-173455369


src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -19,17 +23,34 @@ export default function CustomFieldInputSingleChoice(props) {
...
-      setShowOptions(true);
+      setShowOptions(!props.readOnly);
     } else if (event.key === 'Escape') {
       setShowOptions(false);
     }
   }
 
+  const listOptions = props.choices.map((item, index) => (
+    <ListOption
+      dataId={item.id}
```
Bloodyaugust, line 39: We could potentially use the native `id` attribute, if we can guarantee that they'll be unique. This could be achieved by prepending something like `list-option-${props.id}` and making `ListOption` require an `id` param. How do you feel about that approach?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -41,35 +62,41 @@ export default function CustomFieldInputSingleChoice(props) {
...
         >
-          { props.choices.map(item => <ListOption key={item}>{item}</ListOption>) }
+          { listOptions }
         </Listbox>
       ) }
     </div>
   );
 }
 
+const ChoiceType = PropTypes.shape({
```
Bloodyaugust, line 81: 👍  this is great! Do we want `id` and `label` to be required?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -41,35 +62,41 @@ export default function CustomFieldInputSingleChoice(props) {
...
+  value: ChoiceType,
 };
 
 CustomFieldInputSingleChoice.defaultProps = {
   choices: [],
   placeholder: undefined,
   readOnly: false,
   required: false,
-  value: undefined,
+  value: {},
```
Bloodyaugust, line 101: Since this is a `ChoiceType` now, should we include an `id` and `label` here in the default? 🤔

src/components/list-option/list-option.jsx
```diff
@@ -31,7 +41,7 @@ const ListOption = forwardRef(function ListOption(props, ref) {
   });
 
   return (<li
-    aria-selected={props.selected}
+    aria-selected={selected}
```
Bloodyaugust, line 44: Should this be
```suggestion
    aria-selected={active && selected}
```

To ensure that only `active` components will be `aria-selected`?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
cincospenguinos: Doing so caused failures because the default has `id` and `label` as `undefined`, since it's an empty object.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
cincospenguinos: I thought so, but it was causing spec failures.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Minor: Update `CustomFieldInputSingleChoice` to select things
```
juanca, line 10:
```suggestion
  - Minor: Update `CustomFieldInputSingleChoice` to select choices
```

In the future, we'll probably use the changelog as a means of communicating changes to the rest of the teams. We should consider the language used when drafting the changelog items -- we want to ensure concise but descriptive context. In this case, we know we want the user to select a "choice".

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
juanca: Or, how about using the `value` attribute? Looks like we set `value` to the choice's label but a (single choice) field's value is actually the choice's `id`.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
juanca: Would it be better to allow an `undefined` value? I know it makes more code but it would allow us to enforce the `id` and `label` properties of a choice type -- so the question is better phrased: do we think the benefits outweigh the extra code?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
```diff
@@ -13,17 +13,31 @@ describe('src/components/custom-field-input-single-choice/custom-field-input-sin
...
     it('shows choices when clicked', () => {
-      const choices = ['foo', 'bar'];
       render(<CustomFieldInputSingleChoice label="Foo" id="bar" choices={choices} />);
       userEvent.click(screen.getByLabelText('Foo'));
 
       expect(screen.getByText('foo')).toBeInTheDocument();
       expect(screen.getByText('bar')).toBeInTheDocument();
     });
 
+    it('does not show choices when readOnly is true', () => {
```
juanca, line 32: Is this a better test for the `readOnly` API? (Why do we think it is an accessibility test?)

src/components/listbox/listbox.jsx
```diff
@@ -17,18 +18,44 @@ export default function Listbox(props) {
...
+      const indexOfToggledRef = newSelectedIndexes.indexOf(toggledRefIndex);
+      newSelectedIndexes = newSelectedIndexes.splice(indexOfToggledRef, 1);
+    }
+
+    if (props.selectionLimit && newSelectedIndexes.length > props.selectionLimit) {
+      const deselectIndex = newSelectedIndexes.shift();
+      props.refs[deselectIndex].current.toggleSelected();
+    }
+
+    props.onSelectionChange(props.refs[toggledRefIndex].current.optionData);
```
juanca, line 37: I think it makes more sense to keep the API pretty similar to the select API. Each option has a value. When the option is selected, the select's value is that option's value. 

```suggestion
    props.onSelectionChange(props.refs[toggledRefIndex].current.value);
```

src/components/listbox/listbox.jsx
```diff
@@ -70,14 +104,26 @@ export default function Listbox(props) {
   );
 }
 
+const ListOptionRefType = PropTypes.shape({
+  current: PropTypes.shape({
```
juanca, line 108: We will want `value` in this shape if we refactor `optionData` to `value`

src/components/listbox/listbox.jsx
```diff
@@ -17,18 +18,44 @@ export default function Listbox(props) {
...
 
+  const refIndexOf = target => props.refs.findIndex(ref => ref.current && ref.current.contains(target));
+  const optionsWasSelected = (toggledRefIndex, isSelected) => {
+    let newSelectedIndexes = selectedIndexes.slice(0, selectedIndexes.length);
+
+    if (isSelected) {
+      newSelectedIndexes.push(toggledRefIndex);
+    } else {
+      const indexOfToggledRef = newSelectedIndexes.indexOf(toggledRefIndex);
+      newSelectedIndexes = newSelectedIndexes.splice(indexOfToggledRef, 1);
```
juanca, line 29: `splice`'s output is the array of removed items. How is this working? I think it would return the same array:

```
let newSelectedIndexes = [42];
newSelectedIndexes = newSelectedIndexes.splice(0, 1);
```

Then `newSelectedIndexes` is still `[42]`.

src/components/listbox/listbox.jsx
```diff
@@ -17,18 +18,44 @@ export default function Listbox(props) {
...
+    }
+
+    if (props.selectionLimit && newSelectedIndexes.length > props.selectionLimit) {
```
juanca, line 32: I think we might want to remove `selectionLimit` because we don't have a design for it. I don't think the existing Mavenlink application has any dropdown with a limited set of selected options. It is either: 1 or `multiple`

src/components/listbox/listbox.jsx
juanca: I think `onSelectionChange` might be better off if we pass it the whole list of selected values. Usually, when someone is listening to a change event, the value of the target is the whole value -- not just the diff. 

A possibly crazy point, usually when we are listening to an event, we want to pass an event object. This is going to sound wild but check it out: If `Listbox` had an imperative handle then we could `props.onSelectionChange({ target: ref })`. The consumer of `Listbox` would define their listener as;

```
function onListboxSelectionChange(event) {
  if (valid) {
    setValue(event.target.value);
  } else {
    setInvalid(true)
  }
}
```

Actually, perhaps we could simplify the event handler to just `onChange`.

src/components/listbox/listbox.jsx
```diff
@@ -8,6 +8,7 @@ import styles from './listbox.css';
 export default function Listbox(props) {
   const [active, setActive] = useState(false);
   const [activeIndex, setActiveIndex] = useState(0);
+  const [selectedIndexes, setSelectedIndexes] = useState([]);
```
juanca, line 11:
```suggestion
// const [selectedIndexes, setSelectedIndexes] = useState([]);

  useImperativeHandle(ref, () => ({
    value: props.refs.filter(listOptionRef => listOptionRef.current.selected).map(selectedRef => selectedRef.current.value)
  }))
```

src/components/listbox/listbox.jsx
juanca:
```suggestion
  props.onChange({ target: ref });
```

src/components/list-option/list-option.jsx
cincospenguinos: I think active components can be unselected, and selected components can be inactive

src/components/listbox/listbox.jsx
juanca: This is a bit tricky. I'll do a write up with an analysis of whether this should be a good "web component" API or not.

src/components/list-option/list-option.jsx
Bloodyaugust: Perhaps I'm conflating this incorrectly with `focused`.

src/components/listbox/listbox.jsx
juanca: To summarize the issue(s):

This does not work when we are consuming DOM events (e.g. `click`), setting state, and then calling `onChange` because the `ref.current` is "stale." 

In order to get this to work, we need to consume DOM events, set state, and on state change call `onChange` -- which ensures the `ref.current` is latest. This is a good pattern to propagate.

src/components/listbox/listbox.jsx
juanca: https://github.com/mavenlink/design-system/pull/140#discussion_r447188693

> To summarize the issue(s):

> This does not work when we are consuming DOM events (e.g. click), setting state, and then calling onChange because the ref.current is "stale."

> In order to get this to work, we need to consume DOM events, set state, and on state change call onChange -- which ensures the ref.current is latest. This is a good pattern to propagate.

However, we cannot use `value: props.refs...` because our listboxes will be paginated in the future. We cannot based listbox's value on the DOM because we might only have rendered part of the whole data.

PR #141, opened 2020-06-30
[webinf-why-s-single-choice-crying-173455213] MDS interactive single choice custom field component (3/5)

cincospenguinos:
https://www.pivotaltracker.com/story/show/173455213

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-why-s-single-choice-crying-173455213

### DIFF
--- a/src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
+++ b/src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
@@ -1,5 +1,4 @@
 import React, {
-  useEffect,
   useRef,
   useState,
 } from 'react';
@@ -16,7 +15,6 @@ export default function CustomFieldInputSingleChoice(props) {
   const [showOptions, setShowOptions] = useState(false);
   const [value, setValue] = useState(props.value);
   const [searchValue, setSearchValue] = useState(undefined);
-  const editRef = useRef();
 
   const refs = props.choices.map(() => useRef());
   const caretIcon = (<Icon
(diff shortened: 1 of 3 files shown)


src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -90,9 +138,12 @@ const ChoiceType = PropTypes.shape({
 CustomFieldInputSingleChoice.propTypes = {
   id: PropTypes.string.isRequired,
   choices: PropTypes.arrayOf(ChoiceType),
+  // eslint-disable-next-line react/no-unused-prop-types
```
cincospenguinos, line 141: I only included these because I could not figure out how to get lint to see that I was definitely using these props. If y'all have a way to get lint to chill out without these, that'd be great

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -34,48 +35,71 @@ export default function CustomFieldInputSingleChoice(props) {
...
+  function getOptions() {
+    const choices = {};
+    props.choices.forEach((item, index) => { choices[item.id] = { ...item, index }; });
 
-  function onChange(event) {
+    if (searchValue) {
+      const searchValueLowerCase = searchValue.toLowerCase();
+
+      return props.choices
+        .filter(item => item.label.toLowerCase().includes(searchValueLowerCase))
```
juanca, line 46: At some point, we'll want to ask design / product what kind of filtering algorithm would be ideal. For now, this is good enough.

PR #142, opened 2020-07-01
Interactive custom field multi choice has a popup

juanca:
https://www.pivotaltracker.com/story/show/173568241

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [X] Deployment URL: https://mavenlink.github.io/design-system/interactive-multi-choice-1

### DIFF
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
@@ -21,3 +21,14 @@
 .input-icon {
   composes: input-icon from '../custom-field-input-text/custom-field-input-text.css';
 }
+
+.dropdown-container {
+  composes: container from '../listbox/listbox.css';
+  background-color: white;
+  box-sizing: border-box;
+  max-height: 160px;
+  overflow: scroll;
+  position: absolute;
+  width: 100%;
+  z-index: 10;
+}
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
@@ -24,7 +24,7 @@
   composes: input-icon from '../custom-field-input-text/custom-field-input-text.css';
 }
 
-.dropdown-container {
+.popup-container {
   composes: container from '../listbox/listbox.css';
   background-color: var(--white);
   box-sizing: border-box;
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
@@ -1,3 +1,5 @@
+@import '../../styles/colors-v2.css';
+
 .container {
   composes: tag-list from '../tag-list/tag-list.css';
   composes: input-container from '../custom-field-input-text/custom-field-input-text.css';
@@ -24,7 +26,7 @@
 
 .dropdown-container {
   composes: container from '../listbox/listbox.css';
-  background-color: white;
+  background-color: var(--white);
   box-sizing: border-box;
   max-height: 160px;
   overflow: scroll;
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
@@ -29,7 +29,7 @@
   background-color: var(--white);
   box-sizing: border-box;
   max-height: 160px;
-  overflow: scroll;
+  overflow-y: scroll;
   position: absolute;
   width: 100%;
   z-index: 10;
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
@@ -88,7 +88,7 @@ CustomFieldInputMultipleChoice.propTypes = {
   choices: PropTypes.arrayOf(PropTypes.shape({
     id: PropTypes.string.isRequired,
     label: PropTypes.string.isRequired,
-  })),
+  })).isRequired,
   id: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
   readOnly: PropTypes.bool,
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
@@ -36,6 +36,7 @@ function CustomFieldInputMultipleChoice(props) {
   function onChoiceSelect(event) {
     const selectedChoiceIndex = choicesRefs.findIndex(choiceRef => choiceRef === event.target);
     const selectedChoice = visibleChoices[selectedChoiceIndex];
+    setExpanded(false);
     setValue([...value, selectedChoice]);
   }
 
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
@@ -17,9 +17,17 @@ function CustomFieldInputMultipleChoice(props) {
   const choicesRefs = props.choices.map(() => useRef());
   const valueRefs = props.value.map(() => useRef());
   const [expanded, setExpanded] = useState(false);
+  const [value, setValue] = useState(props.value);
   const classContainer = props.readOnly ?
     styles['read-only-container'] :
     styles['read-write-container'];
+
+  function onChoiceRemove(event) {
+    const newValue = value.filter((choice, index) => (
+      valueRefs[index].current !== event.target
+    ));
+    setValue(newValue);
+  }
 
   function onClick() {
     setExpanded(true);
@@ -38,11 +46,12 @@ function CustomFieldInputMultipleChoice(props) {
         onClick={onClick}
         refs={valueRefs}
       >
-        {props.value.map((choice, index) => (
+        {value.map((choice, index) => (
           <Tag
             defaultActive={index === 0}
             id={`${props.id}-${choice.id}`}
             key={`${props.id}-${choice.id}`}
+            onRemove={onChoiceRemove}
             readOnly={props.readOnly}
             ref={valueRefs[index]}
           >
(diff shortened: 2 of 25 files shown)


src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
```diff
@@ -21,3 +23,14 @@
 .input-icon {
   composes: input-icon from '../custom-field-input-text/custom-field-input-text.css';
 }
+
+.popup-container {
+  composes: container from '../listbox/listbox.css';
+  background-color: var(--white);
+  box-sizing: border-box;
+  max-height: 160px;
+  overflow: scroll;
```
Bloodyaugust, line 32: I think we want
```suggestion
  overflow-y: scroll;
```

Looking at the DRD, items in the dropdown that are too long to render on one line should use an ellipsis.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
```diff
@@ -21,3 +23,14 @@
...
+  position: absolute;
+  width: 100%;
+  z-index: 10;
```
Bloodyaugust, line 35: I can see this being a problem depending on the page it's used in. We should probably tackle documenting z-indices for different use-cases as part of our design consistency effort, but for now, I wonder if we should just set it to `2` and let people compose/override for their particular case?

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -1,51 +1,123 @@
...
 import TagList from '../tag-list/tag-list.jsx';
 import Tag from '../tag/tag.jsx';
 import styles from './custom-field-input-multiple-choice.css';
 
 function CustomFieldInputMultipleChoice(props) {
-  const refs = props.value.map(() => useRef());
+  const listboxRef = useRef();
+  const [expanded, setExpanded] = useState(false);
+  const [value, setValue] = useState(props.value);
+  const visibleChoices = props.choices.filter(choice => !value.includes(choice));
```
Bloodyaugust, line 22: _Very_ nice!

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -1,51 +1,123 @@
...
+  }
+
+  function onClick() {
+    setExpanded(true);
+  }
+
+  function onKeyDown(event) {
+    switch (event.key) {
+      case 'Escape':
+        setExpanded(false);
```
Bloodyaugust, line 50: Do we want to export the `setExpanded` function with `useImperativeHandle` so that parent components can provide a "click-out" functionality for closing the dropdown?

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -1,51 +1,123 @@
...
             readOnly={props.readOnly}
-            ref={refs[index]}
+            ref={valueRefs[index]}
           >
             {choice.label}
           </Tag>
         ))}
         <Icon className={styles['input-icon']} name={props.readOnly ? iconCaretDownDisabled.id : iconCaretDown.id} fill="skip" />
       </TagList>
+      {(expanded && visibleChoices.length !== 0 &&
```
Bloodyaugust, line 90: I think this is a good candidate for its own `renderListbox` function.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
```diff
@@ -16,18 +25,80 @@ describe('<CustomFieldInputMultipleChoice>', () => {
     expect(renderer.create(<CustomFieldInputMultipleChoice {...requiredProps} />).toJSON()).toMatchSnapshot();
   });
 
+  describe('autocompleter popup', () => {
```
Bloodyaugust, line 28:
```suggestion
  describe('listbox popup', () => {
```

Just so we don't confuse ourselves when we add the `autocomplete` functionality.

src/components/listbox/listbox.jsx
```diff
@@ -96,6 +74,29 @@ const Listbox = forwardRef(function Listbox(props, forwardedRef) {
     }
   }
 
+  useEffect(() => {
+    if (active) {
+      props.refs.forEach((optionRef, index) => {
+        if (optionRef.current) optionRef.current.setActive(index === activeIndex);
+      });
+    }
+  });
```
Bloodyaugust, line 83:
```suggestion
  }, []);
```

Elsewhere, we are explicit about having no dependencies. We should probably standardize on that pattern.

src/components/tag/tag.test.jsx
```diff
@@ -107,15 +107,35 @@ describe('Tag', () => {
...
+    it('calls the onRemove handler when the icon is clicked', () => {
+      const ref = createRef();
+      const onRemoveSpy = jest.fn();
+      render(<Tag {...requiredProps} onRemove={onRemoveSpy} ref={ref} />);
+      userEvent.click(screen.getByRole('button'));
+      expect(onRemoveSpy).toHaveBeenCalledWith({ target: expect.anything() });
+    });
 
-      userEvent.click(screen.getAllByRole('gridcell')[1]);
+    it('calls the onRemove handler when enter key is pressed', () => {
```
Bloodyaugust, line 119:
```suggestion
    it('calls the onRemove handler when enter key is pressed while the remove icon is focused', () => {
```

Below as well. Just so we don't confuse that action buttons should remove tag items when the main content is focused.

juanca:
> TagList items that have not yet been focused, when navigated to from the right, jump to their text content instead of their clear icon

That sounds like a bug in TagList. Let's create a new story for the epic "MDS Feedback" to deal with this. I also noticed some issue with focus when clicking on the remove button. 

> Focus indicator around Tag items is very tight and makes the text hard to read

I think we might want design's input on this. Personally, I haven't noticed it but that's just me.

> Does "click anywhere inside field" also include TagList items? Ex: should clicking a TagList item open the listbox dropdown? The DRD seems vague on this

I doubt it because that implies that we remove TagList functionality.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
juanca: I'm not sure yet on that API! Historically, we'll unfortunately want to set an event listener on the body and do some `node.contains` type of thing. That is based on a per-component basis that has a popup. So for now, we're punting so we can exclusively focus on the issue in a future story.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
juanca: `2` doesn't work with React Styleguidist. :( I think it sets a z-index of like 5 or something on each code example so it renders weird. :(

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
juanca: I'll verify with design.

src/components/listbox/listbox.jsx
juanca: This implies only for the on mount lifecycle which we do not want in this case.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
Bloodyaugust: Oh damn, that's unfortunate. Let's leave it for now, then.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
Bloodyaugust: In either case, we don't want the x scrollbar. 😄

src/components/tag/tag.test.jsx
juanca: I've been thinking about this and I think if we try to be overly descriptive then our tests will become unwieldy pretty fast since our UI components will be pretty complex with a11y and composition. 

I think the `it` description should be pretty succinct and then the test itself will setup the correct state. Otherwise, when we want to update behavior then we'll need to revise / reorganize the test descriptions -- which honestly never happens since we tend to look for failed tests or add new tests.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
juanca: It's part of the autocomplete functionality though. The popup are the values for the autocomplete!

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
juanca: Or a variable! I'll do this.

src/components/listbox/listbox.jsx
Bloodyaugust: You're right! In that case, we probably want:
```suggestion
  }, [active]);
```

juanca:
```suggestion
  }, [active, activeIndex]);
```

I think. Lemme see how that plays out.

Bloodyaugust:
> > TagList items that have not yet been focused, when navigated to from the right, jump to their text content instead of their clear icon
> 
> That sounds like a bug in TagList. Let's create a new story for the epic "MDS Feedback" to deal with this. I also noticed some issue with focus when clicking on the remove button.

https://www.pivotaltracker.com/story/show/173632521

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
juanca: Looks like the default is to truncate already without modifying the CSS property. I think the behavior is built into the Listbox.

juanca:
@Bloodyaugust This is ready for a re-review.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
Bloodyaugust: @juanca Still want to make the `overflow` change? We don't need the horizontal scroll.

PR #143, opened 2020-07-02
Publish all files for external mds b/c why not?!

bennapp:
## Motivation
We are doing this because in Interanal Mds we want to use the config in order to publish what is on external mds. We figure the overhead of publishing all files is nbd.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)


juanca:
Maybe add a changelog entry?

juanca:
Also, for posterity's sake: we wanted access to the styleguidist configuration in bigmaven to auto-sync configs. But as the main description states: why not publish everything else?

PR #144, opened 2020-07-06
Interactive multi choice autocompleter

juanca:
https://www.pivotaltracker.com/story/show/173568337

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/interactive-multi-choice-2


juanca:
> Not sure why, or if we want to track this, but it seems like arrowing around in the autocomplete input... "lags"? It seems to miss some navigations between characters.

I haven't noticed it. I wonder if it's computer related? Let's keep note of it and see if we can reproduce at a later time -- probably some hook timing tbh.

PR #145, opened 2020-07-06
[webinf-two-hundred-years-of-custom-fields-171681529] 

cincospenguinos:
[So listen up, here's the story](https://www.pivotaltracker.com/n/projects/2438324/stories/171681529)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-two-hundred-years-of-custom-fields-171681529
- [x] (When ready for review) Reviewer(s)

PR #147, opened 2020-07-07
Webinf clear and error single 173455125

bennapp:
https://www.pivotaltracker.com/story/show/173455125

- [x] Show error state with additional icons
- [x] Show clear icon for single choice

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-clear-and-error-single-173455125/#/Components/CustomFieldInputSingleChoice


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ## 0.28.0 (July 02, 2020)
 - Minor: Publish all files in repo to NPM.
 - Minor: Adjust typography table to more clearly display MDS variable names
+- Minor: Update `CustomFieldInputSingleChoice` to remove all selected choices
```
Bloodyaugust, line 18: This should go in the `unreleased` block.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -26,6 +27,22 @@ export default function CustomFieldInputSingleChoice(props) {
     fill="skip"
   />);
 
+  const clear = () => {
+    setValue(undefined);
+    setSearchValue(undefined);
+  };
+
+  const clearIcon = () => {
+    if (!props.readOnly && (!!value || !!searchValue)) {
```
Bloodyaugust, line 36: Nit: I don't think we really use the `!!` pattern for boolean coercion anywhere, and I think it's a bit less readable than comparing to the values that we don't want. Also, do we actually need more than the default truthy check?

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -2,14 +2,18 @@
...
 
 .input {
   border: 1px solid var(--mds-grey-12);
   border-radius: var(--spacing-x-small);
   box-sizing: border-box;
   color: var(--mds-grey-87);
   font-size: var(--mavenlink-type-base-size); /* stylelint-disable-line mds/typography */
-  padding: 7px var(--spacing-x-large) 7px var(--spacing-medium);
+  padding-top: 7px;
+  padding-right: calc((var(--numIcon, 0) * var(--iconSizeMedium)) + var(--spacing-medium)); /* stylelint-disable-line css-modules/css-variables */
```
Bloodyaugust, line 14: I think it might be worth re-factoring this to multiple classes that add more padding as we need it, which are added depending on the state of the input. With that approach, we wouldn't have more padding than we need for the state, and wouldn't have to do the math here in CSS.

src/components/custom-field-input-text/custom-field-input-text.css
```diff
@@ -40,10 +43,13 @@
...
-  pointer-events: none;
+.icon-container {
   position: absolute;
   right: 6px;
   top: 6px;
 }
+
+.input-icon {
+  composes: icon-base from '../icon/icon.css'; /* stylelint-disable-line css-modules/composed-class-names */
+  pointer-events: none;
```
Bloodyaugust, line 54: This turns off click events, right? Why do we want those to not fire?

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -6,21 +6,21 @@ import FormControl from '../form-control/form-control.jsx';
...
-}
-
 export default function CustomFieldInputText(props) {
   const defaultRef = useRef(null);
   const [validationMessage, setValidationMessage] = useState('');
 
   const inputRef = props.inputRef || defaultRef;
   const labelId = `${props.id}-label`;
 
+  const isInvalid = () => {
```
Bloodyaugust, line 16: Good move! I'm not sure why we need to check `readOnly` though, couldn't we assume that the existing value is valid?

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -70,6 +95,7 @@ export default function CustomFieldInputText(props) {
         aria-haspopup={props.ariaProps.haspopup}
         defaultValue={props.defaultValue}
         className={styles.input}
+        style={{ '--numIcon': numIcons() }}
```
Bloodyaugust, line 98: Very cool usage of a local style variable! See the above comment about a possible refactor of this behavior, though.

An alternate option to the previous suggestion would be to compute the padding here, to keep the CSS as simple as possible.

src/components/custom-field-input-text/custom-field-input-text.css
bennapp: I am not sure, this was just moved.

src/components/custom-field-input-text/custom-field-input-text.css
bennapp: What are the advantages of not doing the math in CSS? We actually started with an approach like that and went to this >.<

src/components/custom-field-input-text/custom-field-input-text.jsx
bennapp: This was just consolidating / refactoring the previous logic. I am not sure why it was like this.

src/components/custom-field-input-text/custom-field-input-text.jsx
bennapp: I think it is nice making use of the css variables

src/components/custom-field-input-text/custom-field-input-text.css
Bloodyaugust: Really just keeping the CSS as simple as possible, I don't think there are any performance implications or anything. Might have a slightly negative effect on redraw? But really, just to #keeplogicoutofcss.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
bennapp: Looks cleaner too :)

src/components/custom-field-input-text/custom-field-input-text.css
bennapp: If there is a nice way to pull `var(--spacing-medium)` into react-land, then I would be all for this. Is there some pattern I could use?

Since we define variables in css, I think it is nice to use those variables to do some simple math in css-land. Otherwise, I think we have to duplicate those variables in react.

Bloodyaugust: I think we process them into JS variables as part of the Webpack process, but I'm not sure exactly how that works or where they're accessible.

src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -33,8 +33,25 @@ export default function CustomFieldInputText(props) {
     }
   });
 
-  const icon = () => {
-    if (isInvalid(props.error, props.readOnly)) {
+  const numIcons = () => {
```
cincospenguinos, line 36: Wow this really clever! I'm worried that it could be too clever for its own good. What if we extracted this out to a hook? Something like

```jsx
import { useState, useEffect } from 'react';

export default function useIconTally(isInvalid, customFieldProps) {
  const cssVarName = `--numIcon-${customFieldProps.id}`;
  const [numIcons, setNumIcons] = useState(0);
  const [iconStyles, setIconStyles] = useState({ [cssVarName]: numIcons })

  useEffect(() => {
    let num = 0;

    if (isInvalid) {
      num += 1;
    }

    if (customFieldProps.icon) {
      num += 1;
    }

    if (customFieldProps.clear) {
      num += 1;
    }

    setNumIcons(num);
    setIconStyles({ [cssVarName]: num });
  });

  return [numIcons, iconStyles];
}
```

You could test it with `@testing-library/render-hook`, and then we have an extra hook to use internally across each of the custom fields as needed.

cincospenguinos: Also, if we use #145's refactor of extracting an abstract custom field, we could include this as a hook on that, and then every custom field we use would receive this behavior.

bennapp:
@cincospenguinos I understand that pulling out that behavior and making it reusable could be a nice refactor.

But why is that solution too clever, I am not clear on that. Is there a way to accomplish that spacing logic in a less complex way?

PR #148, opened 2020-07-08
Webinf currency is what you think it is 171681533

Bloodyaugust:
https://www.pivotaltracker.com/story/show/171681533

## PR checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-currency-is-what-you-think-it-is-171681533
- [x] Green [Bigmaven CI check](https://github.com/mavenlink/mavenlink/pull/20678/commits/f33ebedc5147d0ed19a3972f12ce2a2fc9badfef)


src/components/custom-field-input-currency/custom-field-input-currency.jsx
```diff
@@ -73,6 +74,12 @@ export default function CustomFieldInputCurrency(props) {
     }
   });
 
+  useImperativeHandle(ref, () => ({
+    value: () => {
```
juanca, line 78: We should keep this as a getter for a static value instead of a function:

```suggestion
    value: [parseInt(componentRef.current.value().replace(/\D/g, ''), 10), props.currencyCode],
```

src/components/custom-field-input-currency/custom-field-input-currency.jsx
```diff
@@ -73,6 +74,12 @@ export default function CustomFieldInputCurrency(props) {
...
+      return [parseInt(componentRef.current.value().replace(/\D/g, ''), 10), props.currencyCode];
```
juanca, line 79: I love that this works because virtually all currencies are in decimal!

Bloodyaugust: Even when they're not a `.` separator, they're `,`, which is covered by `\D`! 😄

src/components/custom-field-input-currency/custom-field-input-currency.jsx
Bloodyaugust: @juanca I'm not sure exactly why, but without this in `function` form, the fetched value does not seem to change.

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -63,10 +72,7 @@ CustomFieldInputNumber.propTypes = {
   readOnly: PropTypes.bool,
   required: PropTypes.bool,
   step: PropTypes.number,
-  value: PropTypes.oneOfType([
-    PropTypes.number,
-    PropTypes.oneOf(['']),
-  ]),
+  value: PropTypes.number,
```
juanca, line 75: I just realized this: this is a breaking change. I'll add it to the AC list.

PR #151, opened 2020-07-15
> Webinf some refactors

cincospenguinos:
## Motivation

There was a discrepancy between the ref API of CustomFieldInputSingleChoice here and on big maven, and so I adjusted this to accommodate. This also pulls out the duplicated useImperativeHandle into its own hook specifically for custom fields.

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-currency-is-what-you-think-it-is-171681533
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/custom-field-input-text/custom-field-input-text.jsx
+++ b/src/components/custom-field-input-text/custom-field-input-text.jsx
@@ -8,13 +8,7 @@ const CustomFieldInputText = forwardRef(function CustomFieldInputText(props, ref
   const inputRef = props.inputRef || defaultRef;
 
   const validationMessage = useValidation(props.readOnly, props.errorText, inputRef);
-
-  useImperativeHandle(ref, () => ({
-    id: props.id,
-    get value() {
-      return inputRef.current.value;
-    },
-  }));
+  useImperativeHandle(props.id, ref, () => inputRef.current.value);
 
   return (
     <AbstractCustomField
--- a/src/hooks/use-custom-field-value.jsx
+++ b/src/hooks/use-custom-field-value.jsx
@@ -1,0 +1,10 @@
+import { useImperativeHandle } from 'react';
+
+export default function useCustomFieldValue(id, ref, getValue) {
+  useImperativeHandle(ref, () => ({
+    id,
+    get value() {
+      return getValue();
+    },
+  }));
+}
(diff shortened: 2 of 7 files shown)


src/components/custom-field-input-text/custom-field-input-text.jsx
```diff
@@ -8,13 +8,7 @@ const CustomFieldInputText = forwardRef(function CustomFieldInputText(props, ref
...
 
   const validationMessage = useValidation(props.readOnly, props.errorText, inputRef);
-
-  useImperativeHandle(ref, () => ({
-    id: props.id,
-    get value() {
-      return inputRef.current.value;
-    },
-  }));
+  useImperativeHandle(props.id, ref, () => inputRef.current.value);
```
juanca, line 11: This is pointing to the wrong thing.

src/hooks/use-custom-field-value.jsx
```diff
@@ -0,0 +1,10 @@
+import { useImperativeHandle } from 'react';
+
+export default function useCustomFieldValue(id, ref, getValue) {
```
juanca, line 3: The `useCustomFieldValue` name sounds like a `useState` derivative.

A better name for this module is: `useCustomFieldHandle` since it is creating an imperative handle for a custom field and we want to centralize most of the logic into this module.

juanca:
I'm concerned this module isn't giving us much. It looks like instead of writing:

```
  id: <ID>,
  get value() {
    <value function>
  }
});
```

We are condensing it into a series of function parameters:

```
useCustomFieldValue(ref, id, value)
```

A couple of future use-cases:

1. We add a new ref property: 
  a. Using `useImperativeHandle` implies we need to touch every custom field component to add new property
  b. Using `useCustomFieldValue` implies we need to touch every custom field component to add new function parameter _and_ the centralized module for what is most likely going to be a 1 liner
1. We remove a new ref property:
  a. Same as above
  b. Same as above

What are the pros vs cons for this new custom hook?

cincospenguinos:
I'm unsure of the pros/cons associated with this little change as well, after reflection. Let's close this and we will evaluate whether it is needed at a later date.

PR #156, opened 2020-07-20
Improves accessibility of single and multi choice

bennapp:
## Motivation
Make things accessible woo! 🥳 

https://www.pivotaltracker.com/story/show/173478278

## Acceptance Criteria 
- Single choice bugs
  - (Only in Chrome) The field label is not read when focusing the input - DONE
  - (Only in Chrome) When selecting a new option, the voiceover reads the last option selected instead of the new one. - DONE
  - (Only in Chrome) When navigating between single choice fields, the labels are sometimes not read.  - DONE
  - The clear icon is in page tab sequence (otherwise a keyboard only user cannot deselect a choice field). Voiceover should read "Remove selected choice" or something - DONE

- Multi choice bugs
  - After selecting a choice, the autocomplete input should get focused - DONE
  - After selecting a choice, voiceover reads what was typed in autocomplete instead of the selected choice (or what should it read?) - DONE
  - The clear all icon is in page tab sequence? Voiceover should read "Remove all selected choices" or something - DONE
  - When invalid, voiceover should read the error message - DONE

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
@@ -1,9 +1,9 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
-import Icon from '../icon/icon.jsx';
+import Icon from '../../icon/icon.jsx';
 import styles from './abstract-custom-field.css';
-import cautionSvg from '../../svgs/icon-caution-fill.svg';
-import FormControl from '../form-control/form-control.jsx';
+import cautionSvg from '../../../svgs/icon-caution-fill.svg';
+import FormControl from '../../form-control/form-control.jsx';
 
 export default function AbstractCustomField(props) {
   const labelId = `${props.id}-label`;
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
@@ -32,7 +32,6 @@ export default function AbstractCustomField(props) {
         className={styles['input-icon']}
         currentColor="caution"
         name={cautionSvg.id}
-        ariaLabel={'Error'}
         size="medium"
       />);
     }
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
@@ -32,7 +32,7 @@ export default function AbstractCustomField(props) {
         className={styles['input-icon']}
         currentColor="caution"
         name={cautionSvg.id}
-        titleAttrText={'Error'}
+        ariaLabel={'Error'}
         size="medium"
       />);
     }
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
@@ -77,11 +77,15 @@ describe('AbstractCustomField', () => {
       expect(getByRole('img')).toBeDefined();
     });
 
-    it('renders the error icon first with an error title for accessibility', () => {
-      const icon = <Icon name={calendarSvg.id} currentColor="action" title="Hello" />;
-      const { getAllByRole } = render(<AbstractCustomField {...sharedProps} icon={icon} errorText="yo" />);
-      expect(getAllByRole('img')[0].firstChild).toHaveAttribute('xlink:href', '#icon-caution-fill.svg');
-      expect(getAllByRole('img')[0]).toHaveAttribute('title', 'Error');
+    it('renders the icon passed in', () => {
+      const icon = <Icon name={calendarSvg.id} currentColor="action" title="Hello" ariaLabel={'Label'} />;
+      render(<AbstractCustomField {...sharedProps} icon={icon} errorText="yo" />);
+      expect(screen.getByRole('img', { name: 'Label' }).children[1]).toHaveAttribute('xlink:href', '#icon-calendar-fill.svg');
+    });
+
+    it('renders the error with aria-label for accessibility', () => {
+      render(<AbstractCustomField {...sharedProps} errorText="yo" />);
+      expect(screen.getByRole('img', { name: 'Error' }).firstChild).toHaveAttribute('xlink:href', '#icon-caution-fill.svg');
     });
 
     it('shows no icon by default', () => {
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { fireEvent, render, screen } from '@testing-library/react';
 import AbstractCustomField from './abstract-custom-field.jsx';
-import Icon from '../icon/icon.jsx';
-import calendarSvg from '../../svgs/icon-calendar-fill.svg';
+import Icon from '../../icon/icon.jsx';
+import calendarSvg from '../../../svgs/icon-calendar-fill.svg';
 
 describe('AbstractCustomField', () => {
   const sharedProps = {
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
@@ -27,6 +27,12 @@ describe('AbstractCustomField', () => {
       render(<AbstractCustomField {...sharedProps} ariaProps={ariaProps} />);
       expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-controls', 'test-id-label');
     });
+
+    it('sets aria-describedby to link to a hint and aria-invalid when there are errors', () => {
+      render(<AbstractCustomField {...sharedProps} errorText={'there are errors.'} />);
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-describedby', 'test-idHint');
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-invalid', 'true');
+    });
   });
 
   describe('className API', () => {
@@ -71,10 +77,11 @@ describe('AbstractCustomField', () => {
       expect(getByRole('img')).toBeDefined();
     });
 
-    it('gives preference to the error icon', () => {
+    it('renders the error icon first with an error title for accessibility', () => {
       const icon = <Icon name={calendarSvg.id} currentColor="action" title="Hello" />;
       const { getAllByRole } = render(<AbstractCustomField {...sharedProps} icon={icon} errorText="yo" />);
       expect(getAllByRole('img')[0].firstChild).toHaveAttribute('xlink:href', '#icon-caution-fill.svg');
+      expect(getAllByRole('img')[0]).toHaveAttribute('title', 'Error');
     });
 
     it('shows no icon by default', () => {
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
@@ -83,11 +83,6 @@ describe('AbstractCustomField', () => {
       expect(screen.getByRole('img', { name: 'Label' }).children[1]).toHaveAttribute('xlink:href', '#icon-calendar-fill.svg');
     });
 
-    it('renders the error with aria-label for accessibility', () => {
-      render(<AbstractCustomField {...sharedProps} errorText="yo" />);
-      expect(screen.getByRole('img', { name: 'Error' }).firstChild).toHaveAttribute('xlink:href', '#icon-caution-fill.svg');
-    });
-
     it('shows no icon by default', () => {
       const { queryByRole } = render(<AbstractCustomField {...sharedProps} />);
       expect(queryByRole('img')).toBeNull();
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
@@ -31,7 +31,7 @@ describe('AbstractCustomField', () => {
     it('sets aria-describedby to link to a hint and aria-invalid when there are errors', () => {
       render(<AbstractCustomField {...sharedProps} errorText={'there are errors.'} />);
       expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-describedby', 'test-idHint');
-      expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-invalid', 'true');
+      expect(screen.getByLabelText('Test label')).toBeInvalid();
     });
   });
 
--- a/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
+++ b/src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
@@ -60,10 +60,15 @@ function CustomFieldInputMultipleChoice(props) {
     setAutocompleteValue('');
   }
 
-  function onChoicesClear(event) {
-    event.preventDefault();
+  function clearChoices() {
     setValue([]);
     setExpanded(false);
+    autocompleteRef.current.focus();
+  }
+
+  function onChoicesClear(event) {
+    event.preventDefault();
+    clearChoices();
   }
 
   function onAutocompleteChange(event) {
@@ -151,6 +156,9 @@ function CustomFieldInputMultipleChoice(props) {
                 fill="skip"
                 name={iconClear.id}
                 onClick={onChoicesClear}
+                onEnter={clearChoices}
+                titleAttrText={'Remove selected choice'}
+                tabable={true}
                 ariaLabel={`Remove all selected choices on ${props.label}`}
                 role="button"
               />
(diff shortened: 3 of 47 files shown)


src/components/icon/icon.jsx
```diff
@@ -19,9 +26,10 @@ export default function Icon(props) {
       id={props.id}
       onClick={props.onClick}
       role={props.role}
+      title={props.titleAttrText}
```
juanca, line 29: SVGs don't use the title attribute. They use the nested `<title>` element or the `aria-label` attribute.

> SVG user agents MUST provide an accessible object in the accessibility tree for rendered SVG elements that meet any of the following criteria, unless they are excluded from the accessibility tree per the rules in Excluding Elements from the Accessibility Tree:
> - It has at least one direct child ‘title’ element or ‘desc' element that is not empty after trimming whitespace. User agents MAY include elements with these child elements without checking for valid text content.
> - It has a non-empty (after trimming whitespace) ‘aria-label’ attribute or ‘aria-roledescription’ attribute.
> - It has an ‘aria-labelledby’ attribute or ‘aria-describedby’ attribute containing valid IDREF tokens. User agents MAY include elements with these attributes without checking for validity.

From https://www.w3.org/TR/svg-aam-1.0/

You can test for this too by using user-centric queries (users only click on "visible text" or "screen reader text"):

```
screen.getByRole('img', { name: 'Accessible SVG Name' })
```

More information at https://github.com/testing-library/dom-testing-library/issues/685

src/components/form-control/form-control.jsx
```diff
@@ -37,7 +37,7 @@ export default function FormControl(props) {
         {props.children}
       </div>
       {isInvalid(props.error, props.readOnly) && (
-        <span className={styles['error-message']}>
+        <span id={`${props.id}Hint`} className={styles['error-message']} aria-live="polite" >
```
juanca, line 40: Does the `id` do something in particular in this case? Usually, the ID is associated with another attribute: `label#for` or `element#aria-labelledby`, etc.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
```diff
@@ -258,20 +258,24 @@ describe('src/components/custom-field-input-single-choice/custom-field-input-sin
...
-      userEvent.click(screen.getAllByRole('img')[0]);
-      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveValue('');
+    it('clears a value when enter is pressed on the clear icon and the clear icon can be focused for accessibility', () => {
+      const value = { id: 'some-selection', label: 'Some selection' };
+      render(<CustomFieldInputSingleChoice {...requiredProps} value={value} />);
+      expect(screen.getByLabelText('Test label')).toHaveValue('Some selection');
+      userEvent.click(screen.getByLabelText('Test label', { selector: 'input' }));
+      userEvent.tab();
+      expect(screen.getAllByRole('img')[0].firstChild).toHaveFocus();
+      fireEvent.keyDown(screen.getAllByRole('img')[0].firstChild, { key: 'Enter', code: 'Enter' });
```
juanca, line 277: We will want to avoid using `getAll*` queries whenever possible. Since this is an interactive element, it needs to have an accessible name:

```
screen.getByRole('img', { name: 'le name' })
```

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
```diff
@@ -258,20 +258,24 @@ describe('src/components/custom-field-input-single-choice/custom-field-input-sin
...
-      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveValue('some test text');
-      userEvent.click(screen.getAllByRole('img')[0]);
-      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveValue('');
+    it('clears a value when enter is pressed on the clear icon and the clear icon can be focused for accessibility', () => {
+      const value = { id: 'some-selection', label: 'Some selection' };
+      render(<CustomFieldInputSingleChoice {...requiredProps} value={value} />);
+      expect(screen.getByLabelText('Test label')).toHaveValue('Some selection');
+      userEvent.click(screen.getByLabelText('Test label', { selector: 'input' }));
+      userEvent.tab();
+      expect(screen.getAllByRole('img')[0].firstChild).toHaveFocus();
```
juanca, line 276: `screen.getByRole('img', { name: 'le name' })`

src/components/form-control/form-control.jsx
bennapp: So this is used with `aria-describedby`. If the input has `aria-invalid` and `aria-describedby` points to this span, the screen reader will read the error text when you have focus on the input. Pretty cool!

https://accessibility.huit.harvard.edu/provide-helpful-error-messages

eg.
```
<label for="email">Your email</label>
<input type="text" id="email" aria-invalid="true" aria-describedby="emailHint">
<div class="hint" id="emailHint" style="">Please enter a valid email.</div>
```

src/components/icon/icon.jsx
```diff
@@ -19,9 +26,10 @@ export default function Icon(props) {
       id={props.id}
       onClick={props.onClick}
       role={props.role}
```
juanca, line 28: The role needs to be `button` if we are going to make the SVG interactive. 

Though, another thing to consider: what if the SVG is wrapped in a button? Like another component called `SVGButton` and it is basically

```
function SVGButton() {
  return (
    <button>
      <svg ...>
        ....
      </svg>
    <button>
  );
}
```

Or we can have the logic / validation in the SVG component itself.

src/components/icon/icon.jsx
```diff
@@ -11,6 +11,13 @@ export default function Icon(props) {
     props.currentColor === 'skip' ? '' : styles[`color-${props.currentColor}`],
   ].filter(Boolean);
 
+  const tabindex = props.tabable ? '0' : undefined;
+  const onKeyDown = (event) => {
+    if (event.key === 'Enter') {
```
juanca, line 16: We want to support both: `Enter` and `Space`


https://www.w3.org/TR/wai-aria-practices-1.1/#button
https://www.w3.org/TR/wai-aria-practices-1.1/examples/button/button.html

src/components/icon/icon.jsx
```diff
@@ -11,6 +11,13 @@ export default function Icon(props) {
     props.currentColor === 'skip' ? '' : styles[`color-${props.currentColor}`],
   ].filter(Boolean);
 
+  const tabindex = props.tabable ? '0' : undefined;
```
juanca, line 14: We should base this on `props.role`: if it is interactive then we set a tabindex of `0` otherwise `1`

src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
```diff
@@ -71,10 +77,11 @@ describe('AbstractCustomField', () => {
       expect(getByRole('img')).toBeDefined();
     });
 
-    it('gives preference to the error icon', () => {
+    it('renders the error icon first with an error title for accessibility', () => {
       const icon = <Icon name={calendarSvg.id} currentColor="action" title="Hello" />;
       const { getAllByRole } = render(<AbstractCustomField {...sharedProps} icon={icon} errorText="yo" />);
       expect(getAllByRole('img')[0].firstChild).toHaveAttribute('xlink:href', '#icon-caution-fill.svg');
+      expect(getAllByRole('img')[0]).toHaveAttribute('title', 'Error');
```
juanca, line 84: It would be great to refactor this test to use `getByRole('img', { name: 'Error' })` but since it's pre-existing it's just a nice to have. (We only recently found out about `{name: ''}`)

src/components/form-control/form-control.jsx
juanca: Oh, I think I see a test that ensures this ID is part of some `aria-describedby`. I'll make this as resolved.

src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
```diff
@@ -27,6 +27,12 @@ describe('AbstractCustomField', () => {
       render(<AbstractCustomField {...sharedProps} ariaProps={ariaProps} />);
       expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-controls', 'test-id-label');
     });
+
+    it('sets aria-describedby to link to a hint and aria-invalid when there are errors', () => {
+      render(<AbstractCustomField {...sharedProps} errorText={'there are errors.'} />);
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-describedby', 'test-idHint');
```
juanca, line 33: I think this might be the better matcher: https://github.com/testing-library/jest-dom#tohavedescription

src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
```diff
@@ -27,6 +27,12 @@ describe('AbstractCustomField', () => {
...
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('aria-invalid', 'true');
```
juanca, line 34: This is the better matcher: https://github.com/testing-library/jest-dom#tobeinvalid

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -152,6 +158,9 @@ function CustomFieldInputMultipleChoice(props) {
                 fill="skip"
                 name={iconClear.id}
                 onClick={onChoicesClear}
+                onEnter={clearChoices}
+                titleAttrText={'Remove selected choice'}
```
juanca, line 162: Oh, did we find an issue with `ariaLabel`? We should choose one. On that point, I think it makes sense to keep the `on ${props.label}` since we want to indicate the context for the button.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -43,13 +43,17 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
     setSearchValue(undefined);
+    inputRef.current.focus();
   };
 
   const clearIcon = () => {
     if (!props.readOnly && (value || searchValue)) {
       return (<Icon
         name={iconClear.id}
         onClick={clear}
+        onEnter={clear}
```
juanca, line 54: We can clean up the interface by using more user-centric names: instead of `onClick` and `onEnter` we can say `onPress`.

juanca: Though, to stay backwards compatible, we might want to avoid deprecating `onClick` (yet). (Or we can backfill).

src/components/__internal__/abstract-custom-field/abstract-custom-field.test.jsx
bennapp: Maybe we need to bump testing-library? 
```
TypeError: expect(...).toHaveDescription is not a function
```

src/components/form-control/form-control.test.jsx
```diff
@@ -57,6 +57,16 @@ describe('<FormControl>', () => {
...
       expect(screen.getByText('I am an error message')).toBeInTheDocument();
     });
+
+    it('uses aria-live polite to alert users of an error', () => {
+      render(<FormControl {...requiredProps} error="I am an error message" />);
+      expect(screen.getByText('I am an error message')).toHaveAttribute('aria-live', 'polite');
+    });
+
+    it('uses a hint to alert users of the error message', () => {
+      render(<FormControl {...requiredProps} error="I am an error message" />);
```
cincospenguinos, line 67: I think this prop switched over to `errorText`. Is it still `error` in `FormControl`? If so, we may want to update it as well.

src/components/icon/icon.jsx
cincospenguinos: Instead of calling it SVGButton we may want to call it IconButton

bennapp: I went with making the `tabindex` based on `role == 'button'`. That seems sufficient. Are there other pros of creating a button wrapper?

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -152,6 +158,8 @@ function CustomFieldInputMultipleChoice(props) {
                 fill="skip"
                 name={iconClear.id}
                 onClick={onChoicesClear}
+                onEnter={clearChoices}
+                tabable={true}
```
juanca, line 162: Looks like this isn't used anymore

src/components/icon/icon.jsx
```diff
@@ -11,6 +11,13 @@ export default function Icon(props) {
     props.currentColor === 'skip' ? '' : styles[`color-${props.currentColor}`],
   ].filter(Boolean);
 
+  const tabindex = props.role === 'button' ? '0' : undefined;
```
juanca, line 14: I just realized this might have unintended consequences. 

We have to actively manage the page tab sequence in complex components. For example, in a taglist, the taglist component ensures only 1 element within it has `tabindex` of `0` (else it is `-1`). This ensures that a keyboard-only user can navigate through the page quickly without having to go through each single tag in a large taglist.

Looks like I can't verify this in the staging branch -- we need to go through the PR checklist btw. 

If it's odd, we can discuss how to architect this a bit better -- we already have concepts of a component being "active" or "inactive" which is tied to `tabindex` as well.

src/components/icon/icon.jsx
juanca: I think keeping it simpler is better.

src/components/icon/icon.jsx
```diff
@@ -21,7 +28,7 @@ export default function Icon(props) {
       role={props.role}
     >
       { props.title && <title>{props.title}</title> }
-      <use xlinkHref={`#${props.name}`} />
+      <use tabIndex={tabindex} onKeyDown={onKeyDown} xlinkHref={`#${props.name}`} />
```
juanca, line 31: It's odd that the `onKeyDown` is not onthe same element as the `onClick`. Is there a particular reason for that? Normally, we want to keep them on the same element.

src/components/form-control/form-control.test.jsx
juanca: We can make a new story too to avoid having breaking changes in this particular PR (and thus keep the AC for the story focused)

src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
```diff
@@ -32,6 +32,7 @@ export default function AbstractCustomField(props) {
         className={styles['input-icon']}
         currentColor="caution"
         name={cautionSvg.id}
+        ariaLabel={'Error'}
```
juanca, line 35:
```suggestion
```

As far as I know:

- The error icon isn't interactive (i.e. not focusable) so it does not require a label
- The error icon is already described by the error message so it does not require its own label

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
juanca: Let's not forget to clean this up

juanca:
I can give this another look over once its greened up.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -43,13 +43,17 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
+        ariaLabel={'Remove selected choice'}
```
Bloodyaugust, line 55: Should this say
```suggestion
        ariaLabel={'Clear selected choice'}
```
To keep the nomenclature consistent?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
```diff
@@ -258,20 +258,24 @@ describe('src/components/custom-field-input-single-choice/custom-field-input-sin
...
+      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveFocus();
     });
 
-    it('clears the search value and hides the icon', () => {
-      render(<CustomFieldInputSingleChoice {...requiredProps} />);
-      userEvent.type(screen.getByLabelText('Test label', { selector: 'input' }), 'some test text');
-      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveValue('some test text');
-      userEvent.click(screen.getAllByRole('img')[0]);
-      expect(screen.getByLabelText('Test label', { selector: 'input' })).toHaveValue('');
+    it('clears a value when enter is pressed on the clear icon and the clear icon can be focused for accessibility', () => {
```
Bloodyaugust, line 270: Do we also want an expectation about where the focus moves after clear is pressed?

src/components/icon/icon.jsx
```diff
@@ -80,8 +90,10 @@ Icon.defaultProps = {
   fill: 'none',
   id: undefined,
   onClick: () => {},
+  onEnter: () => {},
   role: 'img',
   size: 'medium',
   stroke: 'none',
   title: undefined,
+  active: true,
```
Bloodyaugust, line 98: This should probably be false, so multiple icons on the same page aren't all active.

src/components/icon/icon.jsx
```diff
@@ -11,18 +11,26 @@ export default function Icon(props) {
...
+        tabIndex={tabindex}
+        aria-label={props.ariaLabel}
+        aria-labelledby={props.ariaLabelledBy}
+        className={classes.join(' ')}
+        id={props.id}
+        onClick={props.onClick}
+        role={props.role}
+      >
+        { props.title && <title>{props.title}</title> }
+        <use onKeyDown={onKeyDown} xlinkHref={`#${props.name}`} />
```
Bloodyaugust, line 32: Perhaps we should only attach the `onKeyDown` handler if `props.role === 'button'` to keep it semantic. If we do make this change, the test should change along with it (and probably a new test should exist to cover the new behavior).

PR #157, opened 2020-07-21
[webinf-i-thought-i-was-only-acting-173779664]

cincospenguinos:
[But it turns out it was all for real](https://www.pivotaltracker.com/story/show/173779664)
[...](https://www.youtube.com/watch?v=sepcj45774I)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-i-thought-i-was-only-acting-173779664
- [ ] Green Bigmaven CI check


src/components/__internal__/no-choice-list-item.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ListOption from '../list-option/list-option.jsx';
+import styles from '../no-choice-list-item.css';
+
+export default function NoChoiceListItem(props) {
```
juanca, line 6: An empty state isn't a list item. It shouldn't be focusable or anything (because otherwise the user might think it is selectable).

src/components/no-choice-list-item.css
```diff
@@ -0,0 +1,7 @@
+@import '../styles/colors-v2.css';
```
juanca, line 1: Why is this CSS file not with its component file? Looks like this might indicate that the no choice list item shouldn't be in `__internal__` or maybe that we need to allow for CSS files to be in `__internal__`?

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -178,6 +179,9 @@ function CustomFieldInputMultipleChoice(props) {
                 {choice.label}
               </ListOption>
             ))}
+            {visibleChoices.length === 0 && (
+              <NoChoiceListItem noOptionText={props.noOptionText} />
```
juanca, line 183: If the "empty state" is not considered a list item then we should not render it in a `Listbox`. We will want to come up with a "popup" component that can render arbitrary components within it. 

To be concrete, it will look something like this:

```
{renderPopup && (
  <Popup>
    {visibleChoices.length === 0 ? (
      <EmptyState text="wuuuuut" /> 
    ) : ( 
      <Listbox>
        ...
      </Listbox>
    )}
  </Popup>
)}

File: src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx [hunk 4]
```diff
@@ -202,6 +207,7 @@ CustomFieldInputMultipleChoice.propTypes = {
 
 CustomFieldInputMultipleChoice.defaultProps = {
   errorText: undefined,
+  noOptionText: 'No options available.',
```
juanca, line 210: It doesn't make sense to be this configurable for the multi choice component since it is always about choices.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
```diff
@@ -1,6 +1,11 @@
 @import '../../styles/colors-v2.css';
 @import '../../styles/spacing.css';
 
+.autocomplete-input {
```
juanca, line 4: I'm confused about the changes to this file. Why are there so many CSS changes for a new sub-component?

src/components/__internal__/no-choice-list-item.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ListOption from '../list-option/list-option.jsx';
+import styles from '../no-choice-list-item.css';
```
juanca, line 4: The styles file for a component should live next to it and follow the same filename pattern. This ensures we're coupling a component to 1 stylesheet and it is always obvious where it should go. When the stylesheet composes from other stylesheets is where the magic happens for associating a component to stylesheets outside its folder.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
cincospenguinos: The selectors weren't alphabetized. I was under the impression that that's the pattern we typically follow, and so I alphabetized all of the CSS rules for us as part of my work

src/components/__internal__/popup/popup.test.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
juanca, line 1: Let's move Popup our of internal. I think it's a component we should be encourage for re-use -- it can be reused in Tooltips, ContextMenus, etc.

src/components/__internal__/no-options/no-options.jsx
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from './no-options.css';
+
+export default function NoOptions(props) {
```
juanca, line 5: I'm also okay with promoting this out of `__internal__`.

src/components/no-options/no-options.css
```diff
@@ -0,0 +1,15 @@
...
+
+.no-options {
+  border: 1px solid var(--mds-grey-38);
+  border-radius: 3px;
+  color: var(--mds-grey-54);
+  display: flex;
+  font: var(--mds-type-content);
+  height: var(--spacing-x-large);
+  justify-content: center;
+  line-height: var(--spacing-x-large); /* stylelint-disable-line mds/typography */
```
Bloodyaugust, line 13:
```suggestion
  align-items: center;
```

We can use this flexbox attribute to do the same thing without needing an exception for `line-height`, which has other behavioral implications we don't want. 😄

src/components/popup/popup.md
```diff
@@ -0,0 +1,9 @@
+The Popup component can be used to show something under a conditional of any sort:
+
+```js
+<Popup show={false}>This text will not show up</Popup>
+```
+
+```js
+<Popup show>This text will show up</Popup>
+```
```
Bloodyaugust, line 9: We should give an example that shows off the usage of `children` elements as well.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.css
```diff
@@ -42,14 +32,15 @@
...
-  pointer-events: none;
+.invalid-container {
+  composes: container;
+  composes: invalid-input from '../__internal__/abstract-custom-field/abstract-custom-field.css';
 }
 
-.clear-icon {
-  composes: icon-base from '../icon/icon.css'; /* stylelint-disable-line css-modules/composed-class-names */
-  cursor: pointer;
+.no-options {
```
juanca, line 40: Should these styling be part of the "No options" component instead?

src/components/popup/popup.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function Popup(props) {
```
juanca, line 4: Should the popup component contain the styling for popups? e.g. absolute positioning, borders, background, etc.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -165,6 +171,7 @@ CustomFieldInputSingleChoice.propTypes = {
   id: PropTypes.string.isRequired,
   choices: PropTypes.arrayOf(ChoiceType),
   label: PropTypes.string.isRequired,
+  noOptionText: PropTypes.string,
```
juanca, line 174: We don't need this since this is already hyper specific to choices.

src/components/popup/popup.md
```diff
@@ -0,0 +1,9 @@
+The Popup component can be used to show something under a conditional of any sort:
```
juanca, line 1: This is an indication that this is _not_ the right abstraction. If the purpose of the component is conditional  rendering then why should we use this instead of the normal React practice of `someBool && <TheComponent />`?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Minor: `CustomFieldInputSingleChoice` and `CustomFieldInputMultipleChoice` now indicates to the user when no options are available
+  - Minor: Add `Popup` component available for use
```
juanca, line 11: I think we're missing a changelog entry for the new "NoOptions" component

src/components/popup/popup.md
cincospenguinos: We ended up deleting this and we will explore making an accessible popup later down the road.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
```diff
@@ -83,10 +83,10 @@ describe('<CustomFieldInputMultipleChoice>', () => {
         expect(screen.queryByRole('listbox')).not.toBeInTheDocument();
       });
 
-      it('does not open when there are no choices', () => {
+      it('informs the user when no choice is available', () => {
```
juanca, line 86: This is odd. How is "no choice" available if it's not a choice? 😜

PR #160, opened 2020-07-22
Webinf you may now so choose one 171681540

Bloodyaugust:
## Motivation
We need to edit single choice fields.

This design-system change brings the `value` of `CustomFieldInputSingleChoice` in line with what the API expects.

## Acceptance Criteria 
- When a custom field is a single choice field, it is displayed as a single choice field in the custom app
- The single choice field displays other things like label attributes
- The single choice field changes can be saved when the field is dirty (and valid)
- The save button can be enabled / disabled when making changes to the choice field

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-you-may-now-so-choose-one-171681540
- [ ] Green Bigmaven CI check

https://www.pivotaltracker.com/story/show/171681540


src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -22,9 +24,10 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
   const [value, setValue] = useState(props.value);
   const [searchValue, setSearchValue] = useState(undefined);
   const inputRef = useRef();
+  const selfRef = ref || useRef();
```
juanca, line 27: We can't be conditionally calling any hook. The `||` is a conditional code flow. 

Instead, we want something like:

```suggestion
  const backupRef = useRef();
  const selfRef = ref || backupRef;
```

Bloodyaugust: Good catch, thanks!

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -118,26 +122,29 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
     setSearchValue(newValue);
     setShowOptions(true);
   }
 
-  useImperativeHandle(ref, () => ({
+  useImperativeHandle(selfRef, () => ({
     id: props.id,
     get value() {
-      return value;
+      return value ? [Number(value.id)] : [];
```
juanca, line 136: Non-blocker but future concern:

I was hoping we would keep symmetry between the `value` prop and `value` ref APIs. I think the current custom field choice server APIs are wrong -- we can't do server side search if the `value` is a list of IDs. Keeping symmetry would also make it easier for developers to grok the `value`s: hey, it's the same as the thing we passed in. 

But we can just keep changing the `prop` and `ref` APIs as we improve the system overall.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
```diff
@@ -99,6 +99,21 @@ describe('src/components/custom-field-input-single-choice/custom-field-input-sin
       userEvent.click(screen.getByLabelText('Test label'));
       expect(screen.queryByText('No options available.')).not.toBeInTheDocument();
     });
+
+    it('does not break if the choices are changed', () => {
+      const { rerender } = render(<CustomFieldInputSingleChoice {...requiredProps} choices={[]} />);
+      userEvent.click(screen.getByLabelText('Test label'));
+      expect(screen.getByText('No options available.')).toBeInTheDocument();
+
+      rerender(<CustomFieldInputSingleChoice {...requiredProps} choices={[{ id: '1', label: 'yo' }]} />);
```
juanca, line 108: Do we need to assert that the new choices are available?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
Bloodyaugust: I agree on all counts, and think we should address this once we have a chance to really sit down and compile/enact changes on the API side, especially if it's us owning a new API endpoint special made for Custom Apps.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.test.jsx
Bloodyaugust: That's a good idea, as it is it only tests for an error being thrown. Your suggestion will make the intent more specific!

PR #161, opened 2020-07-24
Bump testing library; Use toHaveDescription

bennapp:
## Motivation
Latest and greatest baby! (and to use toHaveDescription 😆 )

## PR upkeep checklist

- [ ] Change log entry N/A
- [x] Label(s)
- [x] Assignee(s)
- [ ] Deployment URL: https://mavenlink.github.io/design-system/$BRANCH N/A


bennapp:
Something is wrong with some usages of `not.toBeInTheDocument()`, so I am digging into that now.

package.json
```diff
@@ -33,10 +33,10 @@
     "@babel/plugin-proposal-class-properties": "^7.2.3",
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^10.0.6",
-    "@testing-library/react-hooks": "^3.2.1",
-    "@testing-library/user-event": "^12.0.7",
+    "@testing-library/jest-dom": "^5.11.1",
```
juanca, line 36: the major version bump is a bit scary but if we're green then it's good. do we have any deprecation warnings? or extra output when running tests? it might be a good opportunity to clean it all up

juanca:
We might want to see what is the breaking change between 4.x.x and 5.x.x

bennapp:
Not exactly sure what this means but
https://github.com/testing-library/jest-dom/releases/tag/v5.0.0
Remove extend-expect typings (#182) (c76f8c5), closes #182
BREAKING CHANGES
moving typings to @types/testing-library__jest-dom
This version does not includes types. The next patch release v5.0.1 will. If you need the type definitions please do not update to v.5.0.0.

bennapp:
One thing that is a very odd change is that now `userEvent.tab();` is not triggering `focusin` so two tests in single and multi choice are broken. I can dig a little bit more into the release notes to try to see why this is the case. But I am also considering deleting those tests >.<

juanca:
I can dig into it a bit too. I would prefer to keep the tests around but if it's too difficult we can delete.

PR #166, opened 2020-07-30
Calendar component

juanca:
https://www.pivotaltracker.com/story/show/174021505

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/calendar-component-static
- [x] Green Bigmaven [CI check](https://github.com/mavenlink/mavenlink/pull/20975)
- [x] Added component to internal design system


jest.config.transform.svg.js
```diff
@@ -1,13 +1,22 @@
 /* eslint-disable import/no-commonjs, no-unused-vars, prefer-template */
 
+const fs = require('fs');
 const path = require('path');
 
 module.exports = {
   process(src, filename, config, options) {
     // import iconCautionSvg from 'src/svgs/icon-caution-fill.svg';
```
cincospenguinos, line 8: Given that this is commented out, I think we can safely remove these comments. If we need that code in someway, it's always in our version control system.

Unless `process` processes the code included in the block, which if that's the case, don't worry about this comment.

src/components/calendar/calendar.jsx
```diff
@@ -0,0 +1,199 @@
...
+
+  return (
+    <th className={styles.weekday}>
+      {weekday}
+    </th>
+  );
+}
+
+function getCellClassName(iterator, month, highlightedDate) {
+  if (
```
cincospenguinos, line 34: Could we extract some of these out? How about

```suggestion
const matchesYear = highlightedDate.getFullYear() === iterator.getFullYear();
const matchesMonth = highlightedDate.getMonth() === iterator.getMonth();
const matchesDate = highlightedDate.getDate() === iterator.getDate();

if (matchesYear && matchesMonth && matchesDate) {

```

src/components/calendar/calendar.jsx
```diff
@@ -0,0 +1,199 @@
...
+  if (iterator.getMonth() === month) return styles.date;
+
+  return styles['not-current-date'];
+}
+
+function getCell(iterator, month, highlightedDate) {
+  const date = iterator.getDate();
+  const className = getCellClassName(iterator, month, highlightedDate);
+
+  // This needs to occur at the end of this function
```
cincospenguinos, line 49: I'm pretty cool with this comment and the one above, but would it be possible to include the behavior this comment points out in a test? If so, would it be safe to remove this comment/allow `iterator.setDate` to occur in a different order of other future changes? Either way, a spec ensuring this behavior would be ideal, although a comment like this is significant enough for me to let it go.

src/components/calendar/calendar.jsx
```diff
@@ -0,0 +1,199 @@
...
+      {date}
+    </td>
+  );
+}
+
+function Calendar(props) {
+  const highlightedDate = new Date(props.value ? `${props.value}T00:00` : Date.now());
+  const [year, setYear] = useState(highlightedDate.getFullYear());
+  const [month, setMonth] = useState(highlightedDate.getMonth());
+  // const [date] = useState(highlightedDate.getDate());
```
cincospenguinos, line 63: Could we remove this line?

src/components/calendar/calendar.jsx
```diff
@@ -0,0 +1,199 @@
...
+            {getHeadCell(headIterator)}
+            {getHeadCell(headIterator)}
+            {getHeadCell(headIterator)}
+            {getHeadCell(headIterator)}
+            {getHeadCell(headIterator)}
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            {getCell(iterator, month, highlightedDate)}
```
cincospenguinos, line 132: It looks like we're copypasta-ing here. Could we extract this to a separate function that operates in the same way? Something like

```javascript
function foo() {
  const foo = [];

  for (let i = 0; i < 7; i++) {
    foo.push(getCell(iterator, month, highlightedDate));
  }

  return foo;
}

/* ... */

{ foo() }

jest.config.transform.svg.js
juanca: This is a little bit of tricky configuration -- it's an area of our codebase that is poorly understood. I am opting to leave the code in for future Jest config tweakers -- it has been a

src/components/calendar/calendar.jsx
juanca: Yeah, a test would be ideal. Lemme see if I can pull something together for it.

src/components/calendar/calendar.jsx
juanca: @bennapp did you end up using this variable for the active index management? (or did we decide to explore using it?)

src/components/calendar/calendar.jsx
juanca: I just realized we already have test coverage with our snapshots tests. I'll go ahead and delete these!

src/components/calendar/calendar.jsx
juanca: I'm okay with figuring out how to do this a little bit more concisely. However, I want to wait until we implement interactivity since we might need to add more logic around the rendering process.

src/components/calendar/calendar.jsx
juanca: I hoisted something similar. I'm still not sure the value it provides.

src/components/calendar/calendar.jsx
juanca: Either way, @bennapp I removed this. If there is a merge conflict, feel free to bring it back.

PR #167, opened 2020-08-03
Add aria-haspopup to single and mulit choice inputs for accessibility

bennapp:
## Motivation
https://www.pivotaltracker.com/story/show/173953747

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/$BRANCH
- [ ] Green Bigmaven CI check


bennapp:
The change log has: `- Minor: `CustomFieldInputSingleChoice` and `CustomFieldInputMultipleChoice` accessibility improvements.`

I might just piggy-back on that sentiment for this effort. Unless we feel like we need to comment on `aria-haspopup` specifically.

bennapp:
@juanca should be all good on the aria-expanded front.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
```diff
@@ -56,6 +57,17 @@ describe('<CustomFieldInputMultipleChoice>', () => {
...
+
+      it('has aria-haspopup for accessibility', () => {
+        render(<CustomFieldInputMultipleChoice {...requiredProps} />);
+        expect(screen.getByLabelText('test label', { selector: '[aria-haspopup="listbox"]' })).toBeInTheDocument();
+      });
+
+      it('has aria-expanded for accessibility', () => {
+        render(<CustomFieldInputMultipleChoice {...requiredProps} />);
+        userEvent.click(screen.getByText('test label'));
+        expect(screen.getByLabelText('test label', { selector: 'ul[aria-expanded=true]' })).toBeInTheDocument();
```
bennapp, line 69: @juanca is this the best way to select the element?

src/components/listbox/listbox.jsx
```diff
@@ -123,6 +124,7 @@ Listbox.propTypes = {
   children: PropTypes.node,
   className: PropTypes.string,
   labelledBy: PropTypes.string.isRequired,
+  ariaExpanded: PropTypes.bool,
```
juanca, line 127: I think we shoudn't highlight the "aria" about this prop, instead favoring a more "standard" phrasing. I think this would encourage others to hook more "normal props" to their aria equivalents. i.e. let's rename `ariaExanded` to `expanded`.

bennapp: I can buy it, done!

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -169,7 +169,7 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
         choices.length === 0 ? (<NoOptions className={styles['no-options']} />) : (
           <Listbox
             className={styles.dropdown}
-            ariaExpanded={true}
+            expanded={true}
```
juanca, line 172: Oh, I just realized this is always expanded. I now wonder if this has any meaning to the a11y tree since it is always `true`. 

But at least it is part of the Listbox API!

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
juanca: If possible, I would make the selector as simple as possible but then assert something about the attribute.

```
expect(screen.getByRole('listbox')).toHaveAttribute('aria-expanded', 'true');
```

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
juanca: I think we might not be using the `aria-expanded` attribute correctly here.

https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

The `aria-expanded` is used on the element that has a popup. The `aria-expanded` indicates to the accessibility tree that the popup is either expanded (visible) or collapsed (not visible)_.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -135,6 +135,8 @@ function CustomFieldInputMultipleChoice(props) {
             {!props.readOnly && (
               <input
                 aria-labelledby={`${props.id}-label`}
+                aria-haspopup={'listbox'}
+                aria-expanded={renderPopup}
```
juanca, line 139: Looks like we will need to add `role="combobox"` to the input:

https://www.w3.org/TR/wai-aria/#combobox

It makes sense since we want the voice reader to know that this is a "smart widget"

juanca: Also, I just followed the linter error:

https://circleci.com/gh/mavenlink/design-system/2331?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/role-supports-aria-props.md
https://www.w3.org/TR/wai-aria/#aria-expanded
https://www.w3.org/TR/wai-aria/#combobox

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -135,6 +135,11 @@ function CustomFieldInputMultipleChoice(props) {
...
+                aria-autocomplete={'list'}
+                aria-controls={`${props.id}-multi-choice-listbox`}
```
juanca, line 141: This makes sense to me now. 

I used to be confused as to why `aria-haspopup` would only indicate a role but never the actual element. This was the missing relationship attribute.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -135,6 +135,11 @@ function CustomFieldInputMultipleChoice(props) {
             {!props.readOnly && (
               <input
                 aria-labelledby={`${props.id}-label`}
+                aria-haspopup={'listbox'}
```
juanca, line 138: BTW: you can use strings in JSX:
```suggestion
                aria-haspopup="listbox"
```

PR #170, opened 2020-08-06
Webinf why choose just one 171681550

Bloodyaugust:
## Motivation
We need to edit fields.

## Acceptance Criteria 
- There is an editable multiple choices field in internal MDS
- When a custom field is a multiple choices field, it is displayed as a multiple choices field in the custom app
- The multiple choices field displays other things like label attributes
- The multiple choices field changes can be saved when the field is dirty (and valid)
- The save button can be enabled / disabled when making changes to the choice field

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-why-choose-just-one-171681550
- [x] Green Bigmaven CI check

https://www.pivotaltracker.com/story/show/171681550


src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -26,18 +28,18 @@ function getClassName(readOnly, errorText) {
...
-function CustomFieldInputMultipleChoice(props) {
+const CustomFieldInputMultipleChoice = forwardRef((props, ref) => {
   const autocompleteRef = useRef();
   const [autocompleteValue, setAutocompleteValue] = useState('');
   const [expanded, setExpanded] = useState(false);
   const [value, setValue] = useState(props.value);
-  const visibleChoices = props.choices
-    .filter(choice => choice.label.includes(autocompleteValue))
-    .filter(choice => !value.includes(choice));
+  const [visibleChoices, setVisibleChoices] = useState(getVisibleChoices());
```
juanca, line 36: I don't really see understand visible choices needs to be a stateful property. It looks like this is computed from value and autocompleteValue. Making it stateful seems to just add a bunch of code around managing it.

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.test.jsx
```diff
@@ -183,6 +183,29 @@ describe('<CustomFieldInputMultipleChoice>', () => {
...
+      expect(screen.queryByRole('option', { name: 'Answer to everything' })).toBeInTheDocument();
+
+      rerender(<CustomFieldInputMultipleChoice
+        {...requiredProps}
+        choices={choices}
+        value={choices}
+      />);
+
+      userEvent.click(screen.getByText('test label'));
+      expect(screen.queryByRole('option', { name: 'Answer to everything' })).not.toBeInTheDocument();
```
juanca, line 207: Do we want to add an assertion that the new selected choice (via `value`) is part of the selected choices UI?

Bloodyaugust: Good idea. 👍

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -57,7 +65,8 @@ function CustomFieldInputMultipleChoice(props) {
     const selectedChoiceIndex = choicesRefs.findIndex(choiceRef => choiceRef === event.target);
     const selectedChoice = visibleChoices[selectedChoiceIndex];
     setExpanded(false);
-    setValue([...value, selectedChoice]);
+    const newValue = [...value, selectedChoice].sort((a, b) => a.id - b.id);
```
juanca, line 68: From a UI perspective, we might not want to sort the selected choices. It might cause "unexpected" behavior since it is not apparent that a choice was actually selected if it gets sorted in a large list. However, I think this is something we can iterate on instead of holding up this PR -- possibly yet another pain point with the custom fields API.

Bloodyaugust: Agreed, it kinda sux. Perhaps there's something to be done like only sorting the `value` return, but that does feel like concealing the real problem, which is that the API sorts the selected values instead of maintaining their order.

PR #175, opened 2020-08-12
Ensure markdown paragraphs are styled correctly

juanca:
Looks like Styleguidist might have a bug since it is not re-using the theme fontSizes

https://www.pivotaltracker.com/story/show/174214846
https://www.pivotaltracker.com/story/show/174263131

## PR upkeep checklist

- [x] Change log entry N/A
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/fix-site-typography
- [x] Green Bigmaven CI check N/A

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -92,6 +92,13 @@ module.exports = {
     TableRenderer: path.join(__dirname, 'styleguide/components/table'),
   },
   styleguideDir: 'build',
+  styles: {
+    Para: {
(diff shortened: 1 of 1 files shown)


juanca:
I reported the issue @ https://github.com/styleguidist/react-styleguidist/issues/1661

I wonder if they'll fix it. If they do, we can clean up our config to be more concise.

PR #177, opened 2020-08-17
MDS Linter: spacing variables

bennapp:
## Motivation
https://www.pivotaltracker.com/story/show/172321583

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-typo-fix-172321583
- [ ] Green Bigmaven CI check


src/linters/spacing.js
```diff
@@ -0,0 +1,112 @@
...
+module.exports = stylelint.createPlugin(ruleName, (primary, secondary, context) => {
+  return (root, result) => {
+    root.walkDecls((declaration) => {
+      const property = declaration.prop;
+
+      if (properties.includes(property)) {
+        const valueTokens = declaration.value.split(' ');
+
+        const includesCalc = declaration.value.includes('calc');
+        if (includesCalc) {
```
juanca, line 53: It could be nice to lint within `calc`s too! What was the reasoning to avoid them for this linter?

src/linters/spacing.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        const validSpacingValue = valueTokens.every((token) => {
+          const includesSpacingVariable = validSpacingSettings.find((spacingVariable) => {
+            return token.includes(`var(${spacingVariable})`);
+          });
+          return includesSpacingVariable || token === '0' || !Object.keys(fixmap).includes(token);
+        });
+
+        if (!validSpacingValue) {
```
juanca, line 64: We normally prefer to use positive assertions in the `if` when there is an `else`. 

For this instance, is this variable misnamed? `!!validSpacingValue` is coded to report a violation. ?

juanca:
Looks like there are linting errors in this PR 😅

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mavenlink/design-system",
-  "version": "0.36.1",
+  "version": "0.36.3",
```
cincospenguinos, line 3: We may need to bump this version further, since 0.38 has been released

src/linters/spacing.js
cincospenguinos: We could also have the function return the opposite of what's given, perhaps using `some` instead of `every`, and then rename the variable to `invalidSpacingValue`

src/linters/spacing.js
```diff
@@ -0,0 +1,112 @@
...
+            fixes.push([declaration, valueToFix]);
+
+            let shouldImport = true;
+            try {
+              require.resolve('@mavenlink/design-system');
+            } catch (e) {
+              shouldImport = false;
+            }
+
+            const importString = '@import \'@mavenlink/design-system/src/styles/spacing.css\';';
```
cincospenguinos, line 83: I think this is where double quotes could be ideal:

```suggestion
const importString = "@import '@mavenlink/design-system/src/styles/spacing.css';";
```
```

File: src/linters/spacing.js [hunk 5]
```diff
@@ -0,0 +1,112 @@
...
+          } else {
+            const violation = { ruleName, node: declaration, result };
+            stylelint.utils.report({ ...violation, message: messages.rejected });
+          }
+        }
+      }
+    });
+
+    // eslint-disable-next-line prefer-arrow-callback
+    return new Promise(function (resolve) {
```
cincospenguinos, line 96: Any reason why we went with `function` here instead of `() => {}`?

src/linters/spacing.js
```diff
@@ -0,0 +1,112 @@
...
+      // eslint-disable-next-line prefer-arrow-callback
+      setTimeout(function () {
+        // console.log(fixes);
```
cincospenguinos, line 99: This line needs to be removed, since we're not console logging anything.

src/linters/spacing.test.js
```diff
@@ -0,0 +1,84 @@
...
+describe('src/linters/spacing', () => {
+  it('errors on a mds variable value', async () => {
+    const code = '.className { margin: 2px; }';
+
+    await stylelint.lint(configuration({ code })).then((data) => {
+      expect(data.errored).toBe(true);
+    });
+  });
+
+  it('does not errors on a value differnet from a mds variable value', async () => {
```
cincospenguinos, line 26:
```suggestion
  it('does not errors on a value different from an MDS variable value', async () => {
```

src/linters/spacing.js
bennapp: So I am not sure why, but the changes to the files won't be made without using `function` keyword. This is only the case in practice but not in testing 🤔

package.json
```diff
@@ -70,6 +70,7 @@
     "webpack": "^4.28.3"
   },
   "dependencies": {
-    "prop-types": "^15.6.2"
+    "prop-types": "^15.6.2",
+    "string.prototype.matchall": "^4.0.2"
```
juanca, line 74: It looks like `matchall` is polyfilled with corejs 3.4 but we use corejs 2 in bigmaven. :( 

I am going to see if we can bump it -- it might be locked due to our NodeJS version.

bennapp:
This passed visual testing in Bigmaven. I think it is ready!!

src/linters/spacing.js
```diff
@@ -0,0 +1,124 @@
...
+module.exports = stylelint.createPlugin(ruleName, (primary, secondary, context) => {
+  return (root, result) => {
+    root.walkDecls((declaration) => {
+      const property = declaration.prop;
+
+      if (properties.includes(property)) {
+        const currentValue = declaration.value;
+
+        const invalidSpacingValue = Object.keys(fixmap).some((spacingVariable) => {
+          const pixelValueMatches = [...matchAll(currentValue, /\d+px/g)];
```
juanca, line 54: I am not sure if we want to include yet another library as a dependency for this use case. It looks like `matchAll` is just providing a different return structure:

```
const regexp = /\d+px/g;
const str = 'blue 10px black 11px';

console.log([...str.matchAll(regexp)]); // Array [Array ["10px"], Array ["11px"]]
console.log(str.match(regexp)) // Array ["10px", "11px"]
```

I would prefer to use ES5 native methods instead of pulling in another library. Do we think we can refactor this to use `match`?

package.json
juanca: It looks like this is more work that a quick upgrade: https://github.com/mavenlink/mavenlink/pull/21288 The extra features from core-js 3+ is creating a different testing environment. I think because of that, we will want to fix the tests and then let it bake on team stagings for 2 weeks.

src/linters/spacing.js
bennapp: The structure that `matchAll` returns is quite useful here because it includes additional information that we need. Specifically, that is the index of the match.

bennapp: Yes, we can use match but we will be nearly re-implementing `matchAll`.

PR #182, opened 2020-08-25
MDS Custom Field Date component uses Calendar component

bennapp:
## Motivation
[MDS Custom Field Date component uses Calendar component](https://www.pivotaltracker.com/story/show/173315112)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-calendar-for-cf-date-173315112/
- [ ] Green Bigmaven CI check


src/components/calendar/calendar.jsx
```diff
@@ -102,8 +102,14 @@ function Calendar(props) {
     setActiveDate(newDate);
   }
 
+  function handleDateSelection(date) {
+    props.onDateSelected(date);
```
Bloodyaugust, line 106: We should probably swap this line with the one below, or move it into a separate `useImperativeHandle` so that retrieving the value with a future ref `value` API will receive the new value, not the old one.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -78,41 +88,84 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
...
+    setExpanded(!expanded);
+  }
+
+  function onInputKeyDown(event) {
+    if (event.key === 'Enter' || event.key === 'Space') {
+      setExpanded(!expanded);
+    }
+  }
+
+  function onKeyDown(event) {
```
Bloodyaugust, line 111: Should we also capture `keyDown` events from the `Calendar`, to support this same effect when the `Calendar` is focused? 🤔

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -159,6 +160,37 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
     });
   });
 
+  describe('calendar icon', () => {
+    it('opens to the correct date on click and closes when clicking away', () => {
```
Bloodyaugust, line 164: Do we have coverage for this functionality when the user clicks into the text area of the input as well?

bennapp: We have coverage for this now!

src/components/calendar/calendar.jsx
bennapp: I think the date passed in here is the new value.

src/components/custom-field-input-date/custom-field-input-date.jsx
bennapp: We do handle this when the calendar is focused; that is accomplished with `handleDropdownClose`.

bennapp:
@juanca and @Bloodyaugust this is ready for review! ty ty!

src/components/calendar/calendar.jsx
juanca: It looks like we don't have an imperative handle. This can be refactored when we do implement an imperative handle.

Basically, what @Bloodyaugust was trying to point out:

The state of the component is _not_ up-to-date when we call `props.onDateSelected(date)`. Yes, we did use the new `date` value. But the component's state is not representative of it.

This is evident when we expose the value of the calendar on an imperative handle (maybe we will never do this?):

- A React component is basically a state machine
- On an event handler, all we should ever be doing is setting state: `setSelectedDate(date)`
- As a state transition (`useEffect(..., [state])`), we should be calling external callbacks: `props.onDateSelected(...)`

```jsx
function Calendar(props, ref) {
  const [selectedDate, setSelectedDate] = useState()

  useImperativeHandle(ref, () => {
    selectedDate: selectedDate,
  })

  useEffect(() => {
    props.onSelectedDateChange(selectedDate)
  }, [selectedDate])

  function handleSelectedDateChange(event) {
    setSelectedDate(event.target.value)
  }

  return (
    <SomeJsX />
  )
```

src/components/calendar/calendar.jsx
```diff
@@ -76,6 +76,24 @@ function Calendar(props) {
...
+  useEffect(() => {
+    if (selectedDate && defaultDate && !isSameDate(selectedDate, defaultDate)) {
+      setSelectedDate(defaultDate);
+      setFocusedDate(defaultDate);
+
+      setYear(highlightedDate.getFullYear());
+      setMonth(highlightedDate.getMonth());
+      setActiveDate(defaultDate);
+    }
+  }, [defaultDate]);
```
juanca, line 88: We might want to only run this when we get a new `props.value` instead of a new `new Date` object. Every date object is different.

e.g. 

```
> new Date() === new Date()
< false
```

src/components/calendar/calendar.jsx
```diff
@@ -76,6 +76,24 @@ function Calendar(props) {
...
+
+  useEffect(() => {
+    if (activeDate.getMonth() !== month || activeDate.getFullYear() !== year) {
+      setMonth(activeDate.getMonth());
+      setYear(activeDate.getFullYear());
+    }
+  }, [activeDate]);
```
juanca, line 95: Let's double check whether we need this useEffect.

- I think the Calendar was already working well without it (since this updates the highlighted date when the active date changes)
- I think updating the highlighted date state is already taken care of in the useEffect for `props.value` (above).

src/components/custom-field-input-date/custom-field-date.css
```diff
@@ -0,0 +1,21 @@
+@import '../../styles/spacing.css';
+@import '../../styles/colors-v2.css';
+
+.input-icon {
+  composes: input-icon from '../__internal__/abstract-custom-field/abstract-custom-field.css'; /* stylelint-disable-line css-modules/composed-class-names */
```
juanca, line 5:
```suggestion
  composes: input-icon from '../__internal__/abstract-custom-field/abstract-custom-field.css';
```

src/components/custom-field-input-date/custom-field-date.css
```diff
@@ -0,0 +1,21 @@
...
+  pointer-events: auto;
+  height: 18px;
+  width: 18px;
+}
+
+.date-input input::-webkit-inner-spin-button,
```
juanca, line 11: Does this work in Chrome, Firefox, Safari, and/or Edge?

src/components/calendar/calendar.jsx
Bloodyaugust: Thanks for explaining better than I did, @juanca. 😅

src/components/calendar/calendar.jsx
Bloodyaugust: That, or if we want to maintain running it every time the `render` is invoked, we could just move it to an IIFE.

src/components/custom-field-input-date/custom-field-date.css
Bloodyaugust: I'm gonna guess it doesn't. Do we not have an autoprefixer for these "experimental" selectors?

src/components/custom-field-input-date/custom-field-date.css
```diff
@@ -12,3 +13,9 @@
   display: none;
   -webkit-appearance: none;
 }
+
+.calendar-container {
+  position: absolute;
+  background-color: var(--white);
+  z-index: 2147483647;
```
Bloodyaugust, line 23: Magic numbers on `z-index` always make me nervous. If there's a future where for instance we want to display a tooltip while hovering over the calendar, this might make it impossible (or at least very difficult, with a need to move it to a different z-stack).

I think it's fine for now, but we should probably have a map of `z-index` values somewhere in documentation and backfill for components like this.

Bloodyaugust (approved):
LGTM! Nice changes on the calendar styling, and it sounds like we wanna implement the `forwardRef` and `useImperativeHandle` edition later, so the stateful stuff can wait. I'll let Juanca decide on if his reqs are blocking or no.

src/components/calendar/calendar.jsx
bennapp: This is needed when a user changes the date-input, we want the calendar's month / year to respond accordingly.

src/components/custom-field-input-date/custom-field-date.css
bennapp: It works on chrome, firefox, safari (I haven't yet tested Edge)

bennapp: It works in Edge too 🥳

PR #183, opened 2020-09-01
Icon v2

juanca:
https://www.pivotaltracker.com/story/show/173758226

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/icon-v2
- [x] Green Bigmaven CI check [PR](https://github.com/mavenlink/mavenlink/pull/21326)


bennapp:
So is there a corresponding bigmaven PR where we convert all usages of Icon to fit the new props format? Or will that be a follow-up?

juanca:
Yup! Already building: https://github.com/mavenlink/mavenlink/pull/21326

.eslintrc.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
     env: {
       jest: true,
     },
+    rules: {
+      'jsx-a11y/label-has-for': 'off',
```
bennapp, line 24: This is being deprecated in favor of: label-has-associated-control. 

Is that why we are turning it off here?

bennapp:
Gotcha, so their currently are no uses of IconV2 in bigmaven then?

styleguide.config.js
```diff
@@ -131,5 +135,18 @@ module.exports = {
   },
   title: 'Mavenlink Design System',
   usageMode: 'expand',
+  updateExample(props = {}, exampleFilePath) {
```
bennapp, line 138: What is going on here? :)

bennapp (approved):
LGTM. Just had a few small questions for my own understanding.

.eslintrc.js
juanca: No but that's a good point.

I am disabling the a11y rule for test files. When we write inline React in our tests for example / test code, we don't really care about setting up the "perfect" a11y component since we're testing particular aspects of a component. I turned it off so we don't waste time with test data.

juanca:
> Gotcha, so their currently are no uses of IconV2 in bigmaven then?

Yeah, so far there isn't uses of IconV2 outside our Custom UI code.

styleguide.config.js
juanca: I wanted to make it easier for developers to understand how to apply custom class names to our components.

https://mavenlink.github.io/design-system/icon-v2/#/Components/IconButton

In these examples, I have "an example stylesheet" which is used in the code examples. Hopefully, it is easier to connect the dots on how we are customizing the components (with `composes` and all!)

PR #188, opened 2020-09-09
Add className API to Tag, fix input sizing for multi-choice

Bloodyaugust:
## Motivation
When a selection is made on the multi choice component, the extra spacing below the selection remains after the drop down is closed

## Acceptance Criteria 
- When enough selections are made that the input wraps to multiple lines and the user removes all choices from wrapped lines
  - the input does not show unnecessary whitespace

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-space-me-baby-one-more-time-174498864
- [x] Green Bigmaven CI check N/A


juanca:
I like this solution. Let's double check if this is fine:

Screen Shot 2020-09-09 at 3 35 24 PM

juanca:
Or, I think we can provide a min-width to the input. But, if we do that, we will get the "weird double line" render that was original reported.

juanca (approved):

PR #189, opened 2020-09-14
Form control component improvements for generalized MDS Form

cincospenguinos:
Implementing a Form component:

- dirty aware*
- validity aware
- submission aware

Implementing a changes to our custom fields to better support a generalized form component:

- name ref API: similar to HTML's name, it is the key for the control value when the form serializes itself
- dirty ref API: determines whether the control has a different state value that its provided value

Misc improvements:

- multi-choice can only determine a new value if the serialized array has changed because two different arrays can actually have the same internal values

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-hey-this-is-a-nice-addition-right
- [ ] Green Bigmaven CI check


juanca (approved):

juanca (changes requested):
Looks like we will need to do more changes to accommodate the new MDS Form

juanca:
@cincospenguinos Can you review this PR?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   - Minor: Add `className` and `classNameReadOnly` to `Tag` component API
   - Minor: Fix styling for input of `CustomFieldMultipleChoice` component
   - Minor: Custom Field Date Input is editable and uses the Calendar component
+  - Patch: Perform simple refactor to avoid react re-rendering issue on `CustomFieldInputMultiChoice`
```
cincospenguinos, line 13: Don't we need to add a note for creating the form component?

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -1,91 +1,107 @@
...
-    const tree = renderer.create(<CustomFieldInputDate label="Hello" id="hello" />).toJSON();
-    expect(tree).toMatchSnapshot();
+    const ref = createRef();
+    render(<CustomFieldInputDate {...requiredProps} ref={ref} />);
+    expect(document.body).toMatchSnapshot();
+    expect(ref.current).toMatchSnapshot();
+  });
+
+  describe('dirty ref API', () => {
+    xit('updates on user interactions', () => {
```
cincospenguinos, line 31: Is this still supposed to be skipped?

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
```diff
@@ -106,15 +106,19 @@ const CustomFieldInputMultipleChoice = forwardRef((props, ref) => {
...
 
   useEffect(() => {
-    props.onChange(selfRef.current);
+    props.onChange({ target: selfRef.current });
   }, [value]);
 
   useEffect(() => {
     setValue(props.value);
-  }, [props.value]);
+  }, [props.value.map(choice => choice.id).join('')]);
```
cincospenguinos, line 114: 🎉

cincospenguinos (commented):
Wow looks good to me, man! I didn't think about imagining each individual field as being dirty/clean, and leveraging our existing ref API to incorporate that. My only beef isn't with this PR or even with our solution with custom fields--I just feel a need to make a `useImperativeHandle` factory that generates that custom field API according to what type it is, since we have to maintain that API across all five (six?) of our fields.

Anyway, I can't approve it so I'll just leave this comment.

src/components/custom-field-input-date/custom-field-input-date.test.jsx
juanca: I haven't found a good way to test this. We do have a follow up story to integrate the component into our custom form. I think that will be a good time to deal with this -- nice and focused.

juanca:
> existing ref API to incorporate that. My only beef isn't with this PR or even with our solution with custom fields--I just feel a need to make a `useImperativeHandle` factory that generates that custom field API according to what type it is, since we have to maintain that API across all five (six?) of our fields.

We can play with the factory idea but I think ultimately we will still need to type out the logic for each. Keeping it in each component file might be the most straight forward solution.

juanca (approved):
LGTAndre.

src/components/custom-field-input-date/custom-field-input-date.test.jsx
juanca: /cc @bennapp .

PR #193, opened 2020-09-24
Select Component and CustomFieldInputSingleChoice Refactor

Bloodyaugust:
## Motivation
We want a single select for arbitrary data. It would be really great if we could leverage the work from CustomFieldInputSingleChoice. I know the single choice component has some specific code to deal with custom field logic, so if we were to leverage the component code, we need to figure out which parts belong in MDS (Generic) Select and which parts belong in MDS CustomFieldInputSingleChoice

## Acceptance Criteria 
- MDS Select
  - `name` prop and ref API (symmetry)
  - `value` prop and ref API (symmetry)
  - `onChange` prop API (called when value state changes)
  - `required` prop API
  - `children` prop API (as their options)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-select-all-the-things-174890763
- [x] Green Bigmaven CI check

https://www.pivotaltracker.com/story/show/174890763

### DIFF
--- a/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
+++ b/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
@@ -1,3 +1,0 @@
-.container {
-  position: relative;
-}
--- a/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
+++ b/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
@@ -11,6 +11,7 @@
   background-color: var(--white);
   position: absolute;
   width: 100%;
+  z-index: 10;
 }
 
 .no-options {
--- a/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
+++ b/src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
@@ -1,29 +1,3 @@
-@import '../../styles/colors-v2.css';
-@import '../../styles/spacing.css';
-
 .container {
   position: relative;
-}
-
-.dropdown {
-  composes: container from '../listbox/listbox.css';
-  box-sizing: border-box;
-  background-color: var(--white);
-  position: absolute;
-  width: 100%;
-  z-index: 10;
-}
-
-.no-options {
-  composes: dropdown;
-  composes: no-options from '../no-options/no-options.css';
-}
-
-.input-icon {
-  composes: input-icon from '../__internal__/abstract-custom-field/abstract-custom-field.css';
-}
-
-.clear-button {
-  composes: button from '../icon-button/icon-button.css';
-  padding: var(--spacing-small);
 }
--- a/src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
+++ b/src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
@@ -8,7 +8,6 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 import Select from '../select/select.jsx';
-import styles from './custom-field-input-single-choice.css';
 import ListOption from '../list-option/list-option.jsx';
 
 const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleChoice(props, ref) {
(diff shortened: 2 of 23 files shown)


src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
```diff
@@ -1,28 +1,3 @@
-@import '../../styles/colors-v2.css';
-@import '../../styles/spacing.css';
-
 .container {
```
juanca, line 1: Do we still need this around for single choice? Looks like Select implements the same rule on its root node

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -154,43 +72,21 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
 
   return (
     <div ref={wrapperRef} className={props.className}>
```
juanca, line 74: Looks like we don't use `wrapperRef` anymore. We might even be able to throw away the className too.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -154,43 +72,21 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
-            refs={refs}
-            value={value}
-          >
-            { listOptions(choices) }
-          </Listbox>
-        )
-      )}
+        value={value}
+      >
+        { listOptions(choices) }
```
juanca, line 88: Can we inline the choices generation or simplify the chain of functions into just 1 function call? i think we have a chain of functions that only really ever get called once -- an artifact of probably the previous large single choice component.

src/components/select/select.jsx
```diff
@@ -0,0 +1,231 @@
...
+
+  useEffect(() => {
+    props.onChange({ target: selfRef.current });
+  }, [value]);
+
+  useEffect(() => {
+    updateOptionVisibility();
+  }, [searchValue, props.listOptionRefs]);
+
+  useEffect(() => {
```
juanca, line 147: Looking at this file, there is still a feeling of "woah, there's a lot going on here" -- and I think there isn't a way around it. But, I think an idea that might help:

Looks like we only use `updateOptionSelected` in this `useEffect`. We can just inline the function. Something like:

```suggestion
  useEffect(function updateOptionSelected() {
    ...
  }, [value, showOptions]);
```

juanca: If we like the idea, let's do the same for the other useEffects.

juanca (approved):
Overall, looks great. I think there are some refactors for the sake of readability / maintainability.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.css
Bloodyaugust: Good point, with that the whole style file can be removed.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
Bloodyaugust: I can see removing the `wrapperRef`, but I feel like we should maintain the `className` API to provide for style overrides in the future, and to keep parity with the other components.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
Bloodyaugust: Can do!

src/components/select/select.jsx
Bloodyaugust: I'll try this fo sho

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -216,7 +103,7 @@ CustomFieldInputSingleChoice.propTypes = {
 
 CustomFieldInputSingleChoice.defaultProps = {
   choices: [],
-  className: styles.container,
+  className: '',
```
juanca, line 106: Setting this to an empty string will override the default Select className. Is that intended?

Bloodyaugust: Noooope, definitely not! Fixing...

PR #195, opened 2020-09-24
Webinf improve date ref

bennapp:
## Motivation
We need the imperative ref of date input to be consistent with the others and pass reference to it on callback.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/custom-field-input-date/custom-field-input-date.jsx
+++ b/src/components/custom-field-input-date/custom-field-input-date.jsx
@@ -44,6 +44,7 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
   const [isValid, setIsValid] = useState(isValueValid(currentValue, props.errorText, true));
   const [expanded, setExpanded] = useState(false);
   const [shouldFocusInput, setShouldFocusInput] = useState(false);
+  const [initialValue, setInitialValue] = useState(true);
 
   useEffect(() => {
     if (isEditing && inputRef.current) {
@@ -58,14 +59,6 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
 
     setShouldFocusInput(false);
   }, [shouldFocusInput]);
-
-  useEffect(() => {
-    if (ref && currentValue) {
-      props.onChange({ target: ref.current });
-    } else if (currentValue) {
-      props.onChange();
-    }
-  }, [currentValue]);
 
   function onDateSelected(date) {
     setShouldFocusInput(true);
@@ -89,6 +82,18 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
       return undefined;
     },
   }));
+
+  useEffect(() => {
+    if (!initialValue) {
+      if (ref) {
+        props.onChange({ target: ref.current });
+      } else {
+        props.onChange(currentValue);
+      }
+    } else {
+      setInitialValue(false);
+    }
+  }, [currentValue]);
 
   const handleOnChange = (event) => {
     if (inputRef.current) {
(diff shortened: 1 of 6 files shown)


src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -59,6 +59,14 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
     setShouldFocusInput(false);
   }, [shouldFocusInput]);
 
+  useEffect(() => {
+    if (ref && currentValue) {
```
juanca, line 63: I have a sneaky suspicion that we still want to call `props.onChange` when `currentValue` is undefined or an empty string.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -59,6 +59,14 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
...
+      props.onChange({ target: ref.current });
+    } else if (currentValue) {
```
juanca, line 65: I'm not sure we will ever get into a situation when the state value of the component changes but it is not rendered.

src/components/custom-field-input-date/custom-field-input-date.jsx
bennapp: So the problem with undefined, is that it this will get called on first render, and that was what we were attempting to guard against here ... Empty string should def pass through though.

src/components/custom-field-input-date/custom-field-input-date.jsx
bennapp: So I think the user does not have to pass a ref through, that is was this case is for. Not that it is not rendered.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -59,6 +59,14 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
     setShouldFocusInput(false);
   }, [shouldFocusInput]);
 
+  useEffect(() => {
+    if (ref && currentValue) {
+      props.onChange({ target: ref.current });
```
juanca, line 64: I think we can get coverage on the parameters for `onChange`. At the moment, it looks like the tests only check that it was called. We know that we need this specific signature to make it work with Form.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -68,12 +76,17 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
 
   useImperativeHandle(ref, () => ({
     get dirty() {
-      return props.value !== this.value;
+      return convertToFormat(props.value, 'yyyy-mm-dd') !== this.value;
```
juanca, line 79: The tests for this are `xit`ed at the moment. When I tried writing them, I was getting hung up on trying to type into the date input. Maybe we can give it another go or try setting value without typing.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -68,12 +76,17 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
...
     },
     id: props.id,
     name: props.name,
     get value() {
```
juanca, line 83: I think getting spec coverage with the new functionality will be useful. We want to make sure that the value is serializable and consistent regardless of the state of the component.

juanca (changes requested):
1. I think we're missing test coverage
2. I think there might be some redundant conditionals in the useEffects.

bennapp:
@juanca I got some solid test coverage now on the onChange options and the ref interface 🥳

juanca (approved):

PR #201, opened 2020-10-01
Webinf date input feedback #174936567

bennapp:
## Motivation
Design feedback for the calendar + date input

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
   - Minor: Add new `Select` component
   - Minor: Refactor `CustomFieldInputSingleChoice` to use `Select`
   - Minor: Add new `setSelected` and `setVisible` methods to `ListOption` ref API
+  - Minor: `CustomFieldInputDate` supports placeholder, truncates month to 3 characters, and uses Open Sans
```
cincospenguinos, line 14: For a second I thought this said "uses Comic Sans" and I was concerned

cincospenguinos (approved):
LGTM! I'm glad we're not using comic sans.

src/components/custom-field-input-date/custom-field-date.css
```diff
@@ -1,12 +1,18 @@
...
 @import '../../styles/colors-v2.css';
+@import '../../styles/typography-v2.css';
 
 .date-input input::-webkit-inner-spin-button,
 .date-input input::-webkit-calendar-picker-indicator {
   display: none;
   -webkit-appearance: none;
 }
 
+.date-input input {
```
juanca, line 11: This doesn't have bad implications in this use-case but FYI: we like to avoid using nested generic selectors -- we can just boil this down to using nested selectors.

The reason it is _not good_ to use generic selectors: it discourages composition. 

If we apply styling to all nested HTML `<input>`s in a component, if the component accepted a React node as a child and rendered it, we could be introducing styling pollution across components. 

If we apply styling to all nested HTML `<input>`s in a class name, if the class name is composed into another component, we could be introducing styling pollution to its own HTML markup. 

It is usually better to styling each individual element via CSS modules. We want to encourage composition, and the less gotchas the better.

CHANGELOG.md
```diff
@@ -20,6 +22,7 @@
   - Minor: Add new `Select` component
   - Minor: Refactor `CustomFieldInputSingleChoice` to use `Select`
   - Minor: Add new `setSelected` and `setVisible` methods to `ListOption` ref API
+</details>
```
juanca, line 25: What is this?

juanca (changes requested):
Changelog is formatted incorrectly.

bennapp:
@juanca fixed the changelog and css selector. I would appreciate if you could take another look, thanks!

juanca (approved):

PR #205, opened 2020-10-02
Deprecate `inputRef` APIs on custom fields

### LINKED ISSUE #198: MDS Feedback
MDS CustomFieldInputText:

- `ref` vs `inputRef` (do we need `inputRef` anymore?)

juanca:
We want to encourage pure usages of the React ref component APIs. If we need to reach deeper into a component, it needs to be part of the components `ref` prop API.

Closes #198 

https://www.pivotaltracker.com/story/show/175104430

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds-refs
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/21718

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,13 +40,13 @@ commands:
 jobs:
   build_and_test:
     docker:
-      - image: cypress/base:10.18.0
+      - image: cypress/base:14
     working_directory: ~/repo
     steps:
       - checkout
       - install_dependencies:
           image: "cypress"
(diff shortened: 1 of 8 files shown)


juanca:
> This is to make the custom field API more closely match the React ref API. Is that right?

Yeah! We don't want to expose many refs. A custom field component only really have 1 public ref API.

PR #209, opened 2020-10-05
Internal MDS inputs look good when required

Bloodyaugust:
## Motivation
Look like there is some CSS pollution on our MDS inputs when applied to bigmaven.

## Acceptance Criteria 
- When the input is `required`, the input label uses the MDS red text color
- When the input is `required`, it uses the same styling as the external MDS inputs

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-we-require-more-minerals-175021353
- [x] Green Bigmaven CI check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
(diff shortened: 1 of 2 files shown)


cincospenguinos (approved):
LGTM!

PR #213, opened 2020-10-08
Form `autoSave` prop API

juanca:
https://www.pivotaltracker.com/story/show/175171363

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/form-auto-save/

### DIFF
--- a/src/components/form/form.jsx
+++ b/src/components/form/form.jsx
@@ -19,6 +19,7 @@ const Form = React.forwardRef((props, forwardedRef) => {
     props.onChange(event);
     setValid(ref.current.checkValidity());
     setDirty(ref.current.dirty);
+    if (props.autoSave) onSubmit(event);
   }
 
   function onSubmit(event) {
@@ -68,7 +69,7 @@ const Form = React.forwardRef((props, forwardedRef) => {
   return (
     <form ref={formRef} onSubmit={onSubmit} onChange={onChange}>
       {props.children({ onChange })}
-      {!props.readOnly &&
+      {!props.readOnly && !props.autoSave &&
         <div className={styles['buttons-container']}>
           <Button
             className={styles['primary-button']}
@@ -85,17 +86,19 @@ const Form = React.forwardRef((props, forwardedRef) => {
 });
 
 Form.propTypes = {
-  submitText: PropTypes.string,
+  autoSave: PropTypes.bool,
   children: PropTypes.func.isRequired,
   onChange: PropTypes.func,
   onSubmit: PropTypes.func,
   readOnly: PropTypes.bool,
-  refs: PropTypes.arrayOf(
-    PropTypes.shape({ current: PropTypes.any }).isRequired,
-  ).isRequired,
+  refs: PropTypes.arrayOf((
+    PropTypes.shape({ current: PropTypes.any }).isRequired
+  )).isRequired,
+  submitText: PropTypes.string,
 };
 
 Form.defaultProps = {
+  autoSave: false,
   submitText: 'Save',
   onChange: () => {},
   onSubmit: () => {},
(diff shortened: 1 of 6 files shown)


src/components/form/form.jsx
```diff
@@ -15,27 +17,35 @@ const Form = React.forwardRef((props, forwardedRef) => {
...
+              ...data,
+              [controlRef.current.name]: controlRef.current,
+            };
+          }
+
+          return data;
+        }, {}),
+      });
+    }
+  }, 300, { leading: !props.autoSave, trailing: props.autoSave }), [props.onSubmit]);
```
jarednielson, line 36: Whoever uses this will have to be careful. If their `onSubmit` prop is generated every time their component re-renders then we will be creating new debounce functions here and will lose the debounciness of it all. One possible solution would be to do something like `const memoizedOnSubmit = useCallback(() => props.onSubmit(), []). I guess at that point just removing `props.onSubmit` from the `onDebouncedSubmit` dependency array would get you the same thing. There would be subtle bugs if the `props.onSubmit` function relied on closures.

juanca: I had that originally but found it was buggy since the `onSubmit` callback _can_ rely on closures. 

Instead, calculating a new debounce when the closure changes should be fine. 

1. Create a reference to `onSubmit` (with a dependency on the component's state)
1. Parent passes a reference to `Form`, `Form` caches the debounced function
1. Form invokes the debounce function because the user is typing
1. The reference to `onSubmit` is not yet called, the parent component has no reason to believe it needs to change _anything_
1. Enough time passes by, Form invokes the reference to `onSubmit`. 
1. The parent updates _something_. It most likely created a new closure for the `onSubmit`
1. Parent passes a reference to the new `onSubmit`. `Form` caches the debounced function.
1. ...

There is a low likelihood the parent will re-render its children while the debounce function is being debounced. The user is in the middle of typing stuff so it doesn't make sense that the parent component will re-render.

However, if for some reason the parent does re-render then it's up to the implementation to memoize their own `onSubmit` that will not cause new debounced function. 

Does that make sense? 

Maybe a good way to think about it is that it's buggy either way? At least the onus is on the parent component when we debounce new `onSubmit`s

juanca: TODO: Add documentation about the gotchas.

Bloodyaugust: Keeping the responsibility on the parent component does seem like the lesser of the two evils. Keeps this component as composable(and its behavior reliable) as possible.

src/components/form/form.jsx
```diff
@@ -85,17 +95,19 @@ const Form = React.forwardRef((props, forwardedRef) => {
...
-  submitText: PropTypes.string,
+  autoSave: PropTypes.bool,
   children: PropTypes.func.isRequired,
   onChange: PropTypes.func,
   onSubmit: PropTypes.func,
   readOnly: PropTypes.bool,
-  refs: PropTypes.arrayOf(
-    PropTypes.shape({ current: PropTypes.any }).isRequired,
-  ).isRequired,
+  refs: PropTypes.arrayOf((
```
Bloodyaugust, line 103: Why do we need the extra parens here?

Bloodyaugust (approved):
LGTM! Super clean implementation, and being that there are some unavoidable gotchas, lets make sure implementers know about them in the docs.

src/components/form/form.jsx
juanca: It's a way to get around the "trailing comma" lint for function parameters.

e.g.

```js
someFunction(
  param1,
);
```

vs

```js
someFunction((
  param1
));
```

This is usually pretty annoying with arrow functions:

```jsx
someArray.map(element => 
  <div>{element}</div>,
)
```

vs

```js
someArray.map(element => (
  <div>{element}</div>
))
```

PR #217, opened 2020-10-12
Fix issue where presentDateString was not considering the user's timezone

bennapp:
## Motivation
Timezone bug for date input component 
https://www.pivotaltracker.com/story/show/175209475

## Acceptance Criteria
The presented date should match that date of the value.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-timezoner-175209475
- [x] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/21873

### DIFF
--- a/src/components/custom-field-input-date/custom-field-input-date.jsx
+++ b/src/components/custom-field-input-date/custom-field-input-date.jsx
@@ -36,7 +36,8 @@ function presentDateString(dateString) {
     return '';
   }
 
-  const date = new Date(Date.parse(dateString));
+  const timezone = new Date().toLocaleTimeString('en-us', { timeZoneName: 'short' }).split(' ')[2];
+  const date = new Date(`${dateString}  00:00:00 ${timezone}`);
   return date.toLocaleDateString(undefined, { month: 'short', year: 'numeric', day: 'numeric' });
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Patch: Fix timezone issue with `CustomFieldInputDate`
 </details>
 
 ## v0.48.0


juanca:
Can you break it down for me? How does this "fix" the bug? I can't really tell what is going on, especially since there isn't a test (and I am not sure if it's possible to test this?)

juanca:
Also, can you finish the PR upkeep list?

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -36,7 +36,8 @@ function presentDateString(dateString) {
     return '';
   }
 
-  const date = new Date(Date.parse(dateString));
+  const timezone = new Date().toLocaleTimeString('en-us', { timeZoneName: 'short' }).split(' ')[2];
+  const date = new Date(`${dateString}  00:00:00 ${timezone}`);
   return date.toLocaleDateString(undefined, { month: 'short', year: 'numeric', day: 'numeric' });
```
bennapp, line 41: @juanca, so we are using `toLocaleDateString` to present the date as a string, Eg. 'May 10, 1992'. However, we made the interface for this component such that the input is a date in the format, "YYYY-MM-DD". The downside to this is there is no timezone information.

`new Date(Date.parse(dateString))` will assume that the hour and minutes are `00:00` and the timezone is UTC. If we call `toLocaleDateString` we could see a different date than the date of the users timezone.

bennapp:
I am guessing its possible to test this regression but I don't think its worth it.

src/components/custom-field-input-date/custom-field-input-date.jsx
bennapp: So using Javascript's `Date` is really nice here; it would just be a lot nicer if we also used a javascript date object as the prop for value.

juanca: Do we think we will have the same issue in Calendar?

https://github.com/mavenlink/design-system/blob/82662853c4b4cf99cad5f8356b4690c7410c7ee4/src/components/calendar/calendar.jsx#L67

https://github.com/mavenlink/design-system/blob/82662853c4b4cf99cad5f8356b4690c7410c7ee4/src/components/calendar/calendar.jsx#L74

Also, how are you verifying the changes work? Is there a trick to changing the timezone on your development environment or testing environment?

bennapp: I don't think that has the same issue with the way we pull the date number out of the date object in calendar.

I was kind of lucky and was fixing this bug during the 9 hour-ish window where I think it can occur. But I am sure you could do some clock-trickery to replay the bug.

juanca:
I think it might be pretty easy. It looks like we set our dev timezone to UTC for consistency.

https://github.com/mavenlink/design-system/blob/82662853c4b4cf99cad5f8356b4690c7410c7ee4/jest.config.globalSetup.js#L2

I think we should add a comment about how to choose a different timezone to see if there are timezone issues in the test.

For example, if my computer is running PST, which timezone should I choose to see if there is an issue?

juanca:
Do we want a bigmaven CI check for this PR? Or a staging check?

juanca (approved):
Here's to hoping!

PR #224, opened 2020-10-16
(allow-customizable-text-for-tags)

jdechol:
## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/tag/tag.jsx
+++ b/src/components/tag/tag.jsx
@@ -86,7 +86,7 @@ const Tag = forwardRef(function Tag(props, forwardedRef) {
     >
       <span
         id={contentId}
-        className={styles.content}
+        className={props.classNameText}
         ref={contentRef}
         role="gridcell"
         tabIndex={isActive && tabActiveStates[0] ? '0' : '-1'}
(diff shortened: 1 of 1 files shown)


juanca:
We will need a test for the new prop. 

Outta curiosity, what is the use-case? I want to double check it through my design team .

jdechol:
We were spiking on how to customize text for the tag component. Is there a way spike on local changes to mds without pointing bigmaven to a branch? After discussing with Nick in design, we are going to hold off for now and just use the text styling as is.

juanca:
> Is there a way spike on local changes to mds without needing to push a branch? 

You can always push a branch and _not_ make a PR. But a PR allows it to auto-deploy to Github pages. Up to you. 

As for local changes, there is a [`yarn link` feature](https://classic.yarnpkg.com/en/docs/cli/link/). I have used this in the past and it _kinda_ works. I say kinda because Webpack does not de-duplicate the installed dev dependencies in the targeted library folder. This has caused problems in the past with "singleton"-like behavior -- e.g. Redux. Since MDS is purely presentational, it shouldn't be a problem here 🤞

PR #225, opened 2020-10-16
MDS Input v3

juanca:
https://www.pivotaltracker.com/story/show/175255573

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds-input-fo-sho/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/21935


src/components/input/input.jsx
```diff
@@ -1,69 +1,104 @@
...
       required={props.required}
     >
       <input
+        aria-describedby={`${props.id}Hint`}
         autoFocus={props.autoFocus} // eslint-disable-line jsx-a11y/no-autofocus
-        className={getClassName(props.className, props.invalid, props.readOnly)}
-        defaultValue={props.defaultValue}
-        disabled={props.disabled}
+        className={getClassName(props.className, validationMessage)}
+        defaultValue={props.value}
```
juanca, line 71: We are no longer supporting a controlled input. We are always rendering an uncontrolled input -- but we update the internal `<input>` state whenever a new `value` prop is given. 

If the developer wants to hoist input state then they can listen to events -- either `onChange` or handle bubbled events. Same for all other events. If a developer want to avoid hoisting then they can use the `ref` API on the Input.

Bloodyaugust (approved):
LGTM after pair review!

PR #234, opened 2020-10-23
Open sans: 400, 600

juanca:
https://www.pivotaltracker.com/story/show/175338092

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/open-sans/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/22022


juanca:
Getting this to work well with PostCSS was a little headache inducing. But here is my best attempt at a concise font guide:

https://github.com/fontsource/fontsource/issues/77

juanca:
It looks like bigmaven is running out of memory when processing the new font. 🤔 I may have to revisit the bundling strategy.

styleguide.config.js
```diff
@@ -96,43 +96,18 @@ module.exports = {
...
-  },
+  styles: path.resolve(__dirname, './styleguide.config.styles.js'),
   template: {
     favicon: 'favicon.ico',
     head: {
       links: [
         {
           rel: 'stylesheet',
-          href: 'https://fonts.googleapis.com/css?family=Merriweather|Open+Sans',
+          href: 'https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;600&display=swap',
```
juanca, line 106: While we can set up MDS to process its own fonts, this does not work in our bigmaven codebase. Looks like our bigmaven Webpack configuration runs out of memory processing the fonts -- I think it's more about SASS and outdated dependencies but cest la vie.

package.json
```diff
@@ -72,5 +73,6 @@
   "dependencies": {
     "lodash.debounce": "^4.0.8",
     "prop-types": "^15.6.2"
-  }
+  },
+  "peerDependencies": {}
```
Bloodyaugust, line 77: Seems like this may have been added to, then a dependency removed? Either way, we should nuke it now.

styleguide.config.styles.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
```
Bloodyaugust, line 1: Looks like we didn't migrate sidebar background color, is this intentional?

Bloodyaugust (approved):
LGTM with a couple of minor things!

package.json
juanca: It's auto generated with `yarn add` and `yarn remove`. I'm okay with leaving it for the sake of never seeing it again in a future diff.

styleguide.config.styles.js
juanca: Oh! Good catch. Let me see if it's broken with / without it.

juanca: I think we're good. 

image

We used to have a background color because the old way of having a sticky header + left nav was some fixed positioning. When scrolling, it's transparent background was bleeding other background. But it seems fine from the screenshot now that we use a CSS grid to manage our top-level page sections.

PR #237, opened 2020-10-30
Scope date picker component to the min and max dates of the good data api

Bapark:
…

Co-authored-by: Brian Park <bpark@mavenlink.com>

## Motivation
https://www.pivotaltracker.com/story/show/175469414

## Acceptance Criteria
The date picker component handles it's own errors based on the good data api max and min instead of throwing and our apologies alert from a 422 response

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Deployment URL: https://mavenlink.github.io/design-system/add-good-data-max-and-min-to-date-validation
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/custom-field-input-date/custom-field-input-date.jsx
+++ b/src/components/custom-field-input-date/custom-field-input-date.jsx
@@ -146,7 +146,7 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
     const dateMilliseconds = Date.parse(currentValue);
     const date = new Date(dateMilliseconds);
 
-    if (date > apiLimits.max || date < apiLimits.min) {
+    if (date > props.max || date < props.min) {
       return `Date must be between ${toLocaleDate(apiLimits.min)} and ${toLocaleDate(apiLimits.max)}`;
     }
     return props.errorText;
@@ -287,8 +287,8 @@ CustomFieldInputDate.propTypes = {
 CustomFieldInputDate.defaultProps = {
   disabled: false,
   errorText: '',
-  min: undefined,
-  max: undefined,
+  min: apiLimits.min,
+  max: apiLimits.max,
   onChange: () => {},
   placeholder: undefined,
   required: false,
--- a/src/components/custom-field-input-date/custom-field-input-date.jsx
+++ b/src/components/custom-field-input-date/custom-field-input-date.jsx
@@ -9,8 +9,8 @@ import useDropdownClose from '../../hooks/use-dropdown-close.js';
 import Calendar from '../calendar/calendar.jsx';
 
 const apiLimits = {
-  min: new Date('1900-01-01'),
-  max: new Date('2050-12-31'),
+  min: '1900-01-01',
+  max: '2050-12-31',
 };
 
 const isValidInput = (value) => {
@@ -74,6 +74,8 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
   const [expanded, setExpanded] = useState(false);
   const [shouldFocusInput, setShouldFocusInput] = useState(false);
   const [initialValue, setInitialValue] = useState(true);
+  const minDate = new Date(Date.parse(props.min));
+  const maxDate = new Date(Date.parse(props.max));
 
   useEffect(() => {
     if (isEditing && inputRef.current) {
@@ -146,8 +148,8 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
     const dateMilliseconds = Date.parse(currentValue);
     const date = new Date(dateMilliseconds);
 
-    if (date > props.max || date < props.min) {
-      return `Date must be between ${toLocaleDate(apiLimits.min)} and ${toLocaleDate(apiLimits.max)}`;
+    if (date > maxDate || date < minDate) {
+      return `Date must be between ${toLocaleDate(minDate)} and ${toLocaleDate(maxDate)}`;
     }
     return props.errorText;
   };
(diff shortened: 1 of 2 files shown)


src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -216,11 +231,11 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
         <AbstractCustomField
           {...sharedProps}
           defaultValue={currentValue}
-          errorText={errorText()}
+          errorText={isValid ? '' : errorText()}
           id={props.id}
           key={`${props.id}-editing`}
-          min={formatDateString(props.min)}
-          max={formatDateString(props.max)}
+          min={formatDateString(apiLimits.min)}
```
cincospenguinos, line 237: Instead of ensuring min/max all the time, I think it may be best to allow setting a min/max via props and then set the default value to be the limit instead. That way we can still use this component as a date picker with min/max values that an engineer can set, or the default one for good data.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -216,11 +231,11 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
...
+          max={formatDateString(apiLimits.max)}
```
cincospenguinos, line 238: See above

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -254,4 +255,16 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
       expect(inputRef.current.value).toBe('2016-07-18');
     });
   });
+
+  describe('API limits', () => {
+    xit('shows the expected error text when a date is out of range', () => {
```
cincospenguinos, line 260: Is this still supposed to be skipped?

cincospenguinos (commented):
A few changes requested

src/components/custom-field-input-date/custom-field-input-date.jsx
Bapark: We were going to ask about this since this is a custom fields date picker (and will always have to conform to the good data api) and not a generic date picker. We went with this way for now but definitely were considering just validating on the min max props

src/components/custom-field-input-date/custom-field-input-date.test.jsx
Bapark: So we are having trouble getting userEvent typing to fill out the date field via typing in a date. Due to the nature of the date and multi / single select components, testing them has given us some problems. If you look up in the file there is another suspended spec that you all wrote that maybe due to a similar problem? userEvent.type seems to not be updating the text field. Would love your help in figuring this out, and then we can un pend both specs in this file.

src/components/custom-field-input-date/custom-field-input-date.jsx
Bloodyaugust: It definitely feels odd to implement it with such a tight integration to an API that we don't own, but if we need to do this for every `CustomFieldInputDate` anyway, this might be better.

Bloodyaugust: Perhaps the "meet in the middle" is to move it to props, and have the default value for those props be what the gooddata API expects? 🤔

Bapark: Im down with that. It seems like these custom fields components have an identity crisis of being generic and specific to mavenlink custom fields. Do you all think they are still a good fit for MDS?

Bloodyaugust: @Bapark They definitely aren't generic, and are only for Custom Fields. That being said, I suppose there is an identity crisis, but more on the side of "Custom Fields API vs Gooddata API/integration" IMHO.

Bapark: @Bloodyaugust The Bigmaven custom field API actually throws errors if the provided date is out of this date range (the bigmaven API itself is limited due to the good data limitation) I guess what I mean by an 'identity crisis' is with this component, is the consumer of this component managing state validity or is this component managing itself? I will hit you guys up in slack and we can think about it to arrive in a good spot.

juanca: We will end up extracting a generic MDS Date from CustomFieldInputDate in the near future.

We can hard-code the min/max in CustomFieldInputDate but leave MDS Date with the generic props interface. 

I wonder if there is a pattern for bigmaven date validations. Or maybe this brings up a bigger question whether bigmaven should have consistent min/max values for date attributes.

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -129,11 +136,21 @@ const CustomFieldInputDate = forwardRef(function CustomFieldInputDate(props, ref
...
   const errorText = () => {
     if (!isValid && !isValidInput(props.value)) {
       return `"${props.value}" is an invalid date`;
     }
 
+    const dateMilliseconds = Date.parse(currentValue);
+    const date = new Date(dateMilliseconds);
+
+    if (date > maxDate || date < minDate) {
+      return `Date must be between ${toLocaleDate(minDate)} and ${toLocaleDate(maxDate)}`;
```
juanca, line 152: We will want to leverage the native validation messages. Otherwise, we'll end up in a weird spot for localization. 

What are the validation messages with `min` / `max` dates?

src/components/custom-field-input-date/custom-field-input-date.test.jsx
juanca: Oh right, we also had trouble with `type="date"`. I think we need to open up an issue in `@testing-library/user-event`. 

In the mean time, I think we got test coverage by using the calendar pop-up to change the dates. 

@bennapp should be able to add a bit more detail.

src/components/custom-field-input-date/custom-field-input-date.jsx
Bapark: After a little deliberation we are going to move to a solution to what Ben posted in MDS slack. We will just manage the error text in the parent component for now and probably close this PR.

PR #239, opened 2020-11-03
[webinf-take-it-from-your-number-175444629] Number component

cincospenguinos:
So listen up, [here's the story](https://www.pivotaltracker.com/story/show/175444629)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-take-it-from-your-number-175444629
- [x] Green Bigmaven CI check


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
   - Patch: Ensure `Calendar` day headings are center aligned
+  - Major: Create new `Number` component and integrate it with `CustomFieldInputNumber`
```
juanca, line 11: Theoretically, if there aren't any breaking changes then it would just be a minor change. But I haven't read through the PR to see what is breaking yet.

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
```diff
@@ -40,11 +34,11 @@ exports[`CustomFieldInputNumber has defaults 1`] = `
...
 exports[`CustomFieldInputNumber has defaults 2`] = `
 Object {
-  "dirty": false,
+  "dirty": true,
   "focus": [Function],
   "id": "test-input",
   "name": "field-id",
   "validity": ValidityState {},
-  "value": "",
+  "value": "NaN",
```
juanca, line 42: Do we want this change? It kinda seems weird. What do you think?

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
```diff
@@ -40,11 +34,11 @@ exports[`CustomFieldInputNumber has defaults 1`] = `
 
 exports[`CustomFieldInputNumber has defaults 2`] = `
 Object {
-  "dirty": false,
+  "dirty": true,
```
juanca, line 37: Why is the field dirty by default?

src/components/number/number.css
```diff
@@ -0,0 +1,41 @@
+@import '../../styles/colors-v2.css';
+@import "../../styles/spacing.css";
+@import "../../styles/typography.css";
+
+.input {
```
juanca, line 5: Do we want to compose from the existing styling in `Input`? 

I think the only difference is the padding? I'm not even sure tbh.

src/components/number/number.css
```diff
@@ -0,0 +1,41 @@
...
+  border: 1px solid var(--mds-grey-12);
+  border-radius: 3px;
+  box-sizing: border-box;
+  font-family: var(--mavenlink-type-font-family); /* stylelint-disable-line mds/typography */
```
juanca, line 9: Do we need to skip the linter here?

src/components/number/number.jsx
```diff
@@ -0,0 +1,130 @@
...
+          label={validationMessage}
+        />
+      )}
+    </FormControl>
+  );
+});
+
+Number.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
```
juanca, line 102: We do not want a `disable` prop interface. We decided as a group to stay away from `disabled` in favor of `readOnly`.

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+  describe('placeholder API', () => {
+    it('allows setting placeholder text', () => {
+      render(<Number {...requiredProps} placeholder="Hello!" />);
+      expect(screen.getByLabelText('Test Component')).toHaveAttribute('placeholder', 'Hello!');
+    });
+  });
+
+  describe('required API', () => {
+    it('can be set', () => {
```
juanca, line 91: Let's add an assertion on the validation message. Something like:

```
expect(...).toHaveDescription(...);
```

It tests that the input and the field control components are synced.

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+      render(<Number {...requiredProps} placeholder="Hello!" required />);
+      expect(screen.getByLabelText('Test Component')).toBeRequired();
+      expect(screen.getByLabelText('Test Component')).toBeInvalid();
+    });
+
+    it('can be unset', () => {
```
juanca, line 97: Let's add an assertion on the validation message. Something like:

```
expect(...).toHaveDescription(...);
```

It tests that the input and the field control components are synced.

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+    it('set on the ref', () => {
+      const ref = createRef();
+      render(<Number {...requiredProps} ref={ref} value={10101} />);
+      expect(ref.current.value).toBe(10101);
+      userEvent.type(screen.getByLabelText('Test Component'), '2');
+      expect(ref.current.value).toBe(101012);
+    });
+
+    it('indicates if it is invalid on mount', () => {
```
juanca, line 154: Let's add an assertion on the validation message. Something like:

```
expect(...).toHaveDescription(...);
```

It tests that the input and the field control components are synced.

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
cincospenguinos: I think the value of the number custom field defaults to NaN when the value is empty, and that's because of its ref. I think that's fine? I don't know. What do you think?

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
cincospenguinos: I think it is. That's something I'll have to dig at

src/components/number/number.jsx
cincospenguinos: The disabled prop is required by CustomFieldInputNumber. Should I just have it reroute over to `readOnly` instead?

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+import userEvent from '@testing-library/user-event';
+import Number from './number.jsx';
+
+describe('Number', () => {
+  const requiredProps = {
+    id: 'test-component',
+    label: 'Test Component',
+  };
+
+  afterEach(cleanup);
```
juanca, line 16: AFAIK, we don't actually need to cleanup anything. This is automatic in our environment.

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+  describe('disabled API', () => {
+    it('permits disabling', () => {
+      render(<Number {...requiredProps} disabled />);
+      expect(screen.getByLabelText('Test Component')).toBeDisabled();
+    });
+  });
+
+  describe('label API', () => {
+    it('presents a label', () => {
+      render(<Number {...requiredProps} />);
```
juanca, line 34: We need to use the label to ensure we're asserting some implementation detail. 🤔

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+      expect(screen.getByLabelText('Test Component')).toBeInTheDocument();
+    });
+  });
+
+  describe('ref API', () => {
```
juanca, line 39:
```
get dirty() {
      return props.value !== this.value;
    },
    focus() {
      inputRef.current.focus();
    },
    name: props.name,
    get value() {
      return parseInt(inputRef.current.value, 10);
    },
    get validity() {
      return inputRef.current.validity;
    },
```

Looks like we're missing spec coverage for this describe block. We need name and value coverage too!

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+  afterEach(cleanup);
+
+  it('has defaults', () => {
+    const ref = createRef();
+    render(<Number {...requiredProps} ref={ref} />);
+    expect(document.body).toMatchSnapshot();
+    expect(ref.current).toMatchSnapshot();
+  });
+
+  describe('disabled API', () => {
```
juanca, line 25: While we are moving away from `disabled`, this test coverage is incomplete. When we are testing a boolean, we need to test the `true` and `false` cases. Since it's only two possible states, it's reasonable to get an exhaustive test suite on it.

src/components/number/number.css
cincospenguinos: It's exactly the same styling. Should I pull it back into a separate shared styling sheet?

src/components/number/number.css
cincospenguinos: Yeah the linter gets mad without it for some reason

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+  describe('placeholder API', () => {
+    it('allows setting placeholder text', () => {
+      render(<Number {...requiredProps} placeholder="Hello!" />);
+      expect(screen.getByLabelText('Test Component')).toHaveAttribute('placeholder', 'Hello!');
+    });
+  });
+
+  describe('required API', () => {
+    it('can be set', () => {
+      render(<Number {...requiredProps} placeholder="Hello!" required />);
```
juanca, line 92: We dont' need the placeholder prop here.

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,188 @@
...
+    });
+
+    it('can be unset', () => {
+      render(<Number {...requiredProps} validationMessage="" />);
+      expect(screen.getByLabelText('Test Component')).toBeValid();
+      expect(screen.getByLabelText('Test Component')).toHaveDescription('');
+    });
+  });
+
+  describe('readOnly API', () => {
```
juanca, line 175: We need a test case for `readOnly={false}`.

src/components/number/number.jsx
juanca: Yeah, that would be a good solution!

src/components/number/number.jsx
```diff
@@ -0,0 +1,130 @@
...
+  id: PropTypes.string.isRequired,
```
juanca, line 103: We are missing spec coverage here.

src/components/number/number.jsx
```diff
@@ -0,0 +1,130 @@
...
+  label: PropTypes.string.isRequired,
+  name: PropTypes.string,
+  onBlur: PropTypes.func,
```
juanca, line 106: We are missing spec coverage here

src/components/number/number.jsx
```diff
@@ -0,0 +1,130 @@
...
+  onChange: PropTypes.func,
```
juanca, line 107: We are missing spec coverage here

src/components/number/number.md
```diff
@@ -0,0 +1,39 @@
+```js
```
juanca, line 1: I'm trying out a new pattern to make our examples a bit more consistent.

Can you follow this guide https://github.com/mavenlink/design-system/blob/master/docs/documentation.md ?

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
juanca: I think it's weird that the CustomFieldInputNumber used to emit an empty string and now it emits NaN. It's a breaking change tbh. 

And I think we are relying on it being an empty string in bigmaven for our custom fields app serialization -- I can't fully recall. We'll want to make sure the changes are compatible with usages in bigmaven  -- rogue is already using the custom field input number component as well.

juanca: Also, something to get your gears in the right direction,

> I think the value of the number custom field defaults to NaN when the value is empty, and that's because of its ref.

What has changed? Wasn't it a ref before as well? If it's "because it is a ref" then wouldn't it have been NaN in `master` too?

src/components/form-control/form-control.jsx
```diff
@@ -37,7 +37,12 @@ export default function FormControl(props) {
         {props.children}
       </div>
       {isInvalid(props.error, props.readOnly) && (
-        <span id={`${props.id}Hint`} className={styles['error-message']} aria-live="polite" >
+        <span
+          id={`${props.id}Hint`}
+          className={styles['error-message']}
+          aria-live="polite"
+          data-testid="error-message"
```
juanca, line 44: Basically, we will never want to rely on `testid`. We want to our tests to be user-centric: we want to test with queries about _visible_ or _audible_ information (unless I guess it's a ref?)

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -58,12 +58,12 @@ describe('CustomFieldInputNumber', () => {
       const { rerender } = render(<CustomFieldInputNumber {...requiredProps} errorText="" />);
       expect(screen.queryByText("Here's some help text!")).not.toBeInTheDocument();
       rerender(<CustomFieldInputNumber {...requiredProps} errorText="Here's some help text!" />);
-      expect(screen.queryByText("Here's some help text!")).toBeInTheDocument();
+      expect(screen.getByTestId('error-message')).toBeInTheDocument();
```
juanca, line 61: Follow the same patterns as the other tests you wrote in this PR: use `expect(...).toHaveDescription(...)`

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -58,12 +58,12 @@ describe('CustomFieldInputNumber', () => {
...
     });
 
     it('shows the provided errorText when true', () => {
       render(<CustomFieldInputNumber {...requiredProps} errorText="Here's some help text!" />);
-      expect(screen.getByText("Here's some help text!")).toBeInTheDocument();
+      expect(screen.getByTestId('error-message')).toBeInTheDocument();
```
juanca, line 66: Follow the same patterns as the other tests you wrote in this PR: use `expect(...).toHaveDescription(...)`

src/components/number/__snapshots__/number.test.jsx.snap
```diff
@@ -0,0 +1,42 @@
...
+        />
+      </div>
+    </div>
+  </div>
+</body>
+`;
+
+exports[`Number has defaults 2`] = `
+Object {
+  "dirty": true,
```
juanca, line 36: Same as the other snapshot. The component shouldn't be dirty on first render?

src/components/number/number.css
juanca: No, just compose from it.

src/components/number/number.css
juanca: Why is that? Do we want to be using the correct typography variables?

juanca (changes requested):
Please see requested revisions and questions. Apparently, I need to leave a message here even though I made a bunch of comments in the PR. /shrug

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
```diff
@@ -45,6 +39,6 @@ Object {
   "id": "test-input",
   "name": "field-id",
   "validity": ValidityState {},
-  "value": "",
+  "value": "undefined",
```
juanca, line 42: Because we're changing the API of an existing component, this is a "breaking change". We will _need_ a CI of bigmaven with these changes. 

Please update the PR upkeep list in the PR description.

src/components/custom-field-input-number/custom-field-input-number.jsx
```diff
@@ -35,28 +24,24 @@ const CustomFieldInputNumber = forwardRef(function CustomFieldInputNumber(props,
...
-      inputRef={inputRef}
       label={props.label}
       max={apiLimits.max}
       min={apiLimits.min}
       name={props.name}
       onBlur={props.onBlur}
-      onChange={onChange}
       placeholder={props.placeholder}
-      readOnly={props.readOnly}
+      readOnly={props.readOnly || props.disabled}
```
juanca, line 43: 👌 This is a good way to keep backwards compatibility _but_ allowing ourselves to keep iterating on our components.

src/components/number/number.css
juanca: @cincospenguinos  ?

src/components/number/number.css
juanca: @cincospenguinos ?

src/components/number/number.test.jsx
```diff
@@ -0,0 +1,243 @@
...
+  };
+
+  it('has defaults', () => {
+    const ref = createRef();
+    render(<Number {...requiredProps} ref={ref} />);
+    expect(document.body).toMatchSnapshot();
+    expect(ref.current).toMatchSnapshot();
+  });
+
+  describe('disabled API', () => {
```
juanca, line 24: Are we removing this?

src/components/number/number.jsx
```diff
@@ -0,0 +1,134 @@
...
+      id={props.id}
+      label={props.label}
+      readOnly={props.readOnly}
+      required={props.required}
+    >
+      <input
+        aria-describedby={`${props.id}Hint`}
+        className={getClassName(props.className, validationMessage)}
+        defaultValue={props.value}
+        disabled={props.disabled}
```
juanca, line 79: Why are we keeping this around?

src/components/custom-field-input-number/__snapshots__/custom-field-input-number.test.jsx.snap
cincospenguinos: This is outdated--it now defaults to undefined instead.

src/components/number/number.css
cincospenguinos: I'm not entirely sure. That's something we'll need to explore

src/components/number/number.css
```diff
@@ -0,0 +1,15 @@
+@import '../../styles/colors-v2.css';
```
juanca, line 1: We won't need these imports if we aren't using the css variables in the file.

juanca (approved):
That's a good lookin' component, if I ever seen one.

PR #243, opened 2020-11-13
Deprecate `react-test-renderer` in test patterns

juanca:
## Motivation

We want to stop using `react-test-renderer` as our snapshot generator. Instead, we should always use the `render` method from `@testing-library/react`. This ensures less overhead and actually tests any side effects on the DOM. 

## Acceptance Criteria

- Tests are still green

## PR upkeep checklist

- [ ] Change log entry N/A
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/deprecate-react-test-renderer/


cincospenguinos (approved):
LGTM homie

PR #244, opened 2020-11-13
Get the debounced submit to just be a regular submit

cincospenguinos:
[So listen up, here's the story](https://www.pivotaltracker.com/story/show/175602116)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-moving-steady-now-steady-now-steady-now-175602116/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/form/form.jsx
+++ b/src/components/form/form.jsx
@@ -1,6 +1,5 @@
 import PropTypes from 'prop-types';
 import React, {
-  useCallback,
   useEffect,
   useImperativeHandle,
   useRef,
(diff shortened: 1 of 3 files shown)


src/components/form/form.jsx
```diff
@@ -17,7 +16,7 @@ const Form = React.forwardRef((props, forwardedRef) => {
   const [dirty, setDirty] = useState(false);
   const [valid, setValid] = useState(true);
 
-  const onDebouncedSubmit = useCallback(debounce(() => {
+  const onRegularSubmit = useCallback(() => {
```
juanca, line 19: We can remove the `useCallback` since it's not doing anything useful in this usage.

bennapp (approved):
LGTM

juanca (approved):
I like it. I think we commented out a test for this. Can you update the test suite for coverage? We don't need to have a test about "every 300ms" but it would be nice to have a test for "every change".

PR #248, opened 2020-11-25
Add Popover component

Bloodyaugust:
## Motivation
We need a popover component for the form builder in order for a user to configure an individual component in the canvas. This component should be in the design system so that other teams can utilize it as well.

## Acceptance Criteria
SEE DRD:
https://docs.google.com/document/d/1JpQFBPt0Z6XF8xjjHco5qH8ubb3t2A4H_kwT_oEQE5I/edit?usp=sharing

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-all-about-the-popover-175407314/#/Components/Popover
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/popover/__snapshots__/popover.test.jsx.snap
+++ b/src/components/popover/__snapshots__/popover.test.jsx.snap
@@ -5,3 +5,9 @@ exports[`Popover has defaults 1`] = `
   <div />
 </body>
 `;
+
+exports[`Popover has defaults 2`] = `
+Object {
+  "open": false,
+}
+`;
--- a/src/components/popover/__snapshots__/popover.test.jsx.snap
+++ b/src/components/popover/__snapshots__/popover.test.jsx.snap
@@ -7,6 +7,48 @@ exports[`Popover has defaults 1`] = `
 `;
 
 exports[`Popover has defaults 2`] = `
+<body>
+  <div />
+  <div>
+    <section
+      aria-labelledby="popover-heading"
+      class="container"
+      role="dialog"
+    >
+      <div
+        role="presentation"
+      >
+        <div
+          class="heading-container"
+          id="popover-heading"
+        >
+          <h1
+            class="heading"
+          >
+            Defaults
+          </h1>
+          <svg
+            class="close-button"
+            height="8"
+            role="button"
+            tabindex="0"
+            width="8"
+          >
+            <title>
+              Close popover
+            </title>
+            <use
+              xlink:href="#clear.svg"
+            />
+          </svg>
+        </div>
+      </div>
+    </section>
+  </div>
+</body>
+`;
+
+exports[`Popover has defaults 3`] = `
 Object {
   "open": false,
 }
--- a/src/components/popover/__snapshots__/popover.test.jsx.snap
+++ b/src/components/popover/__snapshots__/popover.test.jsx.snap
@@ -1,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Popover has defaults 1`] = `
+<body>
+  <div />
+</body>
+`;
--- a/src/components/popover/popover.css
+++ b/src/components/popover/popover.css
@@ -16,7 +16,7 @@
   border: 1px solid var(--mds-grey-12);
   box-shadow: var(--spacing-small) var(--spacing-small) var(--spacing-small) var(--mds-grey-8);
   padding: var(--spacing-large);
-  position: fixed;
+  position: absolute;
   z-index: 10;
 }
 
--- a/src/components/popover/popover.css
+++ b/src/components/popover/popover.css
@@ -1,0 +1,32 @@
+@import '../../styles/colors-v2.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography-v2.css';
+
+.close-button {
+  composes: button from '../icon-button/icon-button.css';
+  height: 12px;
+  margin-left: var(--spacing-large);
+  padding: var(--spacing-small);
+  padding-top: 5px;
+  width: 12px;
+}
+
+.container {
+  background-color: var(--white);
+  border: 1px solid var(--mds-grey-12);
+  box-shadow: var(--spacing-small) var(--spacing-small) var(--spacing-small) var(--mds-grey-8);
+  padding: var(--spacing-large);
+  position: fixed;
+  z-index: 10;
+}
+
+.heading {
+  font: var(--mds-type-subhead-1);
+  margin-bottom: var(--spacing-large);
+  margin-top: 0;
+}
+
+.heading-container {
+  display: flex;
+  justify-content: space-between;
+}
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -18,7 +18,7 @@ const Popover = forwardRef(function Popover(props, ref) {
   const selfRef = ref || backupRef;
 
   const onFocusIn = (event) => {
-    if (open && event.target instanceof HTMLElement && !sectionRef.current.contains(event.target)) {
+    if (open && event.target instanceof Node && !sectionRef.current.contains(event.target)) {
       setOpen(false);
     }
   };
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -14,7 +14,14 @@ const Popover = forwardRef(function Popover(props, ref) {
   const [open, setOpen] = useState(false);
   const closeIconRef = useRef();
   const backupRef = useRef();
+  const sectionRef = useRef();
   const selfRef = ref || backupRef;
+
+  const onFocusIn = (event) => {
+    if (open && event.target instanceof HTMLElement && !sectionRef.current.contains(event.target)) {
+      setOpen(false);
+    }
+  };
 
   const onWindowClick = () => {
     if (open) {
@@ -24,6 +31,7 @@ const Popover = forwardRef(function Popover(props, ref) {
 
   useEffect(() => {
     window.addEventListener('click', onWindowClick);
+    window.addEventListener('focusin', onFocusIn);
 
     if (closeIconRef.current && open) {
       closeIconRef.current.focus();
@@ -35,6 +43,7 @@ const Popover = forwardRef(function Popover(props, ref) {
 
     return () => {
       window.removeEventListener('click', onWindowClick);
+      window.removeEventListener('focusin', onFocusIn);
     };
   }, [open]);
 
@@ -57,6 +66,7 @@ const Popover = forwardRef(function Popover(props, ref) {
       aria-labelledby="popover-heading"
       className={styles.container}
       onClick={(event) => { event.stopPropagation(); }}
+      ref={sectionRef}
       role="dialog"
       style={{ left: `${props.left}px`, top: `${props.top}px` }}
     >
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -1,0 +1,94 @@
+import PropTypes from 'prop-types';
+import React, {
+  forwardRef,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
+import IconButton from '../icon-button/icon-button.jsx';
+import iconClear from '../../svgs/clear.svg';
+import styles from './popover.css';
+
+const Popover = forwardRef(function Popover(props, ref) {
+  const [open, setOpen] = useState(false);
+  const closeIconRef = useRef();
+  const backupRef = useRef();
+  const selfRef = ref || backupRef;
+
+  const onWindowClick = () => {
+    if (open) {
+      setOpen(false);
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('click', onWindowClick);
+
+    if (closeIconRef.current && open) {
+      closeIconRef.current.focus();
+    }
+
+    if (!open) {
+      props.onClose();
+    }
+
+    return () => {
+      window.removeEventListener('click', onWindowClick);
+    };
+  }, [open]);
+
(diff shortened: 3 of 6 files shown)


src/components/popover/popover.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+const popoverRef = React.createRef();
+
+const onButtonClicked = () => {
+  popoverRef.current.open = !popoverRef.current.open;
+}
+
+<React.Fragment>
+  <button onClick={onButtonClicked}>Toggle Popover</button>
```
bennapp, line 15: I couldn't get the first example to pop-up on staging.

bennapp (approved):
LGTM!

src/components/popover/popover.md
Bloodyaugust: @bennapp It was probably just lower on the screen than you expected, will you look again?

From the DRD:
> If user scrolls page behind popover, the popover should remain fixed in place and not move with the page.

bennapp: Still can't find it 🤷  👻  I'm looking here: https://mavenlink.github.io/design-system/webinf-all-about-the-popover-175407314/#/Components/Popover

at the first 'Toggle Popover'

Bloodyaugust: Strange, I'm seeing it! What browser are you using to test? Maybe a hard reload?

src/components/popover/popover.md
```diff
@@ -0,0 +1,42 @@
+The `popover` component provides functionality somewhere between a modal and a toggletip. It doesn't interrupt flow like a modal would, but provides more than the purely informational function of a toggletip. It is a way to allow for greater branching functionality while staying in-flow.
```
juanca, line 1: We will want to reference the component as `Popover` since that is how it is used. Please update the rest of the references in this file.

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,94 @@
...
+    if (!open) {
+      props.onClose();
+    }
+
+    return () => {
+      window.removeEventListener('click', onWindowClick);
+    };
+  }, [open]);
+
+  useImperativeHandle(selfRef, () => ({
```
juanca, line 41: We will want a test for the default ref state. https://github.com/mavenlink/design-system/blame/master/docs/testing.md#L36

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const selfRef = ref || backupRef;
+
+  const onWindowClick = () => {
+    if (open) {
+      setOpen(false);
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('click', onWindowClick);
```
juanca, line 26: Should the Popover close when anything _else_ on the page is focused? 

I am thinking we want feature parity between mouse users and keyboard users.

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }));
+
+  if (!open) {
+    return null;
+  }
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
+    <section
+      aria-labelledby="popover-heading"
```
juanca, line 57: If a Popover can be left opened, this ID will be duplicated on the DOM. 

We will want to ensure a unique ID is generated on every instance of a Popover. 

Historically, we've been providing functionality by adding an `id` prop and letting the user provide us with a unique string -- and then we either use it or add to it.

e.g. 

```suggestion
      aria-labelledby={`${props.id}-popover-heading`}
```

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,94 @@
...
+      setOpen(value);
+    },
+  }));
+
+  if (!open) {
+    return null;
+  }
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
```
juanca, line 55: We don't need to bypass the rule if we provide a list of roles for the node, somewhere in this doc https://www.w3.org/TR/2014/REC-wai-aria-20140320/host_languages#host_general_role

e.g. `role="dialog presentation"`

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,94 @@
...
+    <section
+      aria-labelledby="popover-heading"
+      className={styles.container}
+      onClick={(event) => { event.stopPropagation(); }}
+      role="dialog"
+      style={{ left: `${props.left}px`, top: `${props.top}px` }}
```
juanca, line 61: Just food for thought: 

Since we want a separation of CSS vs JS, we've implemented a "bridge" by setting CSS variables. 

```suggestion
      style={{ '--left': props.left, '--top': props.top }}
```

And the CSS file will be self-contained:

```css
.foo {
  left: var(--left);
  top: var(--top);
}
```

src/components/popover/popover.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    userEvent.click(screen.getByText('Toggle Popover'));
+    expect(screen.queryByText('Another title')).toEqual(null);
+  });
+
+  it('can be closed by clicking elsewhere in the window', () => {
+    render(<PopoverWithToggle title="Another title" />);
+
+    userEvent.click(screen.getByText('Toggle Popover'));
+    expect(screen.getByText('Another title').tagName).toEqual('H1');
```
juanca, line 60: I think these test are trying to assert that the popover is on the page at this point? We want to avoid saying that the title is an H1 in case we change it in the future.

```suggestion
    expect(screen.getByText('Another title')).toBeInTheDocument()
```

src/components/popover/popover.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    userEvent.click(screen.getByText('Some text'));
+    expect(screen.queryByText('Another title')).toEqual(null);
+  });
+
+  it('opens with focus on the close iconbutton', () => {
+    render(<PopoverWithToggle title="Another title" />);
+
+    userEvent.click(screen.getByText('Toggle Popover'));
+    expect(document.activeElement.tagName).toEqual('svg');
```
juanca, line 70: You might be able to do a bit better by saying

```suggestion
    expect(document.activeElement).toEqual(screen.getByLabelText('Close button'));
```

since the close button should have a reasonable label already.

src/components/popover/popover.test.jsx
juanca: I do like that you skip this kind of assertion in the following tests -- we're basically saying that we _know_ the popover is on the page after clicking "Toggle Popover"

Maybe we want to do this same here?

src/components/popover/popover.css
```diff
@@ -0,0 +1,32 @@
...
+  padding-top: 5px;
+  width: 12px;
+}
+
+.container {
+  background-color: var(--white);
+  border: 1px solid var(--mds-grey-12);
+  box-shadow: var(--spacing-small) var(--spacing-small) var(--spacing-small) var(--mds-grey-8);
+  padding: var(--spacing-large);
+  position: fixed;
```
juanca, line 19: Just to track this: we decided against `fixed` in favor of `absolute` so the popover element scrolls with the page.

src/components/popover/__snapshots__/popover.test.jsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Popover has defaults 1`] = `
```
juanca, line 3: I think it might make sense to take a snapshot when it is opened as well. I don't wanna get too crazy with snapshotting but it sounds reasonable for something this the core functionality and a boolean.

juanca (changes requested):
Ping me after the transition to abso and stuff. Looking good so far.

src/components/popover/popover.jsx
Bloodyaugust: DRD is going to be amended to show that only one `Popover` can be open at a time.

src/components/popover/popover.jsx
Bloodyaugust: To satisfy the semantics and the linter, I created a new inner `div` with `role="presentation"`, and the event handler is now there instead.

See also: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md#case-this-element-is-catching-bubbled-events-from-elements-that-it-contains

juanca: Does adding two roles instead of an additional div make more sense?

Bloodyaugust: Unfortunately, two roles doesn't seem to satisfy the linter, and by my reading, isn't what the spec outlines either. 😬

src/components/popover/popover.jsx
Bloodyaugust: I'll keep this in mind for next time, but with the move to `absolute`, this props API doesn't exist anymore. 😄

src/components/popover/__snapshots__/popover.test.jsx.snap
Bloodyaugust: Done and done, with a new props API `startOpen`.

Bloodyaugust:
Focus events always closing Popover captured as bug here: https://www.pivotaltracker.com/story/show/175981707

src/components/popover/popover.css
```diff
@@ -0,0 +1,32 @@
+@import '../../styles/colors-v2.css';
+@import '../../styles/spacing.css';
+@import '../../styles/typography-v2.css';
+
+.close-button {
+  composes: button from '../icon-button/icon-button.css';
+  height: 12px;
+  margin-left: var(--spacing-large);
+  padding: var(--spacing-small);
+  padding-top: 5px;
```
juanca, line 10: I'm not sure we need the pixel correction here. But I only briefly looked at it when toggling this style.

src/components/popover/popover.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+const Popover = forwardRef(function Popover(props, ref) {
+  const [open, setOpen] = useState(props.startOpen);
+  const closeIconRef = useRef();
+  const backupRef = useRef();
+  const sectionRef = useRef();
+  const selfRef = ref || backupRef;
+
+  const onFocusIn = (event) => {
+    if (open && event.target instanceof HTMLElement && !sectionRef.current.contains(event.target)) {
```
juanca, line 21: This is a first I'm seeing a type check on event.target. Is it possible to get something without `contains`?

The class inheritance looks like:

EventTarget -> Node -> Element -> HTMLElement

And the `contains` method is defined on the Node class. 

> Element, Document, and Window are the most common event targets, but other objects can be event targets, too. For example XMLHttpRequest, AudioNode, AudioContext, and others.

Did you see something without a `contains` while developing?

src/components/popover/popover.test.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+    render(<Popover ref={popoverRef} title="Defaults with ref" />);
+    expect(popoverRef.current).toMatchSnapshot();
+  });
+
+  it('has a title', () => {
+    render(<PopoverWithToggle title="Another title" />);
+
+    userEvent.click(screen.getByText('Open Popover'));
+    expect(screen.getByText('Another title').tagName).toEqual('H1');
```
juanca, line 37: It's probably more user-centric to make an assertion on the label.

Something like: 

```suggestion
    expect(screen.getByRole('dialog', { name: 'Another title' })).toBeInTheDocument();
```

src/components/popover/popover.test.jsx
```diff
@@ -0,0 +1,106 @@
...
+    userEvent.click(screen.getByText('Some text'));
+    expect(screen.queryByText('Another title')).not.toBeInTheDocument();
+  });
+
+  it('can be closed by focusing outside', () => {
+    render(<PopoverWithToggle title="Another title" />);
+
+    userEvent.click(screen.getByText('Open Popover'));
+
+    fireEvent.focusIn(screen.getByText('Close Popover'));
```
juanca, line 72: It's more user-centric to use user actions to perform the focus switching. Something like:

```suggestion
    userEvent.tab();
```

src/components/popover/popover.jsx
juanca: Oh, bummer. It's probably not popular to use a list of roles. 

Or I am just misinterpreting:

> The attribute value MUST allow a token list as the value;

juanca (approved):

src/components/popover/popover.jsx
Bloodyaugust: @juanca while testing, actually! Sometimes in testing, there seems to be `focus` events happening where `event.target` is not `null || undefined`, but is also not `Node`. 😬

src/components/popover/popover.css
Bloodyaugust: It's pretty subtle, but it definitely isn't center-aligned with the text without it. 😭

src/components/popover/popover.test.jsx
Bloodyaugust: Interesting note: it seems that in test land, the whole `window` object can be focused, so we actually have to tab twice to get back to the button. 😅

Bloodyaugust: This may also explain the odd behavior with `focus` events occurring with an `event.target` that doesn't inherit from `Node`.

src/components/popover/popover.jsx
Bloodyaugust: I must've missed _list_ there, cuz yeah, that does seem like it should work then. 😬 Maybe a mismatch between spec and browser implementation? Or at least between spec and linter...

PR #269, opened 2021-02-03
MDS CustomFieldSingleChoice fetches its own choices

Bloodyaugust:
## Motivation
Use MSW for tests and styleguide site. We will want to figure out how to load MSW on styleguidist for our repo and bigmaven.

We will want to do this nicely because future components will also have a "server" aspect: e.g. all autocompleters, etc.

Make the component:

1. Fetch its own list of choices
1. Fetch the label of its select choice (when given a new value)
1. Make the value API symmetrical: props <=> ref <=> api. This makes it super easy to hook up. (See https://github.com/mavenlink/mavenlink/pull/23247/files#r566537207)

## Acceptance Criteria
- On the documentation page, each example has its own list of choices (besides the empty example). (Note: this is the case for MDS site and internal MDS site.)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-custom-field-single-fetches-integrate-msw-176709298/
- [x] Green Bigmaven CI check


src/mocks/jest-server.js
```diff
@@ -0,0 +1,6 @@
+import { setupServer } from 'msw/node';
+import handlers from './handlers.js';
+
+const jestServer = setupServer(...handlers);
```
juanca, line 4: Let's set up the server to fail tests if there is any unhandled server request -- we don't actually want to ping `api.com`

Bloodyaugust: @juanca I'm confused, what do you mean? Are you saying we don't want any components to actually hit those endpoints in production by accident?

If so, I'm not sure how we'd test for that here, since in test land we _want_ to hit this endpoint. It's like... the opposite of a circular dependency? 🤣

styleguide/assets/mockServiceWorker.js
```diff
@@ -0,0 +1,283 @@
+/**
```
Bloodyaugust, line 1: We should try to make `styleguidist` serve this directly from `node_modules/msw/lib/iife`

jest.config.setup.files.after.env.js
```diff
@@ -1,3 +1,13 @@
+import jestServer from './src/mocks/jest-server.js';
+
+/* eslint-disable no-undef */
```
juanca, line 3: There is a way to fix this: you can make eslint understand that it has Jest loaded when parsing this file. 

But this is fine too.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -7,30 +7,35 @@ import React, {
...
   const listOptions = () => {
-    const choices = {};
-    props.choices.forEach((item, index) => { choices[item.id] = { ...item, index }; });
+    const mungedChoices = {};
+    choices.forEach((item, index) => { mungedChoices[item.id] = { ...item, index }; });
 
-    return props.choices
-      .map(item => choices[item.id])
+    return choices
+      .map(item => mungedChoices[item.id])
```
juanca, line 38: What is going on with this `forEach` -> `map`?

We use the `forEach` to create a map: `id` -> `{ item, index: i }`
Then use the map to set it is as `{ item, index: i }`.

Can we just not do the `forEach`?

```suggestion
      .map((choice, index) => ({ ...choice, index }))
```

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -68,23 +73,54 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
     props.onChange({ target: selfRef.current });
   }, [value]);
 
+  useEffect(() => {
+    const getChoices = async () => {
+      await execute(`${API_ROOT}/custom_field_choices?for_custom_fields=${props.id}`)
+        .then(({ json, mounted }) => {
+          if (mounted) {
+            const mungedChoices = json.results.map((result) => {
+              return {
```
juanca, line 82:
```suggestion
              return json[result.key][result.id];
```

I think it is preferable to keep using the API object as a whole instead of slicing what we need. What do you think?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -68,23 +73,54 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
+                id: result.id,
+                label: json[result.key][result.id].label,
+              };
+            });
+
+            setChoices(mungedChoices);
+            setListOptionRefs(mungedChoices.map(() => createRef()));
```
juanca, line 89: This is interesting. What do you think about simplifying our state management by dropping `setListOptionRefs` as part of state?

It is a calculated piece of state that will never be anything that is not derived from the length of loaded choices.

Instead, do this piece of logic at the start of the function:

(I'll add the suggestion at the relevant line)

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -7,30 +7,35 @@ import React, {
...
+import useFetch from '@bloodyaugust/use-fetch';
 import Select from '../select/select.jsx';
 import ListOption from '../list-option/list-option.jsx';
+import mockConstants from '../../mocks/mock-constants.js';
+
+const { API_ROOT } = mockConstants;
 
 const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleChoice(props, ref) {
+  const [choices, setChoices] = useState([]);
+  const [listOptionRefs, setListOptionRefs] = useState([]);
```
juanca, line 19: This seem unnecessary. Try this:

```suggestion
  const listOptionRefs = choices.map(() => createRef());
```

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -94,8 +130,8 @@ const ChoiceType = PropTypes.shape({
 });
 
 CustomFieldInputSingleChoice.propTypes = {
+  /** Needs to be the Custom Field ID from the database for self-loading of choices */
```
juanca, line 133: I just realized this might not be the case in our builder.

We have two instance of a particular single choice:

1. Canvas
2. Popover default value

Both of these are single choice fields that need to load their own data.

Unfortunately, we use `id` in our React components as a HTML `id` attribute -- where we consider it to be _unique_ on the DOM. 

We will want to use a new prop for the custom field id. I'm okay with `fieldId` or `customFieldId`. You can come up with something too.

src/mocks/jest-server.test.js
```diff
@@ -0,0 +1,29 @@
...
+    }));
+  });
+
+  it('"serves" on http://localhost', async () => {
+    const body = await (await fetch('http://localhost/hello')).text();
+
+    expect(body).toEqual('Hello from MSW!');
+  });
+
+  // This test exists to prove that MSW will fail tests that try to request against an unhandled endpoint. Un-"x"-it if you need to confirm the failure.
```
juanca, line 23: Can it be tested with `expect(() => { ... }).toThrow()`?

src/mocks/jest-server.js
```diff
@@ -0,0 +1,5 @@
+import { setupServer } from 'msw/node'; // eslint-disable-line import/extensions
```
juanca, line 1: Why can't extension?

juanca:
This site isn't loading with MSW:

image

juanca:
> Make the value API symmetrical: props <=> ref <=> api.

Do we want to make this change now or a different PR? I don't think the integration with bigmaven will work otherwise.

juanca:
A point of clarification:

> Fetch the label of its select choice (when given a new value)

I don't think this is the right wording. We already have a list of choices and an ID for the "selected choice". This is meant to say that it will use its loaded list of choices to figure out the label for its given value.

juanca (changes requested):
Let me know if you have any questions or how I can help.

1. PR revisions
2. Broken staging
3. Missing `value` API updates (I think by now, probably a new PR to get this in)

src/mocks/jest-server.test.js
Bloodyaugust: I've tried a few variations on this now, and it seems like the throw does not happen perhaps at the correct time, or maybe not specifically in the call stack of this function? Either way, it doesn't seem to be capturable in this manner. 😬

src/mocks/jest-server.js
Bloodyaugust: I'm not entirely sure why this is the case, but here's the error stack I get:
```
    Cannot find module 'msw/node.js' from 'src/mocks/jest-server.js'

    Require stack:
      src/mocks/jest-server.js
      jest.config.setup.files.after.env.js

    > 1 | import { setupServer } from 'msw/node.js';
```

src/mocks/jest-server.test.js
Bloodyaugust: Nevermind, @bennapp has galaxy brain and fixed it!

src/mocks/jest-server.js
Bloodyaugust: This is, I think, complaining because it _looks_ like a relative import from our own code because it has a `/` in the path. The linter doesn't complain about other module imports without extensions.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -7,39 +7,37 @@ import React, {
   useState,
 } from 'react';
 import PropTypes from 'prop-types';
+import useFetch from '@bloodyaugust/use-fetch';
 import Select from '../select/select.jsx';
 import ListOption from '../list-option/list-option.jsx';
+import mockConstants from '../../mocks/mock-constants.js';
+
+const { API_ROOT } = mockConstants;
+const defaultValue = [-1];
```
juanca, line 16: I'm see there is other code referencing `-1`. I feel like using the default value of `[]` should be better. What's the reasoning for using `[-1]`?

juanca: For posterity, this is used to capture the empty case. It is useful to have a fixed sized array for spreading into the `useEffect`'s dependencies parameter.

src/components/list-option/list-option.jsx
```diff
@@ -23,7 +23,7 @@ const ListOption = forwardRef(function ListOption(props, ref) {
   const [selected, setSelected] = useState(props.selected);
   const [visible, setVisible] = useState(true);
   const rootRef = useRef();
-  const selectedClassName = props.selected ? styles.selected : styles.option;
+  const selectedClassName = selected ? styles.selected : styles.option;
```
juanca, line 26: Nice fix.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -48,73 +46,98 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
-  id: PropTypes.number.isRequired,
-  label: PropTypes.string.isRequired,
-});
-
 CustomFieldInputSingleChoice.propTypes = {
   id: PropTypes.string.isRequired,
-  choices: PropTypes.arrayOf(ChoiceType),
   className: PropTypes.string,
+  /** Needs to be the Custom Field ID from the database for self-loading of choices */
+  customFieldID: PropTypes.number.isRequired,
```
juanca, line 123: Curious about treating ID as two words instead of one: `customFieldId` /shrug We just have to stay consistent moving forward.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -48,73 +46,98 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
...
+        name={props.name}
+        onChange={selectOnChangeHandler}
+        placeholder={props.placeholder}
+        readOnly={props.readOnly}
+        required={props.required}
+        value={value[0] !== -1 ? choices.find(choice => choice.id === value[0]) : undefined}
+      >
+        { listOptions() }
+      </Select>
+      {choices.length === 0 &&
```
juanca, line 112: I know we'll get this cleaned up in a future PR but this logic implies that an empty custom field choice will always render the "loading" text ;) I'm not even sure if the API is okay with a zero-choice custom field.

juanca (approved):

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
Bloodyaugust: Yeah, since `ID` is an abbreviation, I make the letters share case. I'm cool with either, but this is what I default to.

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
Bloodyaugust: Good point. We'd need a "request completed successfully" identifier for this, I think. Also, I'm pretty sure the API _does_ allow zero choice custom fields, because you create the field first, and then the choices (unless the UI is hiding this).

PR #270, opened 2021-02-05
Implement updatable value APIs on Select and SingleChoice components

juanca:
https://www.pivotaltracker.com/story/show/176491263

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/single-value-update/
- [ ] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/23412


juanca:
Uh-oh, I think our Select <=> SingleChoice architecture might have some asynchronous useEffect leakage (in a test env).

image

src/components/select/select.test.jsx
```diff
@@ -320,6 +320,12 @@ describe('src/components/select/select', () => {
```
Bloodyaugust, line 320: We should probably do the same refactor on the above `'provided value sets the corresponding list item as selected'` test that we did for the custom field single choice input test of the same name.

Bloodyaugust (approved):
LGTM! One potential test refactor.

juanca:
@Bloodyaugust I needed to do some more changes to fix a bug in our custom forms application: 

- When mounting the component, the component would always unselect its provided value because a useEffect always runs for a dependency before it is mounted. The fix was to check to see if the component is mounted.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Minor: Implement updatable `value` API on Select component
```
Bloodyaugust, line 10: This feels like a breaking (major) change, since it has potential to break established implementations.

src/hooks/use-validation.jsx
```diff
@@ -4,10 +4,10 @@ export default function useValidation(errorText, inputRef) {
   const [validationMessage, setValidationMessage] = useState('');
 
   useEffect(() => {
-    if (inputRef.current) {
-      inputRef.current.setCustomValidity(errorText);
-      setValidationMessage(errorText);
-    }
+    if (!inputRef.current) return;
```
Bloodyaugust, line 7: I thought you didn't like early returns behind if statements? 😜

Bloodyaugust (approved):
LGTM! Perhaps our changes for how value responds to prop changes should be considered major tho.

CHANGELOG.md
juanca: That's a good question. 🤔 

a. The developer never updates the `value` prop on re-render -- this is a minor
b. The developer updates the `value` prop on re-render due to internal factors (i.e. `onChange` callback) -> component sets value (theoretically, the value is the same thus the state update effects do not run) -- this is a minor
c. The developer updates the `value` prop on re-render due to external factors (e.g. server updates) -> component sets value (and re-render to match the new value) -> the `onChange` callback is ran -> user code runs which was previously not ran -- this does seem like a major change

This brings up a good question:

- Do we want to run an `onChange` callback when the props change? Theoretically, if the props are driving a change then it means the user already handle the effects of the change. Instead, I think our `onChange` shouldn't call `onChange` when it matches the prop. 🤔 

I'm try a commit, see how it plays out in bigmaven.

juanca: Done. It works pretty I think. It will build against bigmaven just in case.

PR #273, opened 2021-02-16
MDS CustomFieldInputNumber: value API is an <integer>

bennapp:
## Motivation
[MDS CustomFieldInputNumber: value API is an <integer>](https://www.pivotaltracker.com/story/show/176949762)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-number-is-number-176949762/
- [x] Green Bigmaven CI check (in progress)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Minor: `ref.value` for CustomFieldInputNumber now returns its type as Integer
 </details>
 
 ## v0.58.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
-  - Minor: `ref.value` for CustomFieldInputNumber now returns its type as Integer
 </details>
+
+## v0.59.0
+  - Major: `ref.value` for CustomFieldInputNumber now returns its type as Integer. To upgrade replace value with String(value)
 
 ## v0.58.0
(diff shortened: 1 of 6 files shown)


bennapp:
Testing change in bigmaven here: https://github.com/mavenlink/mavenlink/pull/23550

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - Minor: `ref.value` for CustomFieldInputNumber now returns its type as Integer
```
juanca, line 10: Theoretically, this is a breaking change so it would be a major. Usually, with breaking changes, it is recommended to add a quick note on how to upgrade.

In this case: Usages of `value` on the ref API can be interpolated in a string. i.e. `String(ref.current.value)`

src/components/custom-field-input-number/custom-field-input-number.test.jsx
```diff
@@ -157,9 +157,17 @@ describe('CustomFieldInputNumber', () => {
 
   describe('value API', () => {
     it('is valid on a positive integer', () => {
-      render(<CustomFieldInputNumber {...requiredProps} value={1} />);
+      const ref = createRef();
+      render(<CustomFieldInputNumber ref={ref} {...requiredProps} value={1} />);
```
juanca, line 161: This is an observation: the order of the supplied props matter!

In this case, it is usually best to spread the `requiredProps` before any test specific props. The test-specific props will clobber any props in `requiredProps`.

While this is technically not wrong, it could cause headaches down the road. (Same below)

juanca (approved):
Let's update the changelog before merging. We want to get into good habits for identifying major/minor/patches.

PR #274, opened 2021-02-18
`undefined` coerces to `null` to avoid browser warnings when clearing…

bennapp:
… Number component's input

## Motivation
The browser was warning when we were clearing the input. It does not like it when we set `current.value` to `undefined`

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
   - Append new items to make git merging easier.
 </details>
 
-## v0.60.0
-  - Minor: `undefined` coerces to `null` to avoid browser warnings when clearing Number component's input 
+## v0.59.1
+  - Patch: `undefined` coerces to `null` to avoid browser warnings when clearing Number component's input 
 
 ## v0.59.0
   - Major: `ref.value` for CustomFieldInputNumber now returns its type as Integer. To upgrade replace value with String(value)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
   - Append new items to make git merging easier.
 </details>
 
+## v0.60.0
+  - Minor: `undefined` coerces to `null` to avoid browser warnings when clearing Number component's input 
+
 ## v0.59.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
   - Append new items to make git merging easier.
 </details>
 
+## v0.60.0
+  - Minor: `undefined` coerces to `null` to avoid browser warnings when clearing Number component's input
```
juanca, line 13: 😓  Wouldn't this just be a fix since it is not a feature enhancement?

src/components/number/number.jsx
```diff
@@ -28,7 +28,11 @@ const Number = React.forwardRef((props, ref) => {
     // The MDS Number is using an uncontrolled `<input>`.
     // In order to set a new provided value prop, we
     // set the internal state of the `<input>`.
-    inputRef.current.value = props.value;
+    if (props.value === undefined) {
+      inputRef.current.value = null;
```
juanca, line 32: Do we have the same warnings in JSDOM?

Can we try spying on `console.warn` and see if it is invoked? I just want to make sure we don't re-introduce this bug again in the future.

If not, it's cool. I do want to give it a try though.

juanca (approved):
There might be a possibility for some tests. 

Also, let's update the PR description with the relevant bits. I'd like to poke at the MDS site before merge.

src/components/number/number.jsx
bennapp: I cannot get a failing test to work on master.
```
    it('does not warn when clearing the input', () => {
      const warnSpy = jest.spyOn(window.console, 'warn');
      render(<Number {...requiredProps} />);
      userEvent.type(screen.getByLabelText('Test Component'), '1');
      fireEvent.keyDown(screen.getByLabelText('Test Component'), { key: 'Backspace' });
      expect(warnSpy.mock.calls.length).toEqual(0);
    });
```

bennapp: I am also not sure there is a *ton* of value having this regression test in the test suite 🤷

PR #282, opened 2021-03-23
Fix first render to not show errors for select.jsx when its required

bennapp:
## Motivation
[The "Where would you like to embed the form?" field is loading on the page in an error state before the user interacts with the field](https://www.pivotaltracker.com/story/show/177444041)

## Acceptance Criteria
When required, the select should not show error text on initial render

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Deployment URL: https://mavenlink.github.io/design-system/webinf-fix-single-select-error/
- [ ] Green Bigmaven CI check


src/components/select/select.jsx
```diff
@@ -121,13 +123,23 @@ const Select = forwardRef(function Select(props, ref) {
...
+      validate();
+    }
 
     if (props.value === value) return;
 
     props.onChange({ target: selfRef.current });
   }, [value]);
 
+  function handleBlur(event) {
+    if (!(listBoxRef.current && listBoxRef.current.contains(event.relatedTarget))) {
```
juanca, line 136: `FocusEvent.relatedTarget` is experimental. But I think it's okay to keep this implementation for now. I think @Bloodyaugust found a React-only solution but it requires upgrading our React version or something. If we do find problems with `relatedTarget` then we can just focus big-time on a React specific solution.

image

src/components/select/select.test.jsx
```diff
@@ -166,6 +166,13 @@ describe('src/components/select/select', () => {
       render(<Select {...requiredProps} errorText="not valid">{baseListOptionElements}</Select>);
       expect(screen.getByLabelText('Test label')).not.toBeValid();
     });
+
+    it('does not show errorText on first render until blurred', () => {
+      render(<Select {...requiredProps} required >{baseListOptionElements}</Select>);
```
juanca, line 171:
```suggestion
      render(<Select {...requiredProps} required>{baseListOptionElements}</Select>);
```

What is this, dangling space town?

src/components/select/select.test.jsx
```diff
@@ -166,6 +166,13 @@ describe('src/components/select/select', () => {
...
+      expect(screen.queryByText('Constraints not satisfied')).toBeNull();
+      fireEvent.blur(screen.getByLabelText('Test label'));
```
juanca, line 173: We prefer `userEvent` testing. Something like:

```suggestion
      userEvent.click(screen.getByLabelText('Test label'));
      userEvent.tab();
```

juanca: Or click the body or something.

juanca (approved):
Minor testing pattern improvement: `userEvent` > `fireEvent` as a rule of thumb

PR #283, opened 2021-03-23
fix(inputs): ensure inputs use MDS fonts with a height of 32px

juanca:
https://www.pivotaltracker.com/story/show/177467078

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/bugfix-mds-inputs-font/
- [x] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/24038

### DIFF
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.css
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.css
@@ -10,7 +10,7 @@
   box-sizing: border-box;
   color: var(--mds-grey-87);
   font: var(--mds-type-content);
-  height: 32px;
+  height: var(--spacing-x-large);
   padding-right: calc((var(--numIcon, 0) * var(--iconSizeMedium)) + var(--spacing-medium)); /* stylelint-disable-line css-modules/css-variables */
   padding-left: var(--spacing-medium);
   width: 100%;
--- a/src/components/__internal__/abstract-custom-field/abstract-custom-field.css
+++ b/src/components/__internal__/abstract-custom-field/abstract-custom-field.css
@@ -2,16 +2,16 @@
 @import '../../../styles/colors-v2.css';
 @import '../../../styles/spacing.css';
 @import '../../../styles/typography.css';
+@import '../../../styles/typography-v2.css';
 
 .input-styles {
   border: 1px solid var(--mds-grey-12);
   border-radius: var(--spacing-x-small);
   box-sizing: border-box;
   color: var(--mds-grey-87);
-  font-size: var(--mavenlink-type-base-size); /* stylelint-disable-line mds/typography */
-  padding-top: 7px;
+  font: var(--mds-type-content);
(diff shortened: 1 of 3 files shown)


src/components/__internal__/abstract-custom-field/abstract-custom-field.css
```diff
@@ -2,16 +2,16 @@
...
 
 .input-styles {
   border: 1px solid var(--mds-grey-12);
   border-radius: var(--spacing-x-small);
   box-sizing: border-box;
   color: var(--mds-grey-87);
-  font-size: var(--mavenlink-type-base-size); /* stylelint-disable-line mds/typography */
-  padding-top: 7px;
+  font: var(--mds-type-content);
+  height: 32px;
```
Bloodyaugust, line 13: Should we codify this as a variable to make it less of a magic number? 🤔

src/components/input/input.css
```diff
@@ -1,16 +1,15 @@
...
   border: 1px solid var(--mds-grey-12);
   border-radius: 3px;
   box-sizing: border-box;
-  font-family: var(--mavenlink-type-font-family); /* stylelint-disable-line mds/typography */
-  font-size: var(--mavenlink-type-base-size); /* stylelint-disable-line mds/typography */
+  font: var(--mds-type-content);
   height: 32px;
   margin: 0;
-  padding: 6px var(--spacing-x-large) 7px var(--spacing-medium);
+  padding: 0 var(--spacing-x-large) 0 var(--spacing-medium);
```
Bloodyaugust, line 12: I wonder why we had such specific top and bottom padding...

Bloodyaugust (approved):
LGTM! One opportunity for a small refactor.

src/components/__internal__/abstract-custom-field/abstract-custom-field.css
juanca: Oh, yeah. That is x-large already. Hah.

src/components/input/input.css
juanca: We basically don't want to specify top or bottom. We want the text to just be vertically aligned (defaults on inputs) and take up as much space as specified in the font.

juanca: This is just the ramification of choosing the short-hand property instead of individual properties like `padding-left` and `padding-right`.

Bloodyaugust: Sorry, I was wondering out loud why the previous values were so specifically set. 😄

juanca: Ah, yeah. That.

I... don't recall. Part of me wants to say that it was originally intended for the inputs to be 32px high but we didn't want to set height? Like we wanted the component to "organically" take up its own space from its sub-dimensions. The 6 vs 7 difference I think was a trick to get the font to look more centered.

PR #287, opened 2021-03-31
fix(custom-field-input-date): does not offset value when a date with …

bennapp:
## Motivation
https://www.pivotaltracker.com/story/show/177518307

## Acceptance Criteria
Entering a date like: `01-01-2020 00:00:00` should not show a value offset by one day.

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/custom-field-input-date/custom-field-input-date.test.jsx
+++ b/src/components/custom-field-input-date/custom-field-input-date.test.jsx
@@ -84,6 +84,11 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
       expect(getByLabelText('Field Date')).toHaveValue('Jul 18, 2016');
     });
 
+    it('accepts a string in format mm-dd-yyyy HH:MM:SS', () => {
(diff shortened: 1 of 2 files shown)


src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -84,6 +84,11 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
       expect(getByLabelText('Field Date')).toHaveValue('Jul 18, 2016');
     });
 
+    it('accepts a string in format mm-dd-yyyy HH:MM:SS', () => {
```
juanca, line 87: I was under the impression that this component API is not desired since the custom field date API only supports date-only. The symmetry between `value` prop API and ref API is usually preferred. 

But I do know we kinda over-did it with the value prop API the first time around. Is this just adding extra test coverage?

juanca:
I ran it on bigmaven and it fixes the show-page-spec too! 💯 

I think we will want Azur to look at this as part of QA. What do you think?

PR #291, opened 2021-04-07
Webinf textarea 176747702

Bloodyaugust:
## Motivation
Here is the DRD: https://docs.google.com/document/d/1MDqt9zCoFSs7022lQtLsXp0vLBKCRd9V9J241lpRUUM/edit?usp=sharing

Dev note: For now, use the same FormControl component as other field components. 

We're going to try this with resizing enabled for both axes, but this may need to change based on behavior observed in https://www.pivotaltracker.com/story/show/176508879

Remember: we like uncontrolled-ish components. The `value` API should be able to set the initial value. Or update the value when it changes. But the user can _still_ modify the value by interacting with the component.

The API will be very similar to MDS Input. Use it as a reference.

## Acceptance Criteria
API:
- label
- value
- placeholder
- required
- validationMessage
- readOnly

Documentation:

- Purposes / spiel
  - The MDS Textarea provide a multi-line text control. 
- A11y keyboard
  - N/A
- A11y label
  - The textarea has an associated label
- Examples

Browser support:

- Chrome
- Firefox
- Safari
- Edge

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-textarea-176747702/
- [x] Green Bigmaven CI check


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
   - Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
   - Append new items to make git merging easier.
+  - feat(Textarea): Implement new `Textarea` component, with new mock handlers in styleguidist
```
juanca, line 10: 💯 

I might leave the mock handlers out of the `feat(...)` changelog entry (for future commits). I think it's implied that every new component gets corresponding documentation (which includes styleguidist work, right?).

src/components/textarea/__snapshots__/textarea.test.jsx.snap
```diff
@@ -0,0 +1,36 @@
...
+
+exports[`Textarea has defaults 1`] = `
+<body>
+  <div>
+    <div
+      role="presentation"
+    >
+      <label
+        class="label"
+        for="foo"
```
juanca, line 11: AFAIK, `for` works for `<input>`s. I wonder if it works for `<textarea>`s. We might need to use `aria-labelledby` instead -- and I think that's part of the `FormControl` API already.

src/components/textarea/textarea.css
```diff
@@ -0,0 +1,40 @@
...
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.invalid-textarea {
+  composes: textarea;
+  border-color: var(--mds-red-100);
+}
+
+.invalid-textarea:hover {
```
juanca, line 38: I'm pretty sure this is not necessary:

src/components/textarea/textarea.jsx
```diff
@@ -0,0 +1,129 @@
...
+  const mounted = useMounted();
+  const [validationMessageValue, validate] = useValidation(validationMessage, textareaRef);
+
+  function blurHandler(event) {
+    validate();
+    onBlur(event);
+  }
+
+  function changeHandler(event) {
+    validate();
```
juanca, line 46: I think we will _not_ want this. Can you check this with Anne when ya pair with her? The validation spec is:

- On blur, the input goes from grey to red or red to grey
- On blur, the native validation message appears or disappears
  - Server validation messages (provided by validationMessage API) are always shown since we assume some other layer is managing it closely

src/components/textarea/textarea.jsx
```diff
@@ -0,0 +1,129 @@
...
+    onChange(event);
+  }
+
+  useEffect(() => {
```
juanca, line 50: If you're making revisions, can you check whether this should just be `useLayoutEffect` instead of `useEffect`? If so, we can remove the `mounted.current` because `useLayoutEffect` are only ran after immediately painting which only happens when the component is mounted. 

Note: we don't wanna get carried away with using `useLayoutEffect` since it will cause longer paint times. AFAIK, we only should be using it when mutating any DOM ref.

src/components/textarea/textarea.jsx
```diff
@@ -0,0 +1,129 @@
...
+import useMounted from '../../hooks/use-mounted.js';
+import useValidation from '../../hooks/use-validation.jsx';
+import styles from './textarea.css';
+
+function getClassName(className, validationMessage) {
+  if (className) return className;
+  return validationMessage ? styles['invalid-textarea'] : styles.textarea;
+}
+
+const Textarea = forwardRef(function Textarea({
```
juanca, line 20: Outta curiosity, do you prefer named (destructured) props instead of `props` parameter? Or are ya feeling things out?

src/components/textarea/textarea.css
```diff
@@ -0,0 +1,40 @@
...
+.textarea:read-only {
+  background-color: var(--mds-grey-3);
+}
+
+.invalid-icon {
+  box-sizing: border-box;
+  pointer-events: none;
+  position: absolute;
+  right: 6px;
+  top: 50%;
```
juanca, line 29: For some odd reason, this is not aligning the invalid icon in the center. If you count the pixels under the icon in Input vs Textarea then its 6px vs 4px. Somehow this `top: 50%` is not actually offsetting the icon 50% of the current height of the textarea's container. Instead its off by 2px. 

I'm not sure what is going on but 2px sounds like a border issue.

juanca: Oh I see the issue. The textarea's container is not correctly assuming the height of its children.

image

There is a gap between the bottom of the textarea and the start of the bottom margin.

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      render(<Textarea {...requiredProps} name="unique-name" ref={ref} />);
+      expect(ref.current.name).toBe('unique-name');
+    });
+  });
+
+  describe('onBlur API', () => {
+    it('sets the onblur handler', () => {
+      const onBlurSpy = jest.fn();
+      render(<Textarea {...requiredProps} onBlur={onBlurSpy} />);
+      fireEvent.blur(screen.getByLabelText('the label'));
```
juanca, line 107: We prefer `userEvent` interactions. Something like:

```suggestion
      userEvent.click(screen.getByLabelText('the label'));
      userEvent.tab();
```

This gives us a better example of what happens after user interactions instead of specific events.

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      expect(onBlurSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('onChange API', () => {
+    it('sets the onchange handler', () => {
+      const onChangeSpy = jest.fn();
+      render(<Textarea {...requiredProps} onChange={onChangeSpy} />);
+      fireEvent.change(screen.getByLabelText('the label'), { target: { value: 'new value' } });
```
juanca, line 116: We prefer `userEvent interactions. Something like:

```suggestion
      userEvent.type(screen.getByLabelText('the label'), 'new value');
```

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+    });
+
+    it('is set with any changes', () => {
+      const ref = createRef();
+      render(<Textarea {...requiredProps} ref={ref} />);
+      userEvent.type(screen.getByLabelText('the label'), 'a');
+      expect(ref.current.dirty).toBe(true);
+    });
+
+    it('is unset with changes back to initial state', () => {
```
juanca, line 53: 💯

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      const ref = createRef();
+      render(<Textarea {...requiredProps} ref={ref} />);
+      userEvent.type(screen.getByLabelText('the label'), 'a');
+      userEvent.type(screen.getByLabelText('the label'), '{backspace}');
+      expect(ref.current.dirty).toBe(false);
+    });
+
+    it('is unset with changes and new value prop', () => {
```
juanca, line 61: 💯

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      render(<Textarea {...requiredProps} ref={ref} value="test value" />);
+      expect(ref.current.value).toEqual('test value');
+    });
+  });
+
+  describe('required API', () => {
+    it('can be set', () => {
+      render(<Textarea {...requiredProps} required={true} />);
+      expect(screen.getByLabelText('the label')).toBeRequired();
+      expect(screen.getByLabelText('the label')).toBeInvalid();
```
juanca, line 147: Let's add something about not seeing the validation message.

```suggestion
      expect(screen.getByLabelText('the label')).toBeInvalid();
      expect(screen.getByText(<le error text>)).not.toBeInTheDocument();
```

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+    it('can be unset', () => {
+      render(<Textarea {...requiredProps} required={false} />);
+      expect(screen.getByLabelText('the label')).not.toBeRequired();
+      expect(screen.getByLabelText('the label')).toBeValid();
+    });
+
+    it('is invalid on mount but does not have an error message', () => {
+      render(<Textarea {...requiredProps} required={true} />);
+      expect(screen.getByLabelText('the label')).toBeInvalid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('');
```
juanca, line 159: You can merge this with the first test. But it is okay to also have another test. Just a preference thang.

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      expect(screen.getByLabelText('the label')).toHaveValue('a');
+      expect(screen.getByLabelText('the label')).toBeValid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('');
+      userEvent.type(screen.getByLabelText('the label'), '{backspace}');
+      expect(screen.getByLabelText('the label')).toHaveValue('');
+      expect(screen.getByLabelText('the label')).toBeInvalid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('Constraints not satisfied');
+    });
+
+    it('is invalid after tabbing through', async () => {
```
juanca, line 174: 💯

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+    it('is invalid after typing', async () => {
+      render(<Textarea {...requiredProps} required={true} />);
+      userEvent.type(screen.getByLabelText('the label'), 'a');
+      expect(screen.getByLabelText('the label')).toHaveValue('a');
+      expect(screen.getByLabelText('the label')).toBeValid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('');
+      userEvent.type(screen.getByLabelText('the label'), '{backspace}');
+      expect(screen.getByLabelText('the label')).toHaveValue('');
+      expect(screen.getByLabelText('the label')).toBeInvalid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('Constraints not satisfied');
```
juanca, line 171: This assertion might now be wrong with the new validation spec. Consult your friendly designer.

src/components/textarea/textarea.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+    });
+  });
+
+  describe('validationMessage API', () => {
+    it('can be set', () => {
+      render(<Textarea {...requiredProps} validationMessage="unique error" />);
+      expect(screen.getByLabelText('the label')).toBeInvalid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('unique error');
+      expect(screen.getByRole('img', { name: 'unique error' })).toBeInTheDocument();
```
juanca, line 192: SVGs are queryable with `screen.getByTitle('unique error')`.

juanca (approved):
Nothing too major. Maybe the height of the textarea container? That will make it seems bigger than the other inputs and the icon might also seem more off next to other inputs.

Other non-blocking minor improvements around testing patterns, unused CSS, new validation patterns, `useLayoutEffect`, etc

src/components/textarea/__snapshots__/textarea.test.jsx.snap
Bloodyaugust: It seems to work, but hopefully QA can be more thorough than I am. 😅

src/components/textarea/mock-handlers.js
```diff
@@ -0,0 +1,18 @@
+import { rest } from 'msw';
+import mockConstants from '../../mocks/mock-constants.js';
```
bennapp, line 2: non-blocking question: why do we need to do anything with mock-handlers for a textarea component?

src/components/textarea/textarea.jsx
Bloodyaugust: Feeling it out. Honestly it doesn't feel too bad. I'm worried about variable collisions, but didn't run into that here, and some extra lines in exchange for not needing `props.` felt nice.

src/components/textarea/textarea.css
Bloodyaugust: I think this is also at the root of the first line alignment issue, I'll try to address this.

src/components/textarea/mock-handlers.js
Bloodyaugust: I wouldn't say we _need_ to, more that this provided a nice way to do a test around an async update of the `value` prop to catch that class of bugs.

src/components/textarea/textarea.css
Bloodyaugust: They ended up being separate issues, but yeah, there's something funky with `textarea` sizing vs `input` sizing that seems to be down to browser implementation. Tweaking the padding and offset seems to have solved both issues though.

PR #293, opened 2021-04-16
MDS Date component

juanca:
https://www.pivotaltracker.com/story/show/177368115

Some nice additions: RefExample component to automatically create the ref example. :)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds-date/
- [x] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/24386


src/components/calendar/__snapshots__/calendar.test.jsx.snap
```diff
@@ -107,6 +107,7 @@ exports[`<Calendar /> has defaults 1`] = `
               class="not-current-date"
               data-epoch="1593302400000"
               role="gridcell"
+              tabindex="-1"
```
juanca, line 110: We need tab indices to populate the `relatedTarget` on Focus events (e.g. blur event)

src/components/custom-field-input-date/custom-field-input-date.md
```diff
@@ -1,4 +1,5 @@
-A simple date field for custom fields. Permits a variety of formats and allows custom validation.
+A form control to be specifically used for custom field date types. 
+It composes the MDS Date component.
```
juanca, line 2: TODO

src/components/date/date.jsx
```diff
@@ -0,0 +1,249 @@
...
+import cautionSvg from '../../svgs/caution.svg';
+import Calendar from '../calendar/calendar.jsx';
+import FormControl from '../form-control/form-control.jsx';
+import IconButton from '../icon-button/icon-button.jsx';
+import Icon from '../icon/icon.jsx';
+import useDropdownClose from '../../hooks/use-dropdown-close.js';
+import styles from './date.css';
+
+/* eslint-disable react/prop-types */
+function ControlIcons(props) {
```
juanca, line 20: Given that this is a "private" component, I opted to avoid defining a prop-type interface for now. If we do move it to its own file, that would be a good time to add a prop-types interface and test it.

src/components/date/date.md
```diff
@@ -0,0 +1,64 @@
+## Purpose
```
juanca, line 1: TODO: Move wordy documentation to Notion for a better collaborative environment.

src/components/form-control/form-control.css
```diff
@@ -1,12 +1,11 @@
...
 @import '../../styles/spacing.css';
-@import '../../styles/typography.css';
+@import '../../styles/typography-v2.css';
 
 .label {
   color: var(--mds-grey-54);
   display: inline;
-  font-family: var(--mavenlink-type-font-family); /* stylelint-disable-line mds/typography */
-  font-size: var(--mavenlink-type-small-size); /* stylelint-disable-line mds/typography */
+  font: var(--mds-type-subtext);
```
juanca, line 8: I figured it was time to use the new typography.

src/components/ref-example/ref-example.jsx
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable react/prop-types, jsx-a11y/no-static-element-interactions */
```
juanca, line 1: Same as the "private" component comment, this is "private" in terms of the MDS site. It is only used as example interfaces. I figured we could relax the linters for these.

yarn.lock
```diff
@@ -9,12 +9,17 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.8.3":
```
juanca, line 12: NOTE! I updated Jest to get typing support for date inputs.

src/components/ref-example/ref-example.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    rerender();
+  }, []);
+
+  return (
+    <div>
+      {props.children({ onChange })}
+      <table>
+        <tr>
```
juanca, line 30: TODO 
image

src/components/date/date.md
```diff
@@ -0,0 +1,64 @@
...
+```jsx
+import Date from '@mavenlink/design-system/src/components/date/date.jsx';
+import RefExample from '@mavenlink/design-system/src/components/ref-example/ref-example.jsx';
+
+const ref = React.createRef();
+
+<RefExample ref={ref}>
+  {({ onChange }) => (
+    <Date 
+      id="example-1"
```
juanca, line 56: image

src/components/date/date.jsx
```diff
@@ -0,0 +1,264 @@
...
+  const backupRef = useRef();
+  const ref = forwardedRef || backupRef;
+  const containerRef = useRef();
+  const inputRef = useRef();
+  const [active, setActive] = useState(false);
+  const [editing, setEditing] = useState(!!props.validationMessage);
+  const [expanded, setExpanded] = useState(false);
+  const [validationMessage, setValidationMessage] = useState(props.validationMessage);
+  const [value, setValue] = useState(fromFullDateFormat(props.value));
+  const classNames = {
```
juanca, line 76: I'm trying out a new CSS API for our components. Normally, I would prefer individual props prefixed with css (e.g. `cssContainer`, `cssCalendarContainer`, and `cssInput`). 

However, I do see that most of our components also have `className`. I think maybe the compromise is `className` _or_ `classNames`. Most of our complex components will need multiple classNames anyways. Though, that does imply features will be introducing one-offs (something to discourage).

src/components/date/date.jsx
```diff
@@ -0,0 +1,264 @@
...
+    label: `${props.id}-label`,
+    validationMessage: `${props.id}Hint`,
+  };
+
+  function onBlur(event) {
+    if (containerRef.current.contains(event.relatedTarget)) return;
+
+    // Given the validation spec, any provided validation messages are cleared on blur.
+    // If the native node is still invalid then re-render with the native validation messages.
+    inputRef.current.setCustomValidity('');
```
juanca, line 96: This tiny bit can be extracted to be the new validation spec.

src/components/date/date.jsx
```diff
@@ -0,0 +1,264 @@
...
+  useEffect(() => {
+    if (active && props.onChange) props.onChange({ target: ref.current });
+  }, [value]);
+
+  useLayoutEffect(() => {
+    if (active) inputRef.current.focus();
+  }, [editing]);
+
+  useLayoutEffect(() => {
+    inputRef.current.setCustomValidity(validationMessage);
```
juanca, line 161: This tiny bit can be extracted to be the new validation hook.

src/components/date/date.jsx
```diff
@@ -0,0 +1,264 @@
...
+    setValue(newDate);
+    setEditing(false);
+    setExpanded(false);
+  }
+
+  useDropdownClose(containerRef, expanded, () => setExpanded(false));
+
+  useEffect(() => {
+    setEditing(!!props.validationMessage);
+    setValidationMessage(props.validationMessage);
```
juanca, line 145: This tiny bit can be extracted to be the new validation hook.

src/components/date/date.jsx
juanca: This private component can be extracted to share logic / presentation about icons in form controls.

src/components/date/date.md
```diff
@@ -0,0 +1,42 @@
+For design and accessibility functionality, see [Form Controls > Date documentation](https://www.notion.so/Date-eb1b5fc8ce324cc39eaaa28772d055da).
```
juanca, line 1: We're going to try to flesh out the Notion document this week. We'll see if it's better -- at the very least we can have design and QA write up their expectations around the component.

src/components/calendar/calendar.jsx
```diff
@@ -37,6 +37,10 @@ function getHeadCell(iterator) {
 }
 
 function isSameDate(date1, date2) {
+  if ((!date1 && date2) || (date1 && !date2)) {
```
Bloodyaugust, line 40: Nice safety early out here

src/components/date/date.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.input {
+  --size-of-icons-container: 37px;
+
+  composes: input from '../input/input.css';
+  padding-right: calc(var(--size-of-icons-container) + var(--spacing-medium));
+}
+
+.input::-webkit-calendar-picker-indicator {
```
Bloodyaugust, line 15: Are there other vendor prefixes we need to cover as well?

src/components/date/date.jsx
Bloodyaugust: We extracted this work and styling into a new `FormControlIcons` component. We can talk about/review that here: https://github.com/mavenlink/design-system/pull/296

src/components/date/date.jsx
Bloodyaugust: For me, `className` has always been about keeping parity with the basic React API, where this sets the root element class.

src/components/date/date.jsx
Bloodyaugust: Are we not using `useValidationMessage` anymore?

src/components/date/date.jsx
Bloodyaugust: I do very much like this new pattern, though. Makes it easy to see what all of the classes actually are without needing to go into the css file.

src/components/date/date.jsx
```diff
@@ -0,0 +1,264 @@
...
+  classNames: PropTypes.shape({
+    input: PropTypes.string,
+    layout: PropTypes.shape({
+      container: PropTypes.string,
+      calendar: PropTypes.string,
+    }),
+  }),
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  /** The latest date to accept in full-date format (i.e. yyyy-mm-dd) */
```
Bloodyaugust, line 238: I think it would be interesting to explore our own validation function for these date props!

src/components/date/date.test.jsx
```diff
@@ -0,0 +1,376 @@
...
+  createRef,
+} from 'react';
+import {
+  render,
+  screen,
+} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import Date from './date.jsx';
+
+function getLocaleDate(date) {
```
Bloodyaugust, line 11: I wonder if this functionality is worth having as part of the `ref` API on the component? 🤔

src/components/date/date.test.jsx
```diff
@@ -0,0 +1,376 @@
...
+      ));
+
+      userEvent.click(screen.getByLabelText('Test label'));
+      userEvent.type(screen.getByLabelText('Test label'), ' ');
+
+      expect(onClickSpy).toBeCalledWith(expect.objectContaining({
+        defaultPrevented: true,
+        target: expect.anything(),
+      }));
+      expect(onKeyDownSpy).toBeCalledWith(expect.objectContaining({
```
Bloodyaugust, line 113: Is there another expectation we can write to ensure that some markup doesn't exist in the page?

src/components/date/date.test.jsx
```diff
@@ -0,0 +1,376 @@
...
+        defaultPrevented: true,
+        target: expect.anything(),
+      }));
+    });
+
+    it('updates the value between input types', () => {
```
Bloodyaugust, line 119: I don't quite follow what behavior we're testing here.

src/components/ref-example/ref-example.jsx
Bloodyaugust: Should we move it into the `__internal__` directory then?

Bloodyaugust (commented):
Not quite ready to hit it with an LGTM because there's a lot of questions on my end here.

In addition to the ones mentioned below, do you think it would be worth changing `Date` to `InputDate`? This would help us avoid confusion around what kind of `Date` thing we're dealing with (is it a formatted label? a functional extension? a component?), and also help us to avoid collision with `window.Date` in usage without forcing either `window.Date` or renaming the import.

src/components/date/date.css
juanca: Nope! I tried it on our supported browser and it looked good! I think previous selector was copy-pasta from SO.

src/components/date/date.jsx
juanca: > For me, `className` has always been about keeping parity with the basic React API, where this sets the root element class.

Yeah, same. I think the `className` or `classNames` might be a good pattern to avoid too much mental overhead. Who knows maybe we just use `className` as a string or an object? 🤔

src/components/date/date.jsx
juanca: At the moment, no. I think the existing hook does too much? I want to try to implement MDS Input validation by replacing the existing hook with these new `useEffect` and `useLayoutEffect` and the `onBlur` callback. But that's another story to avoid too much code changes.

src/components/date/date.jsx
juanca: Good idea! I'll write a story to create a prop-types folder and make the first one called `PropTypes.fullDate`

juanca: https://www.pivotaltracker.com/story/show/177847305

src/components/date/date.test.jsx
juanca: I don't have a use-case. 

I was thinking that maybe this is good for a test query helper thingy. But I also found that bigmaven tests use hard-coded dates. That seems pretty straight-forward to write "2020-04-20" or "Apr 20, 2020" or "April 20" or "April 2020". It makes the tests easier to read. 

I do need this function for these tests since it dynamically uses today's date. That ensures our component works when no value is provided for most of its functionality. There is some test coverage for hard-coded dates too but it's not all of them.

src/components/date/date.test.jsx
juanca: Nah, native calendar pickers are OS-based. There is nothing on the DOM to indicate it is open since the browser communicates with the OS via the event (or a system call if the event's default has not been prevented).

src/components/date/date.test.jsx
juanca: It was an edge case where React would toggle between inputs but not update the value:

- render an `<input type="date" />` with value "2020-04-20"
- switch to `<input type="text" />` with value "2020-04-20" instead of "Apr 20, 2020"

The fix was specifying a `key` between input states. Good ol' React virtual DOM diffing. 

Does that make sense? Do you think I should change something to make it more apparent in the future?

src/components/ref-example/ref-example.jsx
juanca: Nah, it's fine. React Styleguidist is configured to skip components without a `.md` file.

juanca:
> In addition to the ones mentioned below, do you think it would be worth changing `Date` to `InputDate`? This would help us avoid confusion around what kind of `Date` thing we're dealing with (is it a formatted label? a functional extension? a component?), and also help us to avoid collision with `window.Date` in usage without forcing either `window.Date` or renaming the import.

I was thinking of leaving it as `Date` for the following reasons:

1. Global usages should be prefixed with `window.` -- easier to deal with collisions, reason about code, and better linting (no-used-vars) 
2. Importing the file is already quite reasonable to assume a component but it can also be renamed on-the-fly (i.e. `import InputDate from '@mavenlink/design-system/...';`

src/components/calendar/calendar.jsx
```diff
@@ -37,6 +37,10 @@ function getHeadCell(iterator) {
...
+    return false;
+  }
```
cbacon91, line 42: I think we've skipped the (perhaps contrived or unrealistic) case where _both_ are falsey. Is it just not needed?

src/components/custom-field-input-date/custom-field-input-date.jsx
```diff
@@ -1,274 +1,22 @@
...
-    inputClassName: props.errorText ? dateStyles['input-invalid'] : dateStyles.input,
-    icon: calendarIcon(),
-    label: props.label,
-    required: props.required,
-    onClick: onInputClick,
-    onKeyDown: onInputKeyDown,
-    errorText: props.errorText,
+  const classNames = {
+    ...(props.className ? { input: props.className } : {}),
   };
```
cbacon91, line 8: I don't think we need the extraneous spreadage do we?

```suggestion
  const classNames = props.className ? { input: props.className } : {};
```

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -45,25 +45,32 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
...
-      expect(ref.current.name).toEqual('field-id');
+      expect(ref.current.name).toEqual('field-name');
     });
 
     it('has a value attribute; returns the value in yyyy-mm-dd format', () => {
       const ref = createRef();
-      render(<CustomFieldInputDate {...requiredProps} ref={ref} value={'09/27/2020'} />);
+      render(<CustomFieldInputDate {...requiredProps} ref={ref} value="2020-09-27" />);
       expect(ref.current.value).toEqual('2020-09-27');
     });
```
cbacon91, line 55: One thing I liked about the older format of the test was that it proved that even if you gave it a date in the american date format, it still returned it in ISO8601.

Is that still the case? If so, can we revert the change? If it's not the case, can we update the test name?

src/components/custom-field-input-date/custom-field-input-date.test.jsx
```diff
@@ -45,25 +45,32 @@ describe('src/components/custom-field-input-date/custom-field-input-date', () =>
...
   });
 
+  describe('className API', () => {
+    it('sets the input class name', () => {
+      render(<CustomFieldInputDate {...requiredProps} className="unique-input" />);
+      expect(document.body).toMatchSnapshot();
+    });
```
cbacon91, line 62: Is it worth writing an expectation that shows that the class name is respected? I imagine it's on the snapshot but that's not immediately obvious from the spec

src/components/ref-example/ref-example.jsx
```diff
@@ -0,0 +1,45 @@
...
+    setRerender(state => state + 1);
+  }
+
+  function onChange() {
+    rerender();
+  }
+
+  useEffect(() => {
+    rerender();
+  }, []);
```
cbacon91, line 24: super nitpick

```suggestion
  useEffect(rerender, []);
```

src/components/date/date.test.jsx
Bloodyaugust: Oh wow, that's rough! Thanks React! 😆  It makes sense now, and now that I understand it the test name seems perfect... so I'm not really sure what to change it to. 🙃  Maybe adding a comment about how React wasn't switching the value when `type` changed for extra clarity?

Bloodyaugust:
> I was thinking of leaving it as `Date` for the following reasons:
> 
> 1. Global usages should be prefixed with `window.` -- easier to deal with collisions, reason about code, and better linting (no-used-vars)
> 2. Importing the file is already quite reasonable to assume a component but it can also be renamed on-the-fly (i.e. `import InputDate from '@mavenlink/design-system/...';`

1. This makes sense for our codebase, but I think we're going to have a hard time enforcing this in Bigmaven.
2. That's what I was referring to with the "renaming the import" comment. If we're going to use it that way, it seems like it would be better to name the component the same.

src/components/calendar/calendar.jsx
Bloodyaugust: I really hope we're never calling this function in the case where the args are `null, null`, and protecting a function call where all params are `null` isn't really a pattern I want to start getting into. 😅

Bloodyaugust (approved):
LGTM! Thanks for answering all of the questions. 👍

src/components/calendar/calendar.jsx
juanca: In our case, one of the params is always present. We use this helper method to figure out whether the thing we're rendering (that represents a date) is equal to some state (another date).

src/components/custom-field-input-date/custom-field-input-date.test.jsx
juanca: yeah, the `value` prop API now is always full-date ISO8601 (i.e. `YYYY-MM-DD`). 

We should always user ISO8601 full-date or date-time for programming bits purposes (or native Date functionality).

juanca: I think the test name is still consistent. Ideally, I would just go even simpler:

```
it('has a string value')
```

The implementation details are up to the tests themselves.

src/components/custom-field-input-date/custom-field-input-date.test.jsx
juanca: Yeah, it's in the snapshot. Whenever the DOM changes, a developer will need to regenerate the snapshot and make a conscious decision to accept the new markup. For this test, they will focus on class names on the input. 

I still haven't found the perfect balance between snapshots vs exact assertions. 

In the `date/date.test.jsx` suite, we have a similar test for `classNames API`. It makes more sense to use a snapshot on that since it tests 3 difference class names. I think repeating the pattern for `className API` is probably going to be a lower mental overhead.

juanca:
> 1. This makes sense for our codebase, but I think we're going to have a hard time enforcing this in Bigmaven.

Yeah, I'm going to try to get us to the point where the linter automatically enforces this. This is a baby step in the right direction.

> 2. That's what I was referring to with the "renaming the import" comment. If we're going to use it that way, it seems like it would be better to name the component the same.

Good point. Here's a compromise: if we see a lot of renaming or negative feedback about Date component then we plan on renaming it.  `FormDate` doesn't sound too bad -- namespacing our form controls might be a good idea (FormInput, FormNumber, FormDate, ...).

PR #296, opened 2021-04-20
MDS Percentage component

Bloodyaugust:
Co-authored-by: Chris Bacon <cbacon@mavenlink.com>
## Motivation
We need an easy way for a user to set a percentage datum. 

## Acceptance Criteria
- The Percentage component looks and acts like an Number
  - If possible, hide the up/down arrows. 
- The `%` is on the right side with appropriate spacing
  - Note: it is to the left of the invalid icon
  - It is not focusable / interactive
  - It does not have a label
  - It is not clickable -- instead it focuses the <input>
- Validations
  - It validates on blur (any messages appear/disappear)
  - 0.00 <= n <= 100.00
  - Only 2 decimal points (no rounding!)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-mds-percentage-177481296/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/form-control-icons/form-control-icons.css
+++ b/src/components/form-control-icons/form-control-icons.css
@@ -1,0 +1,13 @@
+@import '../../styles/spacing.css';
+@import '../../styles/colors-v2.css';
+
+.icons {
+  align-items: center;
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: var(--spacing-small);
+  position: absolute;
+  right: var(--spacing-medium);
+  transform: translateY(-50%);
+  top: 50%;
+}
--- a/src/components/form-control-icons/form-control-icons.css
+++ b/src/components/form-control-icons/form-control-icons.css
@@ -1,5 +1,4 @@
 @import '../../styles/spacing.css';
-@import '../../styles/colors-v2.css';
 
 .icons {
   align-items: center;
--- a/src/components/form-control-icons/form-control-icons.jsx
+++ b/src/components/form-control-icons/form-control-icons.jsx
@@ -13,7 +13,6 @@ export default function FormControlIcons({
     <div className={className}>
       {!!validationMessage && (
         <Icon
-          className={styles['invalid-icon']}
           icon={cautionSvg}
           label={validationMessage}
         />
--- a/src/components/form-control-icons/form-control-icons.jsx
+++ b/src/components/form-control-icons/form-control-icons.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../icon/icon.jsx';
+import cautionSvg from '../../svgs/caution.svg';
+import styles from './form-control-icons.css';
+
+export default function FormControlIcons({
+  children,
+  className,
+  validationMessage,
+}) {
+  return (
+    <div className={className}>
+      {!!validationMessage && (
+        <Icon
+          className={styles['invalid-icon']}
+          icon={cautionSvg}
+          label={validationMessage}
+        />
+      )}
+      {children}
+    </div>
+  );
+}
+
+FormControlIcons.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  validationMessage: PropTypes.string,
+};
+
+FormControlIcons.defaultProps = {
+  children: undefined,
+  className: styles.icons,
+  validationMessage: '',
+};
--- a/src/components/percentage/__snapshots__/percentage.test.jsx.snap
+++ b/src/components/percentage/__snapshots__/percentage.test.jsx.snap
@@ -46,6 +46,6 @@ Object {
   "dirty": false,
   "id": "foo",
   "name": undefined,
-  "value": NaN,
+  "value": undefined,
 }
 `;
--- a/src/components/percentage/__snapshots__/percentage.test.jsx.snap
+++ b/src/components/percentage/__snapshots__/percentage.test.jsx.snap
@@ -1,0 +1,51 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Percentage has defaults 1`] = `
+<body>
+  <div>
+    <div
+      class="container"
+      role="presentation"
+    >
+      <label
+        class="label"
+        for="foo"
+      >
+        the label
+      </label>
+      <div
+        class="control-container"
+      >
+        <input
+          aria-describedby="fooHint"
+          class="input"
+          id="foo"
+          max="100"
+          min="0"
+          step="0.01"
+          type="number"
+          value=""
+        />
+        <div
+          class="icons-container"
+        >
+          <span
+            class="percent-sign"
+          >
+            %
+          </span>
+        </div>
+      </div>
+    </div>
+  </div>
+</body>
+`;
+
+exports[`Percentage has defaults 2`] = `
+Object {
+  "dirty": false,
+  "id": "foo",
+  "name": undefined,
+  "value": NaN,
+}
+`;
--- a/src/components/percentage/percentage.css
+++ b/src/components/percentage/percentage.css
@@ -8,14 +8,8 @@
 }
 
 .input {
-  border: 1px solid var(--mds-grey-12);
-  border-radius: 3px;
-  box-sizing: border-box;
-  font: var(--mds-type-content);
-  height: var(--spacing-x-large);
-  margin: 0;
+  composes: input from "../input/input.css";
   padding: 0 calc(var(--spacing-x-large) + var(--spacing-large)) 0 var(--spacing-medium);
-  width: 100%;
 }
 
 .input::placeholder {
@@ -26,22 +20,9 @@
   background-color: var(--mds-grey-3);
 }
 
-.invalid-icon {
-  box-sizing: border-box;
-  pointer-events: none;
-  position: absolute;
-  right: 6px;
-  top: 50%;
-  transform: translateY(-50%);
-}
-
 .invalid-input {
+  composes: invalid-input from "../input/input.css";
   composes: input;
-  border-color: var(--mds-red-100);
-}
-
-.invalid-input:hover {
-  border-color: var(--mds-red-100);
 }
 
 .percent-sign {
--- a/src/components/percentage/percentage.css
+++ b/src/components/percentage/percentage.css
@@ -1,0 +1,51 @@
+@import '../../styles/colors-v2.css';
+@import "../../styles/spacing.css";
+@import "../../styles/typography-v2.css";
+
+.icons-container {
+  composes: icons from "../form-control-icons/form-control-icons.css";
+  pointer-events: none;
+}
+
+.input {
+  border: 1px solid var(--mds-grey-12);
+  border-radius: 3px;
+  box-sizing: border-box;
+  font: var(--mds-type-content);
+  height: var(--spacing-x-large);
+  margin: 0;
+  padding: 0 calc(var(--spacing-x-large) + var(--spacing-large)) 0 var(--spacing-medium);
+  width: 100%;
+}
+
+.input::placeholder {
+  color: var(--mds-grey-54);
+}
+
+.input:read-only {
+  background-color: var(--mds-grey-3);
+}
+
+.invalid-icon {
+  box-sizing: border-box;
+  pointer-events: none;
+  position: absolute;
+  right: 6px;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.invalid-input {
+  composes: input;
+  border-color: var(--mds-red-100);
+}
+
+.invalid-input:hover {
+  border-color: var(--mds-red-100);
+}
+
+.percent-sign {
+  color: var(--mds-grey-38);
+  font: var(--mds-type-content);
+  user-select: none;
+}
--- a/src/components/percentage/percentage.jsx
+++ b/src/components/percentage/percentage.jsx
@@ -1,0 +1,148 @@
+import PropTypes from 'prop-types';
+import React, {
+  forwardRef,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+} from 'react';
+import FormControl from '../form-control/form-control.jsx';
+import FormControlIcons from '../form-control-icons/form-control-icons.jsx';
+import styles from './percentage.css';
+import useMounted from '../../hooks/use-mounted.js';
+import useValidation from '../../hooks/use-validation.jsx';
+
+function getClassName(className, validationMessage) {
+  if (className) return className;
+  return validationMessage ? styles['invalid-input'] : styles.input;
+}
+
+const Percentage = forwardRef(function Percentage(props, forwardedRef) {
+  const fallbackRef = useRef();
+  const ref = forwardedRef || fallbackRef;
+  const inputRef = useRef();
+  const mounted = useMounted();
+  const [validationMessage, validate] = useValidation(props.validationMessage, inputRef);
+
+  function onBlur(event) {
+    validate();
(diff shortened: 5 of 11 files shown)


src/components/percentage/__snapshots__/percentage.test.jsx.snap
```diff
@@ -0,0 +1,51 @@
...
+  </div>
+</body>
+`;
+
+exports[`Percentage has defaults 2`] = `
+Object {
+  "dirty": false,
+  "id": "foo",
+  "name": undefined,
+  "value": NaN,
```
cbacon91, line 49: I know this is just a snapshot, but is it weird that the default value for a percentage is `NaN` and not `undefined` or something like that?

I mean, it makes sense - we call `parseFloat` on undefined, just noticed it here

juanca: This might be indicative of something wrong too.

If the user types something, saves it, and then deletes the value, is it NaN? If so, we would try to persist the `value` on the ref and the backend might not be happy.

juanca: I think we'll want to follow the same API as native inputs. If there is no value, it returns undefined or an empty string. I'm not sure which is best for this scenario -- but less code the better?

src/components/percentage/percentage.md
```diff
@@ -0,0 +1,17 @@
+The `Percentage` component allows for representing a percentage in the range of [0, 100] with up to 2 significant digits.
```
juanca, line 1: We need a ref example

src/components/percentage/__snapshots__/percentage.test.jsx.snap
Bloodyaugust: I decided to standardize on `undefined`, because I think we definitely want to avoid mixing string values as a possibility.

src/components/percentage/percentage.md
Bloodyaugust: Added!

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+  const mounted = useMounted();
+  const [validationMessage, validate] = useValidation(props.validationMessage, inputRef);
+
+  function onBlur(event) {
+    validate();
+    props.onBlur(event);
+  }
+
+  function onChange(event) {
+    validate();
```
juanca, line 32: AFAIK, this is no longer the validation spec. We only validate on blur. Is there a use-case for this?

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+    // The MDS Input is using an uncontrolled `<input>`.
+    // In order to set a new provided value prop, we
+    // set the internal state of the `<input>`.
+    inputRef.current.value = props.value || '';
+  }, [props.value]);
+
+  useImperativeHandle(ref, () => ({
+    get dirty() {
+      const providedValue = props.value || undefined;
+      return !Object.is(providedValue, this.value);
```
juanca, line 48: TIL: Object.is 💯

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+    },
+    id: props.id,
+    name: props.name,
+    get value() {
+      return parseFloat(inputRef.current.value) || undefined;
```
juanca, line 53: In case ya wanna get in the habit of referencing the global instead its API. 

```suggestion
       return window.parseFloat(inputRef.current.value) || undefined;
```

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+        className={getClassName(props.className, validationMessage)}
+        defaultValue={props.value}
+        id={props.id}
+        max={100}
+        min={0}
+        step={0.01}
+        name={props.name}
+        onBlur={onBlur}
+        onChange={onChange}
+        onFocus={props.onFocus}
```
juanca, line 78: Do we have a use-case for this event? Or is it some foreseeable use-case? (Same for `onInput`, `onKeyDown`, `props.onBlur`)

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  onInput: PropTypes.func,
+  onKeyDown: PropTypes.func,
+  placeholder: PropTypes.string,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  validationMessage: PropTypes.string,
+  value: (props, propName, componentName) => {
```
juanca, line 110: 🚀

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      render(<Percentage {...requiredProps} name="unique-name" ref={ref} />);
+      expect(ref.current.name).toBe('unique-name');
+    });
+  });
+
+  describe('onBlur API', () => {
+    it('sets the onblur handler', () => {
+      const onBlurSpy = jest.fn();
+      render(<Percentage {...requiredProps} onBlur={onBlurSpy} />);
+      fireEvent.blur(screen.getByLabelText('the label'));
```
juanca, line 114: We prefer to use user-events for interacting with the component. 

```suggestion
      userEvent.tab({ shift: true });
```

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      fireEvent.change(screen.getByLabelText('the label'), { target: { value: '100' } });
+      expect(onChangeSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('onFocus API', () => {
+    it('sets the onFocus handler', () => {
+      const onFocusSpy = jest.fn();
+      render(<Percentage {...requiredProps} onFocus={onFocusSpy} />);
+      fireEvent.focus(screen.getByLabelText('the label'));
```
juanca, line 132: We prefer to use user-events for interacting with the component.

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      expect(onFocusSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('onInput API', () => {
+    it('sets the onInput handler', () => {
+      const onInputSpy = jest.fn();
+      render(<Percentage {...requiredProps} onInput={onInputSpy} />);
+      fireEvent.input(screen.getByLabelText('the label'));
```
juanca, line 141: We prefer to use user-events for interacting with the component.

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      expect(onInputSpy.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('onKeyDown API', () => {
+    it('sets the onKeyDown handler', () => {
+      const onKeyDownSpy = jest.fn();
+      render(<Percentage {...requiredProps} onKeyDown={onKeyDownSpy} />);
+      fireEvent.keyDown(screen.getByLabelText('the label'));
```
juanca, line 150: We prefer user-events.

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+
+  describe('placeholder API', () => {
+    it('sets the placeholder attribute', () => {
+      render(<Percentage {...requiredProps} placeholder="test-placeholder" />);
+      expect(screen.getByLabelText('the label')).toHaveAttribute('placeholder', 'test-placeholder');
+    });
+  });
+
+  describe('readOnly API', () => {
+    it('sets the readOnly attribute', () => {
```
juanca, line 163: Since `readOnly` is a boolean, we should be exhaustive and also set it to `false` in a test.

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      render(<Percentage {...requiredProps} readOnly />);
+      expect(screen.getByLabelText('the label')).toHaveAttribute('readonly');
+    });
+  });
+
+  describe('ref API', () => {
+    it('can be set', () => {
+      const ref = React.createRef();
+      render(<Percentage {...requiredProps} ref={ref} value={100} />);
+      expect(ref.current.value).toEqual(100);
```
juanca, line 173: You know what could be cool (but no use-case for now). We could probably expose the native element as a key (so no key-value copying) on our imperative ref. Just like React events, the real native things can be `ref.current.native`. Just food for thought.

src/components/percentage/percentage.test.jsx
```diff
@@ -0,0 +1,261 @@
...
+      expect(screen.getByLabelText('the label')).toBeValid();
+    });
+
+    it('is invalid on mount but does not have an error message', () => {
+      render(<Percentage {...requiredProps} required={true} />);
+      expect(screen.getByLabelText('the label')).toBeInvalid();
+      expect(screen.getByLabelText('the label')).toHaveDescription('');
+    });
+
+    it('is invalid after typing', async () => {
```
juanca, line 196: This is no longer part of the ref spec. It should not have an error message after typing. Only after blurring.

src/components/percentage/percentage.css
```diff
@@ -0,0 +1,51 @@
+@import '../../styles/colors-v2.css';
+@import "../../styles/spacing.css";
+@import "../../styles/typography-v2.css";
+
+.icons-container {
+  composes: icons from "../form-control-icons/form-control-icons.css";
+  pointer-events: none;
+}
+
+.input {
```
juanca, line 10: Should this stylesheet compose from the MDS input?

src/components/percentage/percentage.css
```diff
@@ -0,0 +1,51 @@
...
+.invalid-icon {
+  box-sizing: border-box;
+  pointer-events: none;
+  position: absolute;
+  right: 6px;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.invalid-input {
```
juanca, line 38: We might be able to get this spec by composing from the MDS input.

src/components/percentage/percentage.css
```diff
@@ -0,0 +1,51 @@
...
+
+.input::placeholder {
+  color: var(--mds-grey-54);
+}
+
+.input:read-only {
+  background-color: var(--mds-grey-3);
+}
+
+.invalid-icon {
```
juanca, line 29: This isn't used.

src/components/percentage/percentage.css
```diff
@@ -0,0 +1,51 @@
...
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.invalid-input {
+  composes: input;
+  border-color: var(--mds-red-100);
+}
+
+.invalid-input:hover {
```
juanca, line 43: Same as the MDS Textarea requested revisions, I'm pretty sure this isn't doing anything.

https://github.com/mavenlink/design-system/pull/291#discussion_r611930754

image

src/components/form-control-icons/form-control-icons.jsx
```diff
@@ -0,0 +1,36 @@
...
+export default function FormControlIcons({
+  children,
+  className,
+  validationMessage,
+}) {
+  return (
+    <div className={className}>
+      {!!validationMessage && (
+        <Icon
+          className={styles['invalid-icon']}
```
juanca, line 16: This isn't a real stylesheet selector.

src/components/form-control-icons/form-control-icons.css
```diff
@@ -0,0 +1,13 @@
+@import '../../styles/spacing.css';
+@import '../../styles/colors-v2.css';
+
+.icons {
+  align-items: center;
+  display: grid;
```
juanca, line 6: Good use of `grid`!

src/components/form-control-icons/form-control-icons.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default function FormControlIcons({
+  children,
+  className,
+  validationMessage,
+}) {
+  return (
+    <div className={className}>
+      {!!validationMessage && (
+        <Icon
```
juanca, line 15: Now that's this is in a separate file, something is itching in my head that there should be some HTML or ARIA relationship between the icon and the control. But I don't know what that is. /shrug

juanca (approved):
A lot of requested changes that are pretty straight-forward:

- clean up unused things
- hoist classes, ids as a pattern
- compose from other MDS stylesheets
- test improvements with user-events instead of low-level events
- ...

I'm not sure sure about introducing a larger API that is needed (onFocus, onBlur, etc.). On the one hand, I would like to avoid making it easy to make UX one-offs in bigmaven. On the other hand, we aren't spreading the props so it's a fairly small one-off vector (just like classnames API).

src/components/percentage/percentage.jsx
Bloodyaugust: I learned about this as part of this PR as well! It's super dope. The use case _was_ to support checking `NaN === NaN`, so I can change this back if you like.

src/components/percentage/percentage.jsx
```diff
@@ -0,0 +1,148 @@
...
+      className={props.cssContainer}
+      error={validationMessage}
+      id={props.id}
+      label={props.label}
+      readOnly={props.readOnly}
+      required={props.required}
+    >
+      <input
+        aria-describedby={`${props.id}Hint`}
+        autoFocus={props.autoFocus} // eslint-disable-line jsx-a11y/no-autofocus
```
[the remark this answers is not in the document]
Bloodyaugust: We should backfill this to our other input-oriented components as well.

src/components/percentage/percentage.jsx
Bloodyaugust: A lot of these are me trying to maintain functional parity with our other `input`-based components. I can remove the stuff we don't currently use, if you like.

src/components/percentage/percentage.css
Bloodyaugust: Probably!

src/components/form-control-icons/form-control-icons.jsx
Bloodyaugust: I've done some digging, and I agree, but no attributes jump out to me as obvious candidates. 😬

src/components/percentage/percentage.test.jsx
Bloodyaugust: Like you said, there's no use-case currently, so I'm avoiding it for this PR. Definitely something we should consider in the future when we have a need for access to it, though.

src/components/percentage/percentage.jsx
Bloodyaugust: I did remove all of these except for `onChange`.

src/components/percentage/percentage.jsx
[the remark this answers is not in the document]
juanca: As long as the inputs aren't using the god abstract field component then i think they're good to go.

src/components/percentage/percentage.jsx
juanca: Yeah, I'm not entirely convinced we need a large events API. Or if we do then it should come with some design UX feedback, like: we want our form fields to auto-save on blur or something.

src/components/percentage/percentage.jsx
juanca: Nah, either way if fine.

PR #298, opened 2021-04-22
Webinf autocompleter 177548593

bennapp:
## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-autocompleter-177548593/
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -32,10 +32,6 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
     return `${API_ROOT}${props.apiEndpoint}`;
   }
 
-  function displayName(modelInfo) {
-    return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
-  }
-
   function onKeyUp(event) {
     fetchModels(event.target.value);
   }
@@ -44,7 +40,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   const listOptionElements = models.map((modelInfo, index) => {
     return (
       <ListOption key={modelInfo.id} ref={listOptionRefs[index]} value={modelInfo}>
-        {displayName(modelInfo)}
+        {props.displayValueEvaluator(modelInfo)}
       </ListOption>
     );
   });
@@ -56,7 +52,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
       name={props.name}
       className={props.className}
       listOptionRefs={listOptionRefs}
-      displayValueEvaluator={displayName}
+      displayValueEvaluator={props.displayValueEvaluator}
       onChange={props.onChange}
       onKeyUp={onKeyUp}
       required={props.required}
@@ -71,14 +67,18 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   );
 });
 
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
 Autocompleter.propTypes = {
   apiEndpoint: PropTypes.string.isRequired,
-  // singleSelect: PropTypes.bool,
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   searchParam: PropTypes.string,
   models: PropTypes.arrayOf(PropTypes.object),
   onChange: PropTypes.func,
+  displayValueEvaluator: PropTypes.func,
   label: PropTypes.string,
   required: PropTypes.bool,
   readOnly: PropTypes.bool,
@@ -94,6 +94,7 @@ Autocompleter.defaultProps = {
   label: '',
   required: false,
   readOnly: false,
+  displayValueEvaluator: displayName,
   onChange: () => {},
   placeholder: '',
   className: '',
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -82,17 +82,17 @@ function displayName(modelInfo) {
 }
 
 Autocompleter.propTypes = {
-  // `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
+  /* `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`. */
   apiEndpoint: PropTypes.string,
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   searchParam: PropTypes.string,
-  // `value` and `models` shape is expected to be an object(s) with an `id` key
+  /* `value` and `models` shape is expected to be an object(s) with an `id` key */
   models: PropTypes.arrayOf(PropTypes.shape({
     id: PropTypes.string.required,
   })),
   onChange: PropTypes.func,
-  // displayValueEvaluator is handled if the key following: `title`, `name`, `full_name`, `currency`; Otherwise, pass in something like `displayValueEvaluator: (model) -> { model.rate_card_name }`
+  /* displayValueEvaluator is handled if the key following: `title`, `name`, `full_name`, `currency`; Otherwise, pass in something like `displayValueEvaluator: (model) -> { model.rate_card_name }` */
   displayValueEvaluator: PropTypes.func,
   label: PropTypes.string,
   required: PropTypes.bool,
@@ -100,7 +100,7 @@ Autocompleter.propTypes = {
   placeholder: PropTypes.string,
   className: PropTypes.string,
   validationMessage: PropTypes.string,
-  // `value` and `models` shape is expected to be an object(s) with an `id` key
+  /* `value` and `models` shape is expected to be an object(s) with an `id` key */
   value: PropTypes.any, // eslint-disable-line react/forbid-prop-types
 };
 
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -103,7 +103,7 @@ Autocompleter.propTypes = {
 };
 
 Autocompleter.defaultProps = {
-  searchParam: 'search',
+  searchParam: 'matching',
   models: [],
   label: undefined,
   required: false,
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -22,6 +22,8 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   useEffect(fetchModels, []);
 
   function fetchModels(searchString) {
+    if (!props.apiEndpoint) { return; }
+
     execute(apiEndpoint(searchString))
       .then((respObj) => {
         return setModels(respObj.json.results.map(result => respObj.json[result.key][result.id]));
@@ -81,7 +83,7 @@ function displayName(modelInfo) {
 
 Autocompleter.propTypes = {
   // `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
-  apiEndpoint: PropTypes.string.isRequired,
+  apiEndpoint: PropTypes.string,
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   searchParam: PropTypes.string,
@@ -103,6 +105,7 @@ Autocompleter.propTypes = {
 };
 
 Autocompleter.defaultProps = {
+  apiEndpoint: undefined,
   searchParam: 'matching',
   models: [],
   label: undefined,
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -91,14 +91,14 @@ Autocompleter.propTypes = {
 Autocompleter.defaultProps = {
   searchParam: 'search',
   models: [],
-  label: '',
+  label: undefined,
   required: false,
   readOnly: false,
   displayValueEvaluator: displayName,
   onChange: () => {},
-  placeholder: '',
-  className: '',
-  errorText: '',
+  placeholder: undefined,
+  className: undefined,
+  errorText: undefined,
   value: undefined,
 };
 
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -11,9 +11,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   const { execute: fetch } = useFetch();
   const [models, setModels] = useState(props.models);
 
-  useEffect(() => {
-    fetchModels();
-  }, []);
+  useEffect(fetchModels, []);
 
   function fetchModels(searchString) {
     fetch(apiEndpoint(searchString))
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -82,17 +82,17 @@ function displayName(modelInfo) {
 }
 
 Autocompleter.propTypes = {
-  /* `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`. */
+  /** `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`. */
   apiEndpoint: PropTypes.string,
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   searchParam: PropTypes.string,
-  /* `value` and `models` shape is expected to be an object(s) with an `id` key */
+  /** `value` and `models` shape is expected to be an object(s) with an `id` key */
   models: PropTypes.arrayOf(PropTypes.shape({
     id: PropTypes.string.required,
   })),
   onChange: PropTypes.func,
-  /* displayValueEvaluator is handled if the key following: `title`, `name`, `full_name`, `currency`; Otherwise, pass in something like `displayValueEvaluator: (model) -> { model.rate_card_name }` */
+  /** displayValueEvaluator is handled if the key following: `title`, `name`, `full_name`, `currency`; Otherwise, pass in something like `displayValueEvaluator: (model) -> { model.rate_card_name }` */
   displayValueEvaluator: PropTypes.func,
   label: PropTypes.string,
   required: PropTypes.bool,
@@ -100,7 +100,7 @@ Autocompleter.propTypes = {
   placeholder: PropTypes.string,
   className: PropTypes.string,
   validationMessage: PropTypes.string,
-  /* `value` and `models` shape is expected to be an object(s) with an `id` key */
+  /** `value` and `models` shape is expected to be an object(s) with an `id` key */
   value: PropTypes.any, // eslint-disable-line react/forbid-prop-types
 };
 
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -31,8 +31,6 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
       return `${API_ROOT}${props.apiEndpoint}?${props.searchParam}=${searchString}`;
     }
 
-    console.log(`${API_ROOT}${props.apiEndpoint}`)
-
     return `${API_ROOT}${props.apiEndpoint}`;
   }
 
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -16,7 +16,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   function fetchModels(searchString) {
     fetch(apiEndpoint(searchString))
       .then((results) => {
-        setModels(results.json.results.map(resultInfo => results.json[resultInfo.key][resultInfo.id]));
+        setModels(respObj.json.results.map(result => respObj.json[result.key][result.id]));
       }).catch((error) => {
         if (error.error && error.error.type !== 'aborted') {
           throw error;
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -1,0 +1,108 @@
+import PropTypes from 'prop-types';
+import React, {forwardRef, useEffect, useState} from 'react';
+import useFetch from '@bloodyaugust/use-fetch';
+import Select from '../select/select.jsx';
+import ListOption from '../list-option/list-option.jsx';
+import mockConstants from '../../mocks/mock-constants.js';
+
+const { API_ROOT } = mockConstants;
+
+const Autocompleter = forwardRef(function Autocompleter(props, ref) {
+  const { execute: fetch } = useFetch();
+  const [models, setModels] = useState(props.models);
+
+  useEffect(() => {
+    fetchModels();
+  }, []);
+
+  function fetchModels(searchString) {
+    fetch(apiEndpoint(searchString))
+      .then((results) => {
+        setModels(results.json.results.map(resultInfo => results.json[resultInfo.key][resultInfo.id]));
+      }).catch((error) => {
+        if (error.error && error.error.type !== 'aborted') {
+          throw error;
+        }
+      });
+  }
+
+  function apiEndpoint(searchString) {
+    if (searchString) {
+      return `${API_ROOT}${props.apiEndpoint}?${props.searchParam}=${searchString}`;
+    }
+
+    console.log(`${API_ROOT}${props.apiEndpoint}`)
+
+    return `${API_ROOT}${props.apiEndpoint}`;
+  }
+
+  function displayName(modelInfo) {
+    return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+  }
+
+  function onKeyUp(event) {
+    fetchModels(event.target.value);
+  }
+
+  const listOptionRefs = models.map(() => React.createRef());
+  const listOptionElements = models.map((modelInfo, index) => {
+    return (
+      <ListOption key={modelInfo.id} ref={listOptionRefs[index]} value={modelInfo}>
+        {displayName(modelInfo)}
+      </ListOption>
+    );
+  });
+
+  return (
+    <Select
+      label={props.label}
+      id={props.id}
+      name={props.name}
+      className={props.className}
+      listOptionRefs={listOptionRefs}
+      displayValueEvaluator={displayName}
+      onChange={props.onChange}
+      onKeyUp={onKeyUp}
+      required={props.required}
+      readOnly={props.readOnly}
+      placeholder={props.placeholder}
+      errorText={props.errorText}
+      value={props.value}
+      ref={ref}
+    >
+      {listOptionElements}
+    </Select>
+  );
+});
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
+  // singleSelect: PropTypes.bool,
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  searchParam: PropTypes.string,
+  models: PropTypes.arrayOf(PropTypes.object),
+  onChange: PropTypes.func,
+  label: PropTypes.string,
+  required: PropTypes.bool,
+  readOnly: PropTypes.bool,
+  placeholder: PropTypes.string,
+  className: PropTypes.string,
+  errorText: PropTypes.string,
+  value: PropTypes.any, // eslint-disable-line react/forbid-prop-types
+};
+
+Autocompleter.defaultProps = {
+  searchParam: 'search',
+  models: [],
+  label: '',
+  required: false,
+  readOnly: false,
+  onChange: () => {},
+  placeholder: '',
+  className: '',
+  errorText: '',
+  value: undefined,
+};
+
+export default Autocompleter;
--- a/src/components/autocompleter/autocompleter.jsx
+++ b/src/components/autocompleter/autocompleter.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { forwardRef, useEffect, useState } from 'react';
 import useFetch from '@bloodyaugust/use-fetch';
+import useMounted from '../../hooks/use-mounted.js';
 import Select from '../select/select.jsx';
 import ListOption from '../list-option/list-option.jsx';
 import mockConstants from '../../mocks/mock-constants.js';
@@ -8,15 +9,22 @@ import mockConstants from '../../mocks/mock-constants.js';
 const { API_ROOT } = mockConstants;
 
 const Autocompleter = forwardRef(function Autocompleter(props, ref) {
-  const { execute: fetch } = useFetch();
+  const { execute } = useFetch();
+  const mounted = useMounted();
   const [models, setModels] = useState(props.models);
+
+  useEffect(() => {
+    if (!mounted.current) return;
+
+    setModels(props.models);
+  }, [props.models]);
 
   useEffect(fetchModels, []);
 
   function fetchModels(searchString) {
-    fetch(apiEndpoint(searchString))
-      .then((results) => {
-        setModels(respObj.json.results.map(result => respObj.json[result.key][result.id]));
+    execute(apiEndpoint(searchString))
+      .then((respObj) => {
+        return setModels(respObj.json.results.map(result => respObj.json[result.key][result.id]));
       }).catch((error) => {
         if (error.error && error.error.type !== 'aborted') {
           throw error;
@@ -32,7 +40,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
     return `${API_ROOT}${props.apiEndpoint}`;
   }
 
-  function onKeyUp(event) {
+  function onInput(event) {
     fetchModels(event.target.value);
   }
 
@@ -54,7 +62,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
       listOptionRefs={listOptionRefs}
       displayValueEvaluator={props.displayValueEvaluator}
       onChange={props.onChange}
-      onKeyUp={onKeyUp}
+      onInput={onInput}
       required={props.required}
       readOnly={props.readOnly}
       placeholder={props.placeholder}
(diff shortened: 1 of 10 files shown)


src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+      ref={ref}
+    >
+      {listOptionElements}
+    </Select>
+  );
+});
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
+  // singleSelect: PropTypes.bool,
```
cbacon91, line 78: I take it this is part of what spurred the discussion around single-vs-multiselect in retro

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,17 @@
+The `Autocompleter` component is a wrapper for Select component with the additional functionality searching for a specific models from an API endpoint.
+
+`apiEndpoint`, should be the remainder of the apiEndpoint (excluding the base api), eg. `/workspaces`.
```
cbacon91, line 3:
```suggestion
 `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
```

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,17 @@
...
+`value` and `models` shape is expected to be an object(s) with an `id` key, and a display attribute that is one of the following: `title`, `name`, `full_name`, `currency`;
+
+For a11z, and additional prop info, see the `Select` component.
```
cbacon91, line 6:
```suggestion
For a11y and additional prop info, see the `Select` component.
```

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+const { API_ROOT } = mockConstants;
+
+const Autocompleter = forwardRef(function Autocompleter(props, ref) {
+  const { execute: fetch } = useFetch();
+  const [models, setModels] = useState(props.models);
+
+  useEffect(() => {
+    fetchModels();
+  }, []);
```
cbacon91, line 16: nitpick

```suggestion
  useEffect(fetchModels, []);
```

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+  function fetchModels(searchString) {
+    fetch(apiEndpoint(searchString))
```
cbacon91, line 19: nitpick: Since this is a promise, it might be a good idea to `return` it. We aren't doing anything with it now, but I think in general that's a good practice

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+  function apiEndpoint(searchString) {
+    if (searchString) {
+      return `${API_ROOT}${props.apiEndpoint}?${props.searchParam}=${searchString}`;
+    }
+
+    return `${API_ROOT}${props.apiEndpoint}`;
+  }
+
+  function displayName(modelInfo) {
+    return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
```
cbacon91, line 38: `currency` is jarring to me here, one thing we could provide is a way for the consumer of this to supply a prop that is something like a key to know what to present, ie 

```

<Metacompleter labelKey="currency" />

function displayName(modelInfo) { 
  if(props.labelKey) { return modelInfo[labelKey]; }

  return modelInfo.title || modelInfo.name || modelInfo.full_name;
}
```

That way we don't need to try to guess at what it could possibly be. In a lot of cases, `title` and `name` will cover it, and `user` is used so often that `full_name` is nice to have here, but there will always be edge cases where someone decides to name the "user-friendly" name of something as like `display_name`, `capitalized_name`, or a one-off like `currency` is.

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+  it('has defaults', () => {
+    const ref = createRef();
+    render(<Autocompleter {...requiredProps} ref={ref} />);
+    expect(document.body).toMatchSnapshot();
+    expect(ref.current).toMatchSnapshot();
+  });
+
+  describe('className API', () => {
+    it('prioritizes className prop', () => {
```
cbacon91, line 30: info for myself - what does `prioritizes` mean in this context? That it uses the class name provided and not the default or something?

If that's the case, should we make sure a class name _isn't_ there? If that's _not_ the case, can this test be a bit clearer like `it('applies the classname to the select')` or something?

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+    it('accepts an ID', () => {
+      render(<Autocompleter {...requiredProps} id="this-is-an-id" />);
+      expect(screen.getByLabelText('Test label')).toHaveAttribute('id', 'this-is-an-id');
+    });
+  });
+
+  describe('label API', () => {
+    it('accepts a label', () => {
+      render(<Autocompleter {...requiredProps} label="Bar" />);
+      expect(screen.getByLabelText('Bar')).toBeDefined();
```
cbacon91, line 53: info for myself - is there a reason to use `toBeDefined` over `toBeInTheDocument`?

It's sixes in this test since the `get` would fail, but I've seen this elsewhere and been curious about it

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+  });
+
+  describe('models API', () => {
+    it('accepts a an array of models', async () => {
+      const models = [{ id: 22, name: 'cool dude' }, { id: 33, name: 'neato burrito' }];
+      render(<Autocompleter {...requiredProps} models={models} />);
+
+      userEvent.click(screen.getByLabelText('Test label'));
+      expect(screen.queryByText('cool dude')).toBeInTheDocument();
+      expect(screen.queryByText('neato burrito')).toBeInTheDocument();
```
cbacon91, line 104: nitpick

```suggestion
      expect(screen.getByText('cool dude')).toBeInTheDocument();
      expect(screen.getByText('neato burrito')).toBeInTheDocument();
```

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+  });
+
+  describe('forwardRef API', () => {
+    it('can be used to get value as array of selected id', async () => {
+      const ref = createRef();
+      render(<Autocompleter {...requiredProps} value={{ name: 'neat', id: 11 }} ref={ref} />);
+
+      expect(ref.current.value).toStrictEqual({ name: 'neat', id: 11 });
+    });
+  });
```
cbacon91, line 115: I really like this spec.

It makes it super clear _why_ using the ref is useful, and I think refs are opaque and odd enough that not everyone in the know would necessarily get it, so this spec is helpful

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+      expect(screen.getByText('Buzz')).toBeInTheDocument();
+    });
+  });
+
+  describe('search', () => {
+    it('searches with the search string', async () => {
+      render(<Autocompleter {...requiredProps} />);
+
+      userEvent.click(screen.getByLabelText('Test label'));
+      userEvent.type(document.activeElement, 'Ba');
```
cbacon91, line 133: I wonder if there is room (in this repo or maybe in user-event public repo) to type or interact with the focused element by default.

I always find it odd that for `type` and `blur` and some other events I have to pass the element in, I would expect the focused element to be the default.

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+      expect(await screen.findByText('Bax')).toBeInTheDocument();
+      expect(screen.queryByText('Baz')).toBeInTheDocument();
```
cbacon91, line 136: nitpick
```suggestion
      expect(screen.getByText('Baz')).toBeInTheDocument();
````

File: src/components/select/select.jsx [hunk 13]
```diff
@@ -177,6 +177,7 @@ const Select = forwardRef(function Select(props, ref) {
         onChange={onSearchChange}
         onClick={onClick}
         onKeyDown={onKeyDown}
+        onKeyUp={props.onKeyUp}
```
cbacon91, line 180: I guess it's because we want to adhere to an allow-list, but this is part of why I think I would generally just spread remaining props on the native element, we're going to keep finding edge cases where a native element supports something that our wrapper doesn't.

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,17 @@
...
+### Default examples
+```js
+import Autocompleter from './autocompleter.jsx';
+<Autocompleter apiEndpoint={'/models'} />
```
cbacon91, line 10: It might be good to show a real-ish example here like `<Autocompleter apiEndpoint="workspaces" />`. I think models is pretty clear but using a model that Mavenlink actually has could cement the idea.

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+  useEffect(() => {
+    fetchModels();
+  }, []);
+
+  function fetchModels(searchString) {
+    fetch(apiEndpoint(searchString))
+      .then((results) => {
+        setModels(results.json.results.map(resultInfo => results.json[resultInfo.key][resultInfo.id]));
+      }).catch((error) => {
+        if (error.error && error.error.type !== 'aborted') {
```
cbacon91, line 23: Do we have aborting requests set up? Should we?

I think in bigmaven it's easy to do, but I wonder if it makes sense for that to be part of the design system component

cbacon91 (approved):
LGTM! This is slick!

Feel free to completely ignore the nitpicks, they're purely stylistic opinions, I don't have any blocking feedback.

src/components/autocompleter/autocompleter.jsx
bennapp: This is an excellent call. Let's take it one step further and make it so the autocompleter-user can pass in a function or a key. And then let's have this nasty logic `modelInfo.title || modelInfo.name || modelInfo.full_name` live in our feature in bigmaven.

Then, we / someone doesn't have to make an mds pr and do the whole song and dance when we want to improve / add to this list.

src/components/autocompleter/autocompleter.md
bennapp: lol

src/components/autocompleter/autocompleter.md
bennapp: I kiiiiinda like the abstract example. My intent was to show that it works for anything. Is that more clear with a specific example 🤷‍♂️

src/components/autocompleter/autocompleter.test.jsx
bennapp: IDK, I copy and pasted it 😞

src/components/autocompleter/autocompleter.test.jsx
bennapp: I think we prefer `toBeInTheDocument` here just for the sake of affordance. But @juanca and @Bloodyaugust probably know better than I.

src/components/autocompleter/autocompleter.md
cbacon91: Works for me! I think it's nit-picky feedback anyway

src/components/select/select.jsx
bennapp: Having to create an MDS PR anytime you want to add something to the native element, is enough of a hassle to make someone think: 'ya, I don't want to use this'. We should probably reconsider this at a high level.

src/components/autocompleter/autocompleter.jsx
juanca: That's a good point. We have abort requests when the component unmounts. But, AFAIK, we need a concept for aborting requests on consecutive interactions. In this case, we need to abort (and probably debounce) any requests when making a new one while the user types.

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+          throw error;
+        }
+      });
+  }
+
+  function apiEndpoint(searchString) {
```
juanca, line 27: We might just want to consider using the URL builder like https://developer.mozilla.org/en-US/docs/Web/API/URL

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
+The `Autocompleter` component is a wrapper for Select component with the additional functionality searching for a specific models from an API endpoint.
+
+`apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
+`value` and `models` shape is expected to be an object(s) with an `id` key, and a display attribute that is one of the following: `title`, `name`, `full_name`, `currency`;
+
+For a11y and additional prop info, see the `Select` component.
+### Default examples
+```js
+import Autocompleter from './autocompleter.jsx';
+<Autocompleter apiEndpoint={'/models'} id='ex-1' name='default-ex' label='Default Example' />
```
juanca, line 10:
```suggestion
<Autocompleter apiEndpoint='/models' id='ex-1' name='default-ex' label='Default Example' />
```

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+### Initial value examples
+```js
+import Autocompleter from './autocompleter.jsx';
+<Autocompleter apiEndpoint={'/models'} value={{id: 11, name: 'cool'}} id='ex-2' name='value-ex' label='Value Example' />
```
juanca, line 16:
```suggestion
<Autocompleter apiEndpoint='/models' value={{id: 11, name: 'cool'}} id='ex-2' name='value-ex' label='Value Example' />
```

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
...
+## Ref API Example
+```jsx
+import Autocompleter from './autocompleter.jsx';
+import RefExample from '@mavenlink/design-system/src/components/ref-example/ref-example.jsx';
+
+const ref = React.createRef();
+
+<RefExample ref={ref}>
+  {({ onChange }) => (
+    <Autocompleter ref={ref} onChange={onChange} apiEndpoint={'/models'} id='ex-3' name='ref-ex' label='Ref Example'  />
```
juanca, line 28:
```suggestion
    <Autocompleter ref={ref} onChange={onChange} apiEndpoint='/models' id='ex-3' name='ref-ex' label='Ref Example'  />
```

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  searchParam: PropTypes.string,
+  models: PropTypes.arrayOf(PropTypes.object),
```
juanca, line 79: I would recommend trying to describe the shape of the data. It seems you document that it requires an `id`? That could look like:

```suggestion
  models: PropTypes.arrayOf(PropTypes.shape({ 
    id: PropTypes.string.required, 
  })),
```

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+describe('src/components/autocompleter/autocompleter', () => {
+  beforeEach(() => {
+    jestServer.use(...mockHandlers());
+  });
+
+  const requiredProps = {
+    id: 'test-id',
+    label: 'Test label',
+    name: 'field-id',
```
juanca, line 18: We might want to update this copy-pasta. This can be confusing since the `name` is not an ID. What about:
```suggestion
    name: 'field_name',
```

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  function apiEndpoint(searchString) {
+    if (searchString) {
+      return `${API_ROOT}${props.apiEndpoint}?${props.searchParam}=${searchString}`;
+    }
+
+    return `${API_ROOT}${props.apiEndpoint}`;
+  }
+
+  function onKeyUp(event) {
```
juanca, line 35: We might want to consider using `onInput` instead.

> Note: If you're looking for a way to react to changes in an input's value, you should use the input event. Some changes are not detectable by keyup, for example pasting text from the context menu in a text input.

-- https://developer.mozilla.org/en-US/docs/Web/API/Document/keyup_event

We don't want to search when a user is just pressing modifier keys, as an example.

src/components/autocompleter/autocompleter.jsx
juanca: I think for request aborting, we can do it in a new PR. That way, we keep this PR focused on a basic autocompleter and we can introduce better functionality as iterations. 

What do y'all think?

src/components/autocompleter/autocompleter.test.jsx
juanca: Yeah.... I blame it on staying agile while writing tests. 😅  

I do have plans for making these copy-pasta tests as a factory of sorts.

Then we can make a new component with tests:

```
describe('foo', () => {
  apiTests({
    className: ['unique-class', () => expect(container.firstChild).toHaveClass('unique-class'),
    ...,
    value: ...,
  });
  
  it('has some unique behavior', () => {
    ...normal looking tests...
  });
});
```

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+  searchParam: PropTypes.string,
+  models: PropTypes.arrayOf(PropTypes.object),
+  onChange: PropTypes.func,
+  displayValueEvaluator: PropTypes.func,
+  label: PropTypes.string,
+  required: PropTypes.bool,
+  readOnly: PropTypes.bool,
+  placeholder: PropTypes.string,
+  className: PropTypes.string,
+  errorText: PropTypes.string,
```
juanca, line 87: Do you mind re-writing this as `validationMessage`? I'm trying to get all our (new) components to align on that prop.

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+  describe('className API', () => {
+    it('prioritizes className prop', () => {
+      const { container } = render(<Autocompleter {...requiredProps} className="prioritize-me" />);
+      expect(container.firstChild).toHaveClass('prioritize-me');
+    });
+  });
+
+  describe('errorText', () => {
+    it('sets the input to be invalid', () => {
```
juanca, line 37: Let's add more assertions:

```
expect(screen.getByLabelText('Test label')).toHaveDescription('not valid');
```

src/components/autocompleter/autocompleter.test.jsx
juanca: Yeah, `toBeInTheDocument()` is preferred. 

Technically, `toBeInTheDocument()` matcher checks whether it's actually in the document. This can be useful when you find an element, do some user interaction, and test that it is _still_ in the document or maybe that it was yeeted.

juanca: But maybe we can improve this test. We really want to assert that the label has a control.

So maybe:

```suggestion
      expect(screen.getByText('Bar').control).toBe(document.getById('test-id'));
```

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    it('unsets the readonly attribute', () => {
+      render(<Autocompleter {...requiredProps} readOnly={false} />);
+      expect(screen.getByLabelText('Test label')).not.toHaveAttribute('readonly', '');
+    });
+  });
+
+  describe('required API', () => {
+    it('sets the required attribute', () => {
```
juanca, line 77: Let's add an assertion:

```
expect(screen.getByText('(Required)')).toBeInTheDocument();
```

```
expect(screen.getByText('(Required)')).not.toBeInTheDocument();
```

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+      expect(screen.getByLabelText('Test label')).toBeRequired();
+    });
+
+    it('unsets the required attribute', () => {
+      render(<Autocompleter {...requiredProps} required={false} />);
+      expect(screen.getByLabelText('Test label')).not.toBeRequired();
+    });
+  });
+
+  describe('value API', () => {
```
juanca, line 88: We want to add a test about re-rendering:

```js
it('accepts a value', async () => {
      const value1 = { id: 22, name: 'cool dude' };
      const value2 = { id: 68, name: 'coolio dude' };
      const { rerender } = render(<Autocompleter {...requiredProps} value={value1} />);
      expect(screen.getByLabelText('Test label')).toHaveValue('cool dude');
      rerender(<Autocompleter {...requiredProps} value={value2} />);
      expect(screen.getByLabelText('Test label')).toHaveValue('coolio dude');
    });
```

src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+    it('accepts a value', async () => {
+      const value = { id: 22, name: 'cool dude' };
+      render(<Autocompleter {...requiredProps} value={value} />);
+
+      expect(screen.getByLabelText('Test label')).toHaveValue('cool dude');
+    });
+  });
+
+  describe('models API', () => {
```
juanca, line 97: Do we want to test the empty case too?

src/components/autocompleter/autocompleter.test.jsx
juanca: You just have to go lower level: https://github.com/testing-library/user-event#keyboardtext-options

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  searchParam: PropTypes.string,
```
juanca, line 78: We're missing test coverage

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+});
+
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
```
juanca, line 77: We're missing test coverage

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+    </Select>
+  );
+});
+
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
+Autocompleter.propTypes = {
+  apiEndpoint: PropTypes.string.isRequired,
```
juanca, line 75: We're missing test coverage

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  searchParam: PropTypes.string,
+  models: PropTypes.arrayOf(PropTypes.object),
+  onChange: PropTypes.func,
```
juanca, line 80: We're missing test coverage

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+  displayValueEvaluator: PropTypes.func,
```
juanca, line 81: We're missing test coverage

src/components/autocompleter/autocompleter.test.jsx
juanca: Also, because we capture the prop as internal state, we also need to add a `rerender` test. 

```
const [models, setModels] = useState(props.models);
```

As it stands, it only accepts the prop once and if a parent component re-renders with new models then it will be thrown away. This is usually not expected.

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,105 @@
...
+const Autocompleter = forwardRef(function Autocompleter(props, ref) {
+  const { execute: fetch } = useFetch();
+  const [models, setModels] = useState(props.models);
+
+  useEffect(fetchModels, []);
+
+  function fetchModels(searchString) {
+    fetch(apiEndpoint(searchString))
+      .then((results) => {
+        setModels(results.json.results.map(resultInfo => results.json[resultInfo.key][resultInfo.id]));
```
juanca, line 19: Reading this is pretty hard.

```suggestion
        setModels(respObj.json.results.map(result => respObj.json[result.key][result.id]));
```

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
+The `Autocompleter` component is a wrapper for Select component with the additional functionality searching for a specific models from an API endpoint.
+
+`apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
```
juanca, line 3: This can be inline documentation so it shows up in the props table. e.g. https://github.com/mavenlink/design-system/blob/ac91df1426e45289eaf10bb985ec8359bc6b5e57/src/components/date/date.jsx#L238

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
...
+`value` and `models` shape is expected to be an object(s) with an `id` key, and a display attribute that is one of the following: `title`, `name`, `full_name`, `currency`;
```
juanca, line 4: If you describe the prop type in `Autocompleter.propTypes` then this would be auto-documented in the props table. 

I think you'll still need to add some text for the display attribute. But maybe it's possible to describe that as prop-types too. I'm not sure though.

src/components/autocompleter/autocompleter.md
```diff
@@ -0,0 +1,31 @@
...
+
+For a11y and additional prop info, see the `Select` component.
```
juanca, line 6:
```suggestion
For a11y and additional prop info, see the `Select` component.

```

src/components/autocompleter/mock-handlers.js
```diff
@@ -0,0 +1,71 @@
+import { rest } from 'msw';
+import mockConstants from '../../mocks/mock-constants.js';
+
+const { API_ROOT } = mockConstants;
+
+const models = [
+  {
+    id: '0',
```
juanca, line 8: I'd recommend using random IDs. We used to have a bug that assumed the response object for the models was an array instead of a lookup. 

I guess at the very least, don't start with `0`. That makes it look very "array-y"

src/components/select/select.jsx
juanca: We moved away from spreading the props because it would lead to usages of components that we are not intending to support. It makes it harder to stay agile when we need to consider esoteric usages in bigmaven.

juanca (changes requested):
Let's use the `onInput` event. Let's add missing test coverage. Everything else looks pretty good! I really like how this new component is efficiently re-using the Select component as its presentational layer. I kinda wish there was a bit more composition but I guess Select just does a lot of it for us, huh?

src/components/select/select.jsx
Bloodyaugust: Yeah, I think our reasoning for moving away from this still applies. We want to provide a focused interface to interact with with these components, reduce cognitive load. And even if you do still need to build your own to support some attribute we don't normally (less applicable with something as complex as `Select`, but still), you can still compose in our styles.

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,106 @@
...
+import React, { forwardRef, useEffect, useState } from 'react';
+import useFetch from '@bloodyaugust/use-fetch';
+import Select from '../select/select.jsx';
+import ListOption from '../list-option/list-option.jsx';
+import mockConstants from '../../mocks/mock-constants.js';
+
+const { API_ROOT } = mockConstants;
+
+const Autocompleter = forwardRef(function Autocompleter(props, ref) {
+  const { execute: fetch } = useFetch();
```
Bloodyaugust, line 11: The original reasoning behind the exported method being called `execute` instead of `fetch` was to avoid collision with global, native window objects. We've since decided that calls to window globals should be explicit, IE `window.fetch`. If we find ourselves renaming this to `fetch` frequently, it might be worth changing the name in the hook library itself.

src/components/autocompleter/autocompleter.jsx
Bloodyaugust: We can totally expose a method to enable this behavior in `useFetch`. It's not there because we didn't need it yet, but if we have a need now... No time like the present!

src/components/autocompleter/autocompleter.jsx
Bloodyaugust: If we're going to do this, we should probably make it standard going forward, and backfill lots of stuff.

src/components/autocompleter/autocompleter.test.jsx
Bloodyaugust: @cbacon91 FWIW I think
> If that's the case, should we make sure a class name isn't there?

is sufficiently covered by the snapshot.

src/components/autocompleter/autocompleter.test.jsx
Bloodyaugust: I think this is just one of those things that becomes strange when you're writing an API to mimic user interaction, but for automated and code-driven tests. Personally, I like the explicit-ness of knowing what I'm "typing" into.

src/components/autocompleter/mock-handlers.js
Bloodyaugust: Probably another thing we should standardize on moving forward. Also seems like it would be made moot in the bright future of importing brainstem factories.

src/components/autocompleter/autocompleter.jsx
bennapp: This is covered. Given we use mock service worker to fetch the request.

src/components/autocompleter/autocompleter.test.jsx
bennapp: Most of our simple prop test do not seem worth while to me.

src/components/autocompleter/autocompleter.test.jsx
bennapp: Is this testing React-core, is there any reason to think this wouldn't work?

bennapp:
@juanca This is ready for re-review when you get a chance

bennapp:
A quick note on the native JS url helper / builder. With the way we have set up mock service worker to assume some base portions of the url, I don't think we will be able to use the native builder from the context of our components.

If that's something we really care about, we should make that refactor to the mock service setup, but I don't think that change should be in this PR.

src/components/autocompleter/autocompleter.jsx
juanca: The test coverage fails if (highly unlikely) `apiEndpoint` is hard-coded to `/models`. Usually, even for required props, we have at least 1 test that sets it to something else that is unique. e.g. Labels are required and most tests set it as `Test label` but we still have another test that sets it to `Unique label`.

src/components/autocompleter/autocompleter.jsx
```diff
@@ -0,0 +1,122 @@
...
+    </Select>
+  );
+});
+
+function displayName(modelInfo) {
+  return modelInfo.title || modelInfo.name || modelInfo.full_name || modelInfo.currency;
+}
+
+Autocompleter.propTypes = {
+  // `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`.
```
juanca, line 85: If we want this to be on the documentation page, you need to use styleguidist comments:

```suggestion
  /** `apiEndpoint` should be the route of the api's endpoint (excluding the base api), eg. `/workspaces`. */
```

Same for the rest of these comments in the prop-types interface.

src/components/select/select.jsx
```diff
@@ -224,13 +228,17 @@ Select.propTypes = {
   id: PropTypes.string.isRequired,
   children: PropTypes.node,
   className: PropTypes.string,
-  /** Function is passed `value`, default returns value without modification, should always return a `string`. You *should* set this if your `value` is not of type `string`. */
-  displayValueEvaluator: PropTypes.func,
+  /** Function is passed `value`, default returns value without modification, should always return a `string`. You *should* set this if your `value` is not of type `string`. Pass in `false` to prevent filtering. */
```
juanca, line 231: 💯

src/components/select/select.jsx
```diff
@@ -224,13 +228,17 @@ Select.propTypes = {
...
+  displayValueEvaluator: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.bool,
+  ]),
   errorText: PropTypes.string,
   label: PropTypes.string.isRequired,
   listOptionRefs: PropTypes.arrayOf(ListOptionRefType).isRequired,
   name: PropTypes.string.isRequired,
   onChange: PropTypes.func,
+  onInput: PropTypes.func,
```
juanca, line 241: We're missing this test coverage on Select 😅

juanca (commented):
Non-blocking, nice to haves.

PR #299, opened 2021-04-26
Add auto-flush to popover component

cbacon91:
https://www.pivotaltracker.com/story/show/177860915

## PR upkeep checklist

- [x] Change log entry
- [ ] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check


src/components/popover/popover.jsx
```diff
@@ -56,14 +72,16 @@ const Popover = forwardRef(function Popover(props, ref) {
     },
   }));
 
+  const className = useMemo(() => `${styles.container} ${styles[flush]}`, [flush]);
```
juanca, line 75: A couple of notes:

- RE: useMemo: A tiny bit overkill tbh. Memoization should be on heavy computations IMO. 
- RE: interpolated className: I would prefer to see CSS composition. But I'm not sure what is the best approach in this case. 

Another thought: for stateful, 1-off "single-line" styling, I think it's okay to use the `styles` prop too. We have used this to pass state as a CSS variable from React components to CSS modules. 

e.g. How many icons are we going to show? Instead of making n-classes (`.one-icon`, `.two-icons`, ...). We set a var (`styles={{--num-of-icons: 3}}`) and use it in the styling rule (`padding-right: var(--num-of-icons, 1);`).

src/components/popover/popover.jsx
```diff
@@ -86,14 +104,18 @@ const Popover = forwardRef(function Popover(props, ref) {
 });
 
 Popover.propTypes = {
+  autoflush: PropTypes.bool,
```
juanca, line 107: We need tests for this prop.

src/components/popover/popover.jsx
```diff
@@ -86,14 +104,18 @@ const Popover = forwardRef(function Popover(props, ref) {
...
   children: PropTypes.node,
+  flush: PropTypes.oneOf(['left', 'right']),
```
juanca, line 109: For these kinds of props, we can to be exhaustive in our test suite. We need one for setting it to `left` and another for `right`. 

We skip approximately large prop values like `string` or `number` in favor of a single or few good cases.

juanca (changes requested):
A bit of test coverage is missing. 

I'm not sure about how to deal with the CSS concatenation so we can skip.

src/components/popover/popover.jsx
cbacon91: Mentioned on coscreen but for posterity:

I ran into struggling how to test this because `getBoundingClientRect` always returns 0's (`{x: 0, y: 0, left: 0, ... }`) in jsdom. Some people have gotten around this by overwriting `Object.prototype.getBoundingClientRect` but that felt super wrong to do. 

I'm going to take a stab at mocking it out on the individual section ref. Another idea is to extract this as a hook and test that hook instead, but that doesn't really guarantee that this component works and it's also not a super reusable hook to begin with.

src/components/popover/popover.jsx
cbacon91: Yep, good call! - Good call! e631eac6

src/components/popover/popover.jsx
cbacon91: I used the style api instead - b01eceb9

I didn't pass a variable to the css, instead I just set style directly on the element. I wasn't sure how to choose which css property to change using the variable method. I _think_ you could just set the `left` between 0 and the element width, but I think one thing this leaves us open to is maybe one day flushing top or bottom if that makes sense.

 Let me know what you think

src/components/popover/popover.jsx
cbacon91: 98c0bdc2

Looks like stubbing the element's function worked, just takes a little finagling.

juanca (approved):

PR #300, opened 2021-04-28
Create checkbox

cbacon91:
https://www.pivotaltracker.com/story/show/177913789

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/create-checkbox/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/checkbox/checkbox.css
+++ b/src/components/checkbox/checkbox.css
@@ -1,10 +1,8 @@
-@import '../../styles/colors-v2.css';
 @import "../../styles/spacing.css";
-@import "../../styles/typography-v2.css";
 
 .input {
-  height: 20px;
-  width: 20px;
+  height: var(--spacing-x-large);
+  display: flex;
 }
 
 .invalid-icon {
--- a/src/components/checkbox/checkbox.css
+++ b/src/components/checkbox/checkbox.css
@@ -1,0 +1,12 @@
+@import '../../styles/colors-v2.css';
+@import "../../styles/spacing.css";
+@import "../../styles/typography-v2.css";
+
+.input {
+  height: 20px;
+  width: 20px;
+}
+
+.invalid-icon {
+  composes: invalid-icon from '../input/input.css';
+}
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -57,6 +57,11 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
   useImperativeHandle(ref, () => ({
     id: props.id,
     name: props.name,
+    get value() {
+      if(props.value) return props.value;
+
+      return inputRef.current.checked ? 'on' : 'off';
+    },
     get checked() {
       return inputRef.current.checked;
     },
@@ -113,6 +118,7 @@ Checkbox.propTypes = {
   readOnly: PropTypes.bool,
   required: PropTypes.bool,
   validationMessage: PropTypes.string,
+  value: PropTypes.string,
 }
 
 Checkbox.defaultProps = {
@@ -126,6 +132,7 @@ Checkbox.defaultProps = {
   readOnly: undefined,
   required: undefined,
   validationMessage: '',
+  value: undefined,
 }
 
 
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -15,6 +15,9 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
         className={props.className}
         id={props.id}
         name={props.name}
+        onBlur={props.onBlur}
+        onChange={props.onChange}
+        onFocus={props.onFocus}
         ref={ref}
         type="checkbox"
       />
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -18,6 +18,7 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
         onBlur={props.onBlur}
         onChange={props.onChange}
         onFocus={props.onFocus}
+        readOnly={props.readOnly}
         ref={ref}
         type="checkbox"
       />
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -68,6 +68,7 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
 
   return (
     <FormControl
+      className={props.cssContainer}
       error={validationMessage}
       id={props.id}
       label={props.label}
@@ -100,7 +101,9 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
 });
 
 Checkbox.propTypes = {
+  checked: PropTypes.bool,
   className: PropTypes.string,
+  cssContainer: PropTypes.string,
   id: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
   name: PropTypes.string,
@@ -110,11 +113,12 @@ Checkbox.propTypes = {
   readOnly: PropTypes.bool,
   required: PropTypes.bool,
   validationMessage: PropTypes.string,
-  checked: PropTypes.bool,
 }
 
 Checkbox.defaultProps = {
+  checked: undefined,
   className: undefined,
+  cssContainer: undefined,
   name: undefined,
   onBlur: () => {},
   onChange: () => {},
@@ -122,7 +126,6 @@ Checkbox.defaultProps = {
   readOnly: undefined,
   required: undefined,
   validationMessage: '',
-  checked: undefined,
 }
 
 
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -14,6 +14,7 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
       <input
         className={props.className}
         id={props.id}
+        name={props.name}
         ref={ref}
         type="checkbox"
       />
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -1,10 +1,30 @@
 /* eslint-disable react/prop-types */
-import React, { forwardRef, useRef } from 'react';
+import React, { forwardRef, useEffect, useImperativeHandle, useRef } from 'react';
 import FormControl from '../form-control/form-control.jsx';
+import useMounted from '../../hooks/use-mounted';
 
 const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
   const fallbackRef = useRef();
   const ref = forwardedRef || fallbackRef;
+  const inputRef = useRef();
+  const mounted = useMounted();
+
+  useEffect(() => {
+    if (!mounted.current) return;
+
+    inputRef.current.checked = props.checked ?? false;
+  }, [props.checked]);
+
+  useImperativeHandle(ref, () => ({
+    id: props.id,
+    name: props.name,
+    get checked() {
+      return inputRef.current.checked;
+    },
+    get dirty() {
+      return false;
+    },
+  }));
 
   return (
     <FormControl
@@ -12,6 +32,7 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
       label={props.label}
     >
       <input
+        aria-describedby={`${props.id}Hint`}
         className={props.className}
         id={props.id}
         name={props.name}
@@ -19,7 +40,7 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
         onChange={props.onChange}
         onFocus={props.onFocus}
         readOnly={props.readOnly}
-        ref={ref}
+        ref={inputRef}
         type="checkbox"
       />
     </FormControl>
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -10,13 +10,8 @@ import PropTypes from 'prop-types';
 import FormControl from '../form-control/form-control.jsx';
 import useMounted from '../../hooks/use-mounted.js';
 import Icon from '../icon/icon.jsx';
-import styles from '../input/input.css';
+import styles from './checkbox.css';
 import cautionSvg from '../../svgs/caution.svg';
-
-function getClassName(className, validationMessage) {
-  if (className) return className;
-  return validationMessage ? styles['invalid-input'] : styles.input;
-}
 
 const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
   const fallbackRef = useRef();
@@ -56,6 +51,8 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
     },
   }));
 
+  const className = props.className || styles.input;
+
   function onBlur(event) {
     inputRef.current.setCustomValidity('');
     setValidationMessage(inputRef.current.validationMessage);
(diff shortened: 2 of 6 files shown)


src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+import PropTypes from 'prop-types';
+import FormControl from '../form-control/form-control.jsx';
+import useMounted from '../../hooks/use-mounted.js';
+import Icon from '../icon/icon.jsx';
+import styles from './checkbox.css';
+import cautionSvg from '../../svgs/caution.svg';
+
+const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
+  const fallbackRef = useRef();
+  const ref = forwardedRef || fallbackRef;
```
cbacon91, line 18: I copied this pattern from other components, but I think this can be shortened to just `const ref = useRef(forwardedRef && forwardedRef.current)`?

src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+    inputRef.current.checked = props.checked;
+  }, [props.checked]);
+
+  useEffect(() => {
+    setValidationMessage(props.validationMessage);
+  }, [props.validationMessage]);
+
+  useLayoutEffect(() => {
+    inputRef.current.setCustomValidity(validationMessage);
+  }, [validationMessage]);
```
cbacon91, line 35: All of the validation logic (including the `onBlur` and `onChange` events) are pretty similar to what was copied from the `Date` component.

I started with the `useValidation` hook, but since `value` is `on`/`off` instead of falsey or truthy, the internals didn't work.

I spent some time on an extracted version of _these_ events as a hook and planned on backfilling it to date, but it seemed like having two `validation` hooks felt bad / weird. The commit is here (ed6f7b0ba2ae309d5831f8dd2b915aa2faa472eb) but I reverted it.

bennapp:
This looks great to me! But let's have @juanca give the final review / sign off here.

src/components/checkbox/checkbox.md
```diff
@@ -0,0 +1,45 @@
+For design and accessibility functionality, see [Form Controls > Checkbox documentation](dead-link-talk-to-juanca).
```
cbacon91, line 1: blocking until I put this link in place

src/components/checkbox/checkbox.jsx
juanca: React hooks cache their initialization. If we were to `useRef(forwardedRef && forwardedRef.current)` then it needs a `useEffect(() => ref.current = forwardedRef.current, [forwardedRef.current])`. But changes in refs don't propagate React renders... so there is a high likely hood that it will be stale.

src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+import cautionSvg from '../../svgs/caution.svg';
+
+const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
+  const fallbackRef = useRef();
+  const ref = forwardedRef || fallbackRef;
+  const inputRef = useRef();
+  const mounted = useMounted();
+  const [validationMessage, setValidationMessage] = useState(props.validationMessage);
+
+  useEffect(() => {
```
juanca, line 23: Since this hook relies on a _rendered_ DOM node, we should switch this to

```suggestion
  useLayoutEffect(() => {
```

juanca: And remove the mounted check since we know we have a DOM node.

src/components/checkbox/checkbox.jsx
juanca: Yeah, I wanted to see how the new validation spec holds up in a few implementations before committing to redoing the validation hook.

src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+    get checked() {
+      return inputRef.current.checked;
+    },
+    get dirty() {
+      const providedValue = props.checked || false;
+      return providedValue !== this.checked;
+    },
+  }));
+
+  const className = props.className || styles.input;
```
juanca, line 54: Please use default props instead of the logical assignment.

src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+  function onBlur(event) {
+    inputRef.current.setCustomValidity('');
+    setValidationMessage(inputRef.current.validationMessage);
+    props.onBlur(event);
+  }
+
+  function onChange(event) {
+    inputRef.current.setCustomValidity('');
```
juanca, line 63: We don't want validations on change event. Only blur. This is the new validation hook.

src/components/checkbox/checkbox.md
juanca, line 1:
```suggestion
For design and accessibility functionality, see [Form Controls > Checkbox documentation](https://www.notion.so/Checkbox-2c117a30d5c14e50a171a263e04b4fd9).
```

src/components/checkbox/checkbox.css
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/colors-v2.css';
+@import "../../styles/spacing.css";
+@import "../../styles/typography-v2.css";
+
+.input {
+  height: 20px;
```
juanca, line 6: We want the input or its container's height to be `height: var(--spacing-x-large);`. I am digging the thicker checkbox but I am going to guess design won't dig it?

Instead, let's make its container 32px, leave native sizing of the checkbox, and align center. Something like:

image

I talked with Anne about it and we didn't come to a final design for a checkbox field. We'll probably iterate a bit after the fact.

src/components/checkbox/checkbox.css
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/colors-v2.css';
```
juanca, line 1: We aren't using any of these CSS variables in this file. Let's remove unused imports.

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      const ref = createRef();
+      render(<Checkbox {...requiredProps} name="unique-name" ref={ref} />);
+      expect(ref.current.name).toBe('unique-name');
+    });
+  });
+
+  it('sets the onblur handler', () => {
+    const onBlur = jest.fn();
+    render(<Checkbox {...requiredProps} onBlur={onBlur} />);
+    fireEvent.blur(screen.getByRole('checkbox', { name }));
```
juanca, line 71: We don't want to have test coverage with `fireEvent`. Instead always prefer `userEvent` -- our tests should try to reflect user behavior whenever possible.

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      userEvent.click(screen.getByRole('checkbox', { name }));
+      expect(onChange.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('onFocus API', () => {
+    it('sets the onFocus handler', () => {
+      const onFocus = jest.fn();
+      render(<Checkbox {...requiredProps} onFocus={onFocus} />);
+      fireEvent.focus(screen.getByRole('checkbox', { name }));
```
juanca, line 88: We don't want to have test coverage with `fireEvent`. Instead always prefer `userEvent` -- our tests should try to reflect user behavior whenever possible.

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      expect(screen.getByRole('checkbox', { name })).toBeValid();
+    });
+
+    it('is invalid on mount but does not have an error message', () => {
+      render(<Checkbox {...requiredProps} required />);
+      expect(screen.getByRole('checkbox', { name })).toBeInvalid();
+      expect(screen.getByRole('checkbox', { name })).toHaveDescription('');
+    });
+
+    it('is invalid after deselecting', () => {
```
juanca, line 209: We should only be validating "on blur"

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+  describe('onFocus API', () => {
+    it('sets the onFocus handler', () => {
+      const onFocus = jest.fn();
+      render(<Checkbox {...requiredProps} onFocus={onFocus} />);
+      fireEvent.focus(screen.getByRole('checkbox', { name }));
+      expect(onFocus.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('readonly api', () => {
```
juanca, line 93: Since readOnly is boolean, please be exhaustive in the test coverage.

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      userEvent.click(screen.getByRole('checkbox', { name }));
+      expect(ref.current.dirty).toBe(true);
+
+      rerender(<Checkbox {...requiredProps} ref={ref} checked={false} />);
+      expect(ref.current.dirty).toBe(false);
+    });
+  });
+
+  describe('required api', () => {
+    it('can be set', () => {
```
juanca, line 191: Do we want to add an assertion for the `(Required)` text as well?

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+    });
+
+    it('is set on the ref', () => {
+      const ref = createRef();
+      render(<Checkbox {...requiredProps} name="unique-name" ref={ref} />);
+      expect(ref.current.name).toBe('unique-name');
+    });
+  });
+
+  it('sets the onblur handler', () => {
```
juanca, line 68: We're missing a describe block for this prop

src/components/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,132 @@
...
+  cssContainer: PropTypes.string,
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  name: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  validationMessage: PropTypes.string,
```
juanca, line 114: We're missing test coverage for this.

juanca (changes requested):
Looking good. A few revisions:

- Missing test coverage
- New validation spec (no onchange please)
- Better test patterns: userEvent vs fireEvent
- Better prop patterns: defaultProps vs the other thing
- Better hook patterns: useLayoutEffect vs useEffect
- Notion link!
- - etc?

src/components/checkbox/checkbox.test.jsx
cbacon91: Yeah, I know. The issue I ran into is that 

1. I copied this from `input` specs so I wasn't sure if this was `fireEvent` for a reason, and 
2. I don't know how to test `blur` with userEvent. Is there some variety of `userEvent.tab()` or something?

src/components/checkbox/checkbox.test.jsx
cbacon91: Yeah, but I don't really know how to focus on something using userEvent. How would you do that?

src/components/checkbox/checkbox.test.jsx
cbacon91: My reasoning was that since the snapshot doesn't have the boolean value, adding the opposite case felt redundant.

I'm not 100% I grok the use case for snapshot testing when we also double-check the same values.

src/components/checkbox/checkbox.jsx
cbacon91: 4e2f968353519f3dcbc9eec5298d64aa8c0ee459

src/components/checkbox/checkbox.jsx
cbacon91: d33d055d784c4fd2e19e67819f23de989b8bc394

src/components/checkbox/checkbox.jsx
cbacon91: 5fa726efb5484c8688c48a615076de349af0d5c1

src/components/checkbox/checkbox.css
cbacon91: db5d72ea5b088457688e7001e039823573617265

I didn't add the align centering because it didn't seem needed, but the container is the right size now.

src/components/checkbox/checkbox.css
cbacon91: D'oh. I copy/pasted css from another file and forgot to remove the imports.

Part of db5d72ea5b088457688e7001e039823573617265

src/components/checkbox/checkbox.test.jsx
cbacon91: 5fa726efb5484c8688c48a615076de349af0d5c1

src/components/checkbox/checkbox.test.jsx
cbacon91: 456d2fb27de3ad8edc3fe2d8c286ecbbf483eaca

I added it (I'd rather have too much testing than not enough), I'm just not sure I understand the value of snapshot testing

src/components/checkbox/checkbox.jsx
cbacon91: f92589672bb4e96a5965b09e62c8ac1eae4c2a8b

src/components/checkbox/checkbox.test.jsx
cbacon91: f92589672bb4e96a5965b09e62c8ac1eae4c2a8b

src/components/checkbox/checkbox.test.jsx
cbacon91: f92589672bb4e96a5965b09e62c8ac1eae4c2a8b

src/components/checkbox/checkbox.test.jsx
juanca: `userEvent.click(the_thing)` ! Or you can tab into it: `userEvent.tab()`

juanca: The idea behind `userEvent` is: "What would a user do?" Think of the actions you need to do to get the component into the desired state. 

In this case, you would say "I would click the label" or "I would tab to it"

src/components/checkbox/checkbox.test.jsx
juanca: Snapshot tests are for defaults.

API tests are for particular component states. 

For example, right now the default is it is _not_ readonly. The markups shows that. That bit comes from `Checkbox.defaultProps.readOnly`. If it were changed to `true` then we update the snapshot. But without the API test coverage, we would _now_ be missing the coverage. And a developer most likely would assume there already exists coverage and all they need to do is double check the new snapshot and voila!

src/components/checkbox/checkbox.test.jsx
```diff
@@ -0,0 +1,248 @@
...
+  describe('onFocus API', () => {
+    it('sets the onFocus handler', () => {
+      const onFocus = jest.fn();
+      render(<Checkbox {...requiredProps} onFocus={onFocus} />);
+      fireEvent.focus(screen.getByRole('checkbox', { name }));
+      expect(onFocus.mock.calls.length).toEqual(1);
+    });
+  });
+
+  describe('readonly api', () => {
```
juanca, line 95:
```suggestion
  describe('readOnly api', () => {
```

juanca (approved):
Nice! 

Hope the snapshot vs API test explanation makes sense.

PR #309, opened 2021-05-11
MDS Popover#shouldClose API, MDS Checkbox#readOnly fix

juanca:
https://www.pivotaltracker.com/story/show/178062061

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/popover-should-close/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/24703

### DIFF
--- a/src/components/checkbox/checkbox.css
+++ b/src/components/checkbox/checkbox.css
@@ -1,4 +1,3 @@
-@import "../../styles/colors-v2.css";
 @import "../../styles/spacing.css";
 
 .input {
--- a/src/components/checkbox/checkbox.css
+++ b/src/components/checkbox/checkbox.css
@@ -1,3 +1,4 @@
+@import "../../styles/colors-v2.css";
 @import "../../styles/spacing.css";
 
 .input {
@@ -7,4 +8,8 @@
 
 .invalid-icon {
   composes: invalid-icon from '../input/input.css';
+}
+
+.input[readonly] {
+  opacity: 0.3;
 }
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -53,6 +53,14 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
     props.onBlur(event);
   }
 
+  function onClick(event) {
+    if (props.readOnly) event.preventDefault();
+  }
+
+  function onKeyDown(event) {
+    if (props.readOnly && event.key === ' ') event.preventDefault();
+  }
+
   return (
     <FormControl
       className={props.cssContainer}
@@ -70,7 +78,9 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
         name={props.name}
         onBlur={onBlur}
         onChange={props.onChange}
+        onClick={onClick}
         onFocus={props.onFocus}
+        onKeyDown={onKeyDown}
         readOnly={props.readOnly}
         ref={inputRef}
         required={props.required}
--- a/src/components/date/date.test.jsx
+++ b/src/components/date/date.test.jsx
@@ -89,7 +89,7 @@ describe('src/components/date/date.test.jsx', () => {
       render(<Date {...requiredProps} />);
       userEvent.type(screen.getByLabelText('Test label'), '');
       expect(screen.getByText(today.calendarHeading)).toBeInTheDocument();
-      userEvent.type(screen.getByLabelText('Test label'), '{esc}', { skipClick: true });
+      userEvent.keyboard('{Escape}');
       expect(screen.queryByText(today.calendarHeading)).not.toBeInTheDocument();
     });
 
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -30,11 +30,11 @@ const Popover = forwardRef(function Popover(props, ref) {
   }
 
   useLayoutEffect(() => {
+    if (!mounted.current) return;
+
     if (open) {
       closeIconRef.current.focus({ preventScroll: true });
-    }
-
-    if (mounted.current && !open) {
+    } else {
       props.onClose();
     }
   }, [open]);
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -68,7 +68,7 @@ const Popover = forwardRef(function Popover(props, ref) {
           active={true}
           className={styles['close-button']}
           icon={iconClear}
-          label={'Close popover'}
+          label="Close popover"
           onPress={() => { setOpen(false); }}
           ref={closeIconRef}
         />
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -23,8 +23,8 @@ const Popover = forwardRef(function Popover(props, ref) {
 
   function onBlur(event) {
     // Target is set to `null` when losing focus to a non-interactive element (e.g. text)
-    const defaultShouldClose = event.relatedTarget === null || !sectionRef.current.contains(event.relatedTarget);
-    if (props.shouldClose(event) && defaultShouldClose) {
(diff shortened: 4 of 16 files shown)


src/components/checkbox/checkbox.jsx
```diff
@@ -53,6 +53,14 @@ const Checkbox = forwardRef(function Checkbox(props, forwardedRef) {
     props.onBlur(event);
   }
 
+  function onClick(event) {
+    if (props.readOnly) event.preventDefault();
+  }
+
+  function onKeyDown(event) {
+    if (props.readOnly && event.key === ' ') event.preventDefault();
+  }
```
cbacon91, line 62: d'oh.

I'm a bit surprised about this if I'm honest, I guess I would have expected the `readonly` on the native `input` to do this already, but it's my bad for missing the test coverage. Good catch!

src/components/checkbox/checkbox.css
```diff
@@ -1,3 +1,4 @@
+@import "../../styles/colors-v2.css";
```
cbacon91, line 1: I don't think this is needed

src/components/date/date.test.jsx
```diff
@@ -89,7 +89,7 @@ describe('src/components/date/date.test.jsx', () => {
       render(<Date {...requiredProps} />);
       userEvent.type(screen.getByLabelText('Test label'), '');
       expect(screen.getByText(today.calendarHeading)).toBeInTheDocument();
-      userEvent.type(screen.getByLabelText('Test label'), '{esc}', { skipClick: true });
```
cbacon91, line 92: Ah, yeah, I think this is what I was talking about in slack - I knew there was a _way_ to do keypresses, but `keyboard` is definitely more intuitive

src/components/popover/popover.jsx
```diff
@@ -93,16 +83,17 @@ Popover.propTypes = {
   children: PropTypes.node,
   flush: PropTypes.oneOf(['left', 'right']),
   onClose: PropTypes.func,
-  startOpen: PropTypes.bool,
+  /** Handles either a FocusEvent or MouseEvent. This is an additional predicate for configuring the closing behavior. */
+  shouldClose: PropTypes.func,
```
cbacon91, line 87: I'm bad at naming, and this is a bit pedantic, but whenever I see questions as an argument to something, I expect it to a boolean. I get why this is a function - its a function _that returns_ a boolean - but I wonder if there is a better name hidden somewhere? Nothing comes to mind for me so maybe not..

src/components/popover/popover.jsx
```diff
@@ -10,43 +10,33 @@ import IconButton from '../icon-button/icon-button.jsx';
...
   const selfRef = ref || backupRef;
   const { flush } = useFlush({ ref: sectionRef, initialDirection: props.flush, autoflush: props.autoflush, open });
+  const mounted = useMounted();
 
-  const onFocusIn = (event) => {
-    if (open && event.target instanceof Node && !sectionRef.current.contains(event.target)) {
+  function onBlur(event) {
+    // Target is set to `null` when losing focus to a non-interactive element (e.g. text)
+    const defaultShouldClose = event.relatedTarget === null || !sectionRef.current.contains(event.relatedTarget);
+    if (props.shouldClose(event) && defaultShouldClose) {
```
cbacon91, line 27: 1. This is pedantic, but this isn't really a `default`, it's just internal logic. `default` to me implies that it can be overridden (like default props). Related, I wonder if we can use that excuse to find a name that explains why and not what, and roll the explanation from the comment into the name of the variable, like `hasFocusShifted` or something along those lines
2. Kinda dumb, but would it make sense to swap the order of these two conditions? `defaultShouldClose` seems like it'll always be fast to evaluate, but since `shouldClose` comes from `props`, it could require a large calculation (depending on the consumer of the design system). We might get random short-circuiting benefits from an order change.

src/components/popover/popover.jsx
```diff
@@ -66,24 +56,24 @@ const Popover = forwardRef(function Popover(props, ref) {
     <section
       aria-labelledby="popover-heading"
       className={styles.container}
+      onBlur={onBlur}
       ref={sectionRef}
       role="dialog"
       style={flush}
+      tabIndex={-1}
```
cbacon91, line 63: This is really confusing me.

`onblur` is something that triggers when focus is lost, but the tabindex of -1 indicates to me it _can't_ be focused, so how dose `onblur` ever get triggered? I feel like I'm getting something mixed up here

Moreover, aren't sections out of the tabbing order already?

src/components/popover/popover.jsx
```diff
@@ -66,24 +56,24 @@ const Popover = forwardRef(function Popover(props, ref) {
...
-          />
-        </div>
-        {props.children}
+      <div className={styles['heading-container']} id="popover-heading">
+        <h1 className={styles.heading}>{props.title}</h1>
+        <IconButton
+          active={true}
+          className={styles['close-button']}
+          icon={iconClear}
+          label={'Close popover'}
```
cbacon91, line 71:
```suggestion
          label="Close popover"
```

src/components/popover/popover.test.jsx
```diff
@@ -106,7 +113,26 @@ describe('Popover', () => {
     expect(screen.getByText('Another title').tagName).toEqual('H1');
   });
 
-  describe('flush api', () => {
+  describe('autoflush API', () => {
+    // We don't know how to test this well. Manually QA for now.
+  });
```
cbacon91, line 118: TODO for me: I want to find a way to test this, I know there is one, I should come back to it

I do like that you added _something_ here for it, I feel like that points out that we intentionally don't have test coverage here instead of making it look like its something we just missed.

cbacon91 (approved):
LGTM, I have a few questions and the tabindex / section bit confuses me but nothing blocking

src/components/checkbox/checkbox.jsx
juanca: I looked it up.... HTML spec defines read-only to only stop modifying the `value` of the element. A checkbox does not modify its value. It modifies its checked state. So native read-only functionality does not apply to checkboxes. 🤦 

Random internet answers suggest having hidden checkboxes and a disabled checkbox. It feels like the wrong solution though? 

I went with this solution in the end. Markup is simple. Implementation details are simple too.

src/components/popover/popover.jsx
juanca: Yeah, I tried looking up best practice around predicates and couldn't find much. 

I figured React-ers might be familiar with the `should*`, `will*`, `did*` name patterns given React class methods. I was hoping for a cooler `use*` approach but didn't come up with anything then .

src/components/popover/popover.jsx
juanca: 1. Yeah, maybe just `shouldClose`? 
2. That's interesting, yeah, I can switch the order.

src/components/popover/popover.jsx
juanca: A tabindex of -1 implies it can be focused with JavaScript and it _not_ part of the page tab sequence. It is needed to populate the `FocusEvent#relatedTarget` property too. 

Without this bit, clicking _in_ the Popover but something non-interactive implies that _nothing_ is focused. The activeElement is the document but the event does not have a relatedTarget (`null`). 

By setting the tab index, the browser thinks the Popover itself gets focused (and is actually the case in Safari). This means that there is a relatedTarget. In addition, once the user tries to focus anything else the `onBlur` is still invoked. 

It's kinda nice because it solves the "did the composite widget lose focus" question really easily. While focus keeps shifting within the component, we know we're focused. Otherwise, we know we lost focus -- regardless if the relatedTarget is not populated.

PR #311, opened 2021-05-13
MDS Multi-Select

Bloodyaugust:
## Motivation
We want a multi-select component.

## Acceptance Criteria
| API | prop | ref |
| --- | --- | --- |
| className(s) | x | |
| dirty | | x |
| label | x | |
| placeholder | x | |
| required | x | |
| readOnly | x | |
| validationMessage | x | |
| value | x | x |

Notion documentation:

- Purposes / spiel
- A11y keyboard
- A11y label

MDS documentation: 

- Examples
- At least 1 ref example 

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-mds-multi-select-177518329/#/Components/MultiSelect
- [ ] ~~Green Bigmaven CI check~~ Not necessary, no uses yet in bigmaven


src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+MultiSelect.propTypes = {
+  classNames: PropTypes.shape({
+    container: PropTypes.string,
+    formControlChildrenContainer: PropTypes.string,
+    input: PropTypes.string,
+    tagList: PropTypes.string,
+  }),
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  listboxChildren: PropTypes.func, // eslint-disable-line react/no-unused-prop-types
```
Bloodyaugust, line 286: Note: These are ignored because the linter doesn't see them being used. Check out `dropdownContents` on line 57.

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,80 @@
...
+
+.icon-clear-readonly {
+  composes: icon-clear;
+  cursor: unset;
+  pointer-events: none;
+}
+
+.icon-container {
+  composes: icons from '../form-control-icons/form-control-icons.css';
+  top: calc(var(--spacing-large) + var(--spacing-small));
```
Bloodyaugust, line 25: Note: This is necessary to keep the icons in the top-right-hand corner of the input.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+    );
+  };
+
+  function getOption(id) {
+    return props.options.find(option => props.optionIDGetter(option) === id);
+  }
+
+  function getVisibleOptions() {
+    return props.options
+      .filter(option => props.optionLabelGetter(option).toLowerCase().includes(autocompleteValue.toLowerCase()))
```
Bloodyaugust, line 108: Note: the use of `toLowerCase` here fixes what I _think_ is a bug/unintended consequence of the `CustomFieldInputMultiChoice` implementation. If we _don't_ want to do a case-insensitive match, I can remove this.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+
+        if (typeof optionID === 'string') {
+          return val.toLowerCase() === optionID.toLowerCase();
+        }
+
+        return val === optionID;
+      }));
+  }
+
+  function onAutocompleteBlur() {
```
Bloodyaugust, line 120: Note: we're not using `useValidate` anymore, so this code is a one-off. We should refactor it after the appropriate changes to `useValidate` are made, or to use whatever replaces it.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+
+  function onOptionRemove(event) {
+    const newValue = value.filter((option, index) => (
+      valueRefs[index].current !== event.target
+    ));
+    setValue(newValue);
+  }
+
+  function onOptionsClear(event) {
+    event.preventDefault();
```
Bloodyaugust, line 149: Note: Do we always want to prevent the icon click from propagating? That was my assumption, but this can be moved inside the `if` block below otherwise.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+    setValue(props.value);
+  }, [props.value.map(val => JSON.stringify(val)).join(',')]);
+
+  useEffect(() => {
+    props.onChange({ target: selfRef.current });
+  }, [value.map(val => JSON.stringify(val)).join(',')]);
+
+  useImperativeHandle(selfRef, () => ({
+    get dirty() {
+      return props.value.map(val => JSON.stringify(val)).join(',') !== this.value.map(val => JSON.stringify(val)).join(',');
```
Bloodyaugust, line 197: Note: this uses `JSON.stringify` to handle the object option case. It will break if that object contains cyclical references. Are we okay with that? Should we leave a comment about it somewhere?

src/components/multi-select/multi-select.test.jsx
```diff
@@ -0,0 +1,500 @@
...
+            label: 'Foo',
+          },
+          {
+            id: 'test-2',
+            label: 'Bar',
+          },
+        ]}
+        optionIDGetter={option => option.id}
+        optionLabelGetter={option => option.label}
+      />);
```
cbacon91, line 137: Seeing this in a test really made me start to think about how I'd like to use this component as an IC outside of Web-inf - I think this is the most common case (`id` or `value` coupled with `label`), I think all strings is less common.

Strings are really only useful for enums, but id/label pair is useful for most things - selecting a user (label is their name / email, id is their .. id) or something like that.

I'd personally say that I think `{ value, label }` is the coupling I've seen the most anecdotally, that's what [angular material uses](https://material.angular.io/components/select/overview), it's what [react-select uses by default](https://react-select.com/home) and I think it's also what [bigmaven typically uses, at least historically](https://github.com/mavenlink/mavenlink/blob/e9b60c29e80a006c9aff0f058c734d83eaa75f7f/frontend/components/inputs/select/select.jsx#L180).

I think providing a familiar API is pretty key to a shared library.

For people who _want_ to use strings, it's obviously very easy for them to do that:

```jsx
const options = ['fart', 'dart', 'cart', 'mart'];
<MultiSelect options={options.map(o => { label: o, value: o })} />
```

I kind of like the idea of providing these getters for when there are edge cases (like `id`/`label`), but one thing that I think makes the _most_ sense is having an opinion here.

If we tell people that the selects take `{ value, label }`, then when the props are handed to the component, they are mapped to this structure, but it only happens once (when the props are passed), instead of a ton of times internally in our component.

```jsx
const users = [{
  id: 2,
  full_name: 'Tom',
}, {
  id: 3,
  full_name: 'Dick',
}, { 
  id: 4,
  full_name: 'Harry',
}];
const options = users.map(user => ({ label: user.full_name, value: user.id }));
<MultiSelect options={options} />
```

I think this is lower friction than telling people to provide a list and two functions on how to fetch the data - just pass us a list in the right format. It can also make our documentation clearer - we don't have to explain the getter functions or how they work, we just say "use `label` and `value` pls". We could even have examples on how to map if that's a concern (which it should not be).

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+            labelledBy={ids.label}
+            refs={valueRefs}
+          >
+            {value.length !== 0 &&
+              props.tagChildren ?
+              props.tagChildren(value, valueRefs, onOptionRemove) :
+              value.map((val, index) => (
+                <Tag
+                  defaultActive={index === 0}
+                  id={`${props.id}-option-${props.optionIDGetter(getOption(val))}`}
```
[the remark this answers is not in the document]
cbacon91: EDIT: This comment might be superflous depending on the discussion of `value` and `label` I pointed out https://github.com/mavenlink/design-system/pull/311#discussion_r634537793

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,80 @@
...
+
+.popup-container {
+  composes: container from '../listbox/listbox.css';
+  background-color: var(--white);
+  box-sizing: border-box;
+  max-height: 160px;
+  overflow-y: scroll;
+  position: absolute;
+  width: 100%;
+  z-index: 10;
```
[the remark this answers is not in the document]
Bloodyaugust: This is one of those things that I'd _really_ love to have in the design system (z-indicies mappings). Until that happens, I tend to go low, until there's a reason to bump it up. This `z-index` is copied from the existing `CustomFieldInputMultiChoice`, so we know it at least works for all the same use-cases (including the configure modal in Custom Forms).

I'd prefer not to change it to be higher until we have a specific use-case that is exemplifying bad behavior.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+  listboxChildren: PropTypes.func, // eslint-disable-line react/no-unused-prop-types
+  name: PropTypes.string.isRequired,
+  onChange: PropTypes.func,
+  options: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.object), PropTypes.arrayOf(PropTypes.string)]).isRequired,
+  optionIDGetter: PropTypes.func,
+  optionLabelGetter: PropTypes.func,
+  placeholder: PropTypes.string,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  showLoader: PropTypes.bool, // eslint-disable-line react/no-unused-prop-types
```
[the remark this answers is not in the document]
Bloodyaugust: See my notes, this is indeed used, the linter is just wrong. 😭

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,80 @@
+@import '../../styles/colors-v2.css';
+@import '../../styles/spacing.css';
+
+.container {
+  composes: tag-list from '../tag-list/tag-list.css';
+  composes: input-container from '../__internal__/abstract-custom-field/abstract-custom-field.css';
```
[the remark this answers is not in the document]
Bloodyaugust: I agree that it seems weird, but at this point, the `abstract-custom-field` mostly just encapsulates the styles we apply to many of our input and input-like components. Until we refactor that styling away to a different, better-named component, these are the correct styles for our container, so I think that composing them makes sense.

src/components/multi-select/multi-select.jsx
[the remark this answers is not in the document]
Bloodyaugust: Good question! I don't think we should structure all our code based on one-liners from Knuth, but on the other hand, `useMemo` does seem like it might be overkill here... Especially considering it would complicate the component by a not insignificant amount (needing to re-memoize when `value` and `options` change).

src/components/multi-select/test-queries.js
```diff
@@ -0,0 +1,50 @@
...
+  if (optionLabel) {
+    return screen.getByRole('button', { name: `Remove ${optionLabel}` });
+  }
+
+  return screen.getByText(`Remove all selected options on ${fieldLabel}`);
+}
+
+export function getSelectedOption(fieldLabel, optionLabel) {
+  const taglist = screen.getByRole('grid', { name: fieldLabel });
+  return getByRole(taglist, 'gridcell', { name: optionLabel });
```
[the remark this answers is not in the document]
Bloodyaugust: Exactly! See the `test-queries` of `CustomFieldInputMultiChoice` for more examples of this.

src/components/multi-select/multi-select.test.jsx
```diff
@@ -0,0 +1,500 @@
...
+      expect(document.querySelector('.unique-input-class')).toBeInTheDocument();
+      expect(document.querySelector('.unique-tag-list-class')).toBeInTheDocument();
+    });
+  });
+
+  describe('id API', () => {
+    it('sets id of container and selected options', () => {
+      render(<MultiSelect {...requiredProps} id="unique-id" value={['Foo']} />);
+
+      expect(document.querySelector('#unique-id')).toBeInTheDocument();
```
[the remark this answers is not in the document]
Bloodyaugust: I think it's indeed sixes, but if we wanna go with a particular style, I can definitely change this.

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: Yeah, this is tricky. On the one hand, I think for maximum composability in the simplest use-case, we do want to provide support for string literals. OTOH, if it's functionality we'll literally never use, then why write it in as a possibility?

Further, in the case where an object is the only thing we allow, what's the best default? It seems like it could either be `value` or `id` (`id` makes more sense to me, as that's what we really need, a unique identifier), but even inside your post there doesn't seem to be a real consensus. In our discussion during arch meeting, we couldn't land on a specific format either.

In the case that we do use an object, I think we still want to provide the `getter*` props API. IMO it's much simpler to tell a component how to parse my format in a very explicit way than to have to transform my own objects to pass them in, then do some transformation or key mapping _again_ when I pull the values out.

All this is to say... I can definitely see dropping support for string literals if we think we won't ever use them, and also having a default object format that conforms to `{id: 'unique-identifier', label: 'some-label}`, but I think anything else is going to take some convincing on my end.

Bloodyaugust: By way of example for supporting `getter*` props API:
```
const someData = [{account: '1', name: 'test 1'}, {account: '2', name: 'test 2'}...];

// MultiSelect does not support `getter*` props API
<MultiSelect options={someData.map(data => {id: data.account, label: data.name})} ... />

// doing something with value
const dataToSendToBackend = multiSelectRef.current.value.map(val => {account: val.id, name: val.name});
// OR
const dataToSendToAnotherComponent = multiSelectRef.current.value.map(val => someData.find(data => data.account === val.id));

// MultiSelect supports `getter*` props API
<MultiSelect options={someData} optionIDGetter={option => option.account} optionLabelGetter={option => option.name} />

const dataToSendToBackend = multiSelectRef.current.value;
```

cbacon91: So a big part of the reason I think `{ value, label }` is used so frequently is because it mimics what a native `<option>` does - 


https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#basic_select

```html
<select name="choice">
  <option value="first">First Value</option>
  <option value="second" selected>Second Value</option>
  <option value="third">Third Value</option>
</select>
```

`id` is useful for us, sure, but not everything will be ids. What about enum values? The value of an enum isn't necessarily an `id`. It's a _unique identifier_, so if you twisted my arm I'd call it an id, but I would never describe something like `premier_plan` as an id.

I personally think that we can just eschew the getter args _entirely_ in favor of just being opinionated and saying that `options` must always be of the format `{label, value}`, but I could also accept `{label, id}` (it's just not my _favorite way).

I think we can _keep_ the functions, I just don't think they're needed. The way I see it, if we tell consumers of the api that we accept `{label, value}` but we have functions for working around that, there's no reason to provide a default. I'd _rather_ we have the one format and just tell people to make their data structures fit into our world view. 

I don't know if what I said in this comment is any different from my first one - I just kinda feel like every time _I_ have used a `<select>`, I've assumed it accepted an array of `{ label, value }` and been surprised when it accepted something different, primarily because that's what angular, react-select, and native option elements use. I really, genuinely feel that any shared system should [follow the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

I like that we give folks the freedom to abandon that paradigm when it makes sense (such as having a list of users which are all `{ id, full_name }`), but the way I would get around that is just by mapping over the data before handing it to a `<select>` component.

Bloodyaugust: @cbacon91 as an IC using this component, and having a look at my example above, which would you prefer to use?

cbacon91: If I'm brutally honest, I have never seen another team use `ref` outside of very specific instances, like when `focus` is relevant, so I would expect that my usage would be more along the lines of 

```jsx
const [selected, setSelected] = useState([]);

const onChange = (data) => {
  setSelected([...selected, data.event.target.ivalue]);
}

<MultiSelect onChange={onChange} />
```

In a case where I'm using callbacks and some parent state, it's 100% sixes. I can use `target.id` or `target.value`, I don't care a ton. The only friction I have at all is the level of configuration I have when building the component initially, and `collection + map`, to me, is easier than `collection + value getter + label getter`.

I'm super happy to be wrong here, I just think if I was still on RPM and was trying to use this select component, this would be my experience.

Bloodyaugust: In your example, you've elided the added complexity of data munging by not using `getter*` I was trying to point out. Even if using `onChange`, you still need to munge the data going in and coming out in the case the data you're working with does not already match a prescribed format. Whether or not we use `ref` is not what I was trying to get at.

cbacon91: I guess I don't understand.

Regardless, the base of everything I'm trying to say is that what _I would assume_ is that the default structure of `options` is of the format `{ label, value }`. Whether or not there are getters or helpers or something is something, if I'm honest, I don't think I would ever use unless I had to. Added complexity from munging or no, the familiarity of a fairly common `label`/`value` pair would be beneficial to me. I could take or leave the functions. 

What _I_ would prefer the most is that the base assumption (for the default getter / setter functions) is that `options` is an array honoring that structure, but like I said before, it's all sixes. I don't think it _matters_ a ton, I think since the configuration is there, it doesn't matter. I'm personally just a fan of leaving off as much configuration as I can and just rolling with well-reasoned defaults, and I don't think I agree that bare strings are well-reasoned defaults, but _in my opinion_ since the configuration is there, I don't think it's worth discussing a ton more. I just know that 6 months ago, if I saw this component, I know the assumptions I would make, and I feel like I'd be very confused without trudging through documentation.

Bloodyaugust: I think we agree that having sane defaults is the way to go, and I'm definitely on board for removing string literals as a possibility for `options`.

I think where we disagree is on whether it's easier to munge a dataset as it goes in/out of the component, or telling the component what it needs to know about the structure of the data going in.

I appreciate the reference to how a native HTML select works! I think that in and of itself is a good enough reason to make the `getter*` default match that format, if we go that route (removing support for string literals). I'm loathe to make any decisions based on what `react-select` does because if we wrote something that complex I would die... But native HTML is great to mimic. 😄

src/components/multi-select/multi-select.css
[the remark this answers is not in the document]
juanca: If possible, let's re-use the styling from MDS Input instead!

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,80 @@
...
+
+.input {
+  border: none;
+  flex-grow: 1;
+  flex-shrink: 1;
+  height: var(--spacing-x-large);
+  width: 0;
+}
+
+.input-readonly {
```
juanca, line 36: Read-only should not have any "interactive buttons". In this case, it should not have a clear button (regardless if it is disabled).

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,80 @@
...
+  composes: input;
+  background-color: transparent;
+}
+
+.form-control-children-container {
+  composes: input-styles from '../__internal__/abstract-custom-field/abstract-custom-field.css';
```
juanca, line 42: If possible, let's re-use the styling from MDS Input as a means of deprecating the god abstract input component.

src/components/multi-select/multi-select.jsx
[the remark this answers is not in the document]
juanca: > You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

I think we can just avoid `useMemo` unless we're in a feature space that loads 1000s of objects onto the screen. For a per-component basis, it does not matter.

src/components/multi-select/test-queries.js
```diff
@@ -0,0 +1,50 @@
+import {
```
juanca, line 1: While I was upgrading `@testing-library/react` in bigmaven, I realized we also have a lot of helper queries in our features. It would have made upgrading much easier if all the test queries were `findBy*` because of loading issues. 

I think we can take a pro-tip from this as well: if we want features to re-use our components then we need to make it easier to test and change up their tests. I think if we just provide `findBy*`s then they can easily always `await findBy*`. 

What do you think?

src/components/multi-select/test-queries.js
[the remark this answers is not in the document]
juanca: Another reason to do it like this: if want to ~~require~~ encourage accessibility so each "option" is really part of a highly-accessible form control -- i.e. has a label.

src/components/multi-select/__snapshots__/multi-select.test.jsx.snap
```diff
@@ -0,0 +1,77 @@
...
+            <div
+              class="icon-container"
+            >
+              <svg
+                height="6"
+                role="img"
+                width="10"
+              >
+                <title>
+                  Open multi select options
```
juanca, line 55: It would be interesting to interpolate the field label in here. I think it's a better experience to read:

> Open test label select options

And, if we ensure our accessible items are _unique_ then it means we can `userEvent.click(screen.findByTitle(<unique label>))` instead of having to do other shenanigans like a `getAllByTitle` and indexing the array or finding the field container and then querying the inside.

juanca: And the test failure message would be nicer to read! Instead of a generic message, it might be faster for developers to read that a particular SVG was not found. 🤔

src/components/multi-select/__snapshots__/multi-select.test.jsx.snap
```diff
@@ -0,0 +1,77 @@
...
+              class="tag-list"
+              role="grid"
+            >
+              <input
+                aria-autocomplete="list"
+                aria-controls="test-id-listbox"
+                aria-describedby="test-id-empty"
+                aria-expanded="false"
+                aria-haspopup="listbox"
+                aria-labelledby="test-id-label"
```
juanca, line 38: We don't necessarily need this. A native form control is labelled by `<label for="...">` or `aria-label` or `aria-labelledby`. I think the rule of thumb is to use either native or `aria` stuff and never both? I'm not sure tbh but I think simpler code the better.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+              onBlur={onAutocompleteBlur}
+              onChange={onAutocompleteChange}
+              placeholder={value.length === 0 ? props.placeholder : undefined}
+              readOnly={props.readOnly}
+              required={props.required ? value.length === 0 : false}
+              ref={autocompleteRef}
+              value={autocompleteValue}
+            />
+          </TagList>
+          <FormControlIcons validationMessage={props.validationMessage} className={classNames.iconsContainer}>
```
juanca, line 259: For some odd reason, we're missing the invalid icon. 
image

juanca: While we're talking about an the invalid state, we need to also make the border red. See previous implementation:

image

src/components/multi-select/multi-select.jsx
juanca: @cbacon91 I recall you fixed something like recently. Do you know how to fix these? I couldn't figure it out -- is our eslint just outta date or something? 🤔

juanca: It might have to do with the fact that the prop is used in a "internal function component" so maybe it can't correlate the usage of props in there with the parent's closure. smh linters.

Bloodyaugust: I have the same feeling. Can't differentiate separate function scopes sometimes, for some reason. 😬

src/components/multi-select/multi-select.css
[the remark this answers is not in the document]
Bloodyaugust: I can look into that!

src/components/multi-select/multi-select.css
Bloodyaugust: Ah, good to know! I'll make that change.

src/components/multi-select/__snapshots__/multi-select.test.jsx.snap
Bloodyaugust: I can make that change!

src/components/multi-select/__snapshots__/multi-select.test.jsx.snap
Bloodyaugust: Cool, I'll remove this attribute.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Interesting, this may be an unintended regression. I'll check it out and get it working properly.

src/components/multi-select/test-queries.js
Bloodyaugust: My plan was to compose these queries into new, `findBy`-based ones for the autocompleter, since the multi-select has no inherent async state. I don't have a strong opinion on this however, so if we think it'll be easier to use `findBy` here than compose these functions later, I can do that.

src/components/multi-select/multi-select.css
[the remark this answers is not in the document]
juanca: The Date stuff was probably copy-pasta or incorrect z-index usage tbh.

AFAIK, a dialog would rely on block context where we would force a new component to be in a new "context". This implies that any previous context can _never_ be rendered above the new context. In this sense,

```
<html>
  <body>
    <normal app>
      <popup z-index="over 9000" />
    </normal app>
    <new context>
      <dialog z-index="1" />
    </new context>
  </body>
</html>
```

The normal app's popup would have a z-index of its value + its parent's values. Regardless of the computed value, it can never be "on-top" of the new context -- even if the dialog's computed z-index is lower than the popup's. 

It's some crazy stuff and really annoying at the same time. But it would help us in this case. Last I looked into this kind of stuff was when I was re-creating Gantt in Backbone Marionette so that was like 5 years ago. I wonder if browsers have changed 🤔 

Regardless, I think we're good. We should strive for smaller z-index bumping. Given that, the "max integer for date" sounds like a relatively small mistake.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+                icon={iconClear}
+                label={`Remove all selected options on ${props.label}`}
+                onPress={onOptionsClear}
+                className={classNames.iconClear}
+              />
+            )}
+            <Icon icon={iconCaretDown} label="Open multi select options" />
+          </FormControlIcons>
+        </div>
+        {dropdownContents()}
```
juanca, line 271: The original design intent was to render the popup under any validation message. This was meant to help users fix their own problems without "losing context" when choosing a new valid option.


Multi-select (bad):
image

Multi-choice (still bad): 
image


Date (good):
image

juanca: On top of that, notice the tiny little spacing between the form control and the popup. It is the same as the spacing in date:

image

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+  options: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.object), PropTypes.arrayOf(PropTypes.string)]).isRequired,
+  optionIDGetter: PropTypes.func,
+  optionLabelGetter: PropTypes.func,
+  placeholder: PropTypes.string,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  showLoader: PropTypes.bool, // eslint-disable-line react/no-unused-prop-types
+  tagChildren: PropTypes.func,
+  validationMessage: PropTypes.string,
+  value: PropTypes.arrayOf(PropTypes.string),
```
juanca, line 298: What about the usages of `value` being a list of objects?

src/components/multi-select/multi-select.jsx
juanca: if you want, you can extract the "internal component" into a private component in this file. I usually wrap it in `/* eslint-disable prop-types-stuff */` and do not write tests (since the tests are on the actual exported component). It might "clean up" the component too?

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+  const ids = {
+    emptyMessage: `${props.id}-empty`,
+    label: `${props.id}-label`,
+    listbox: `${props.id}-listbox`,
+    textbox: `${props.id}-autocomplete`,
+  };
+
+  const classNames = {
+    container: styles.container,
+    formControlChildrenContainer: props.readOnly ? styles['form-control-children-container-readonly'] : styles['form-control-children-container'],
```
juanca, line 47: I think this is more aptly named as something along the lines of "input" and the input below is more of a "combobox"? 

Or maybe this is: `combobox-container` (since it contains a combobox)? 

This does separate the classnames from the React and instead relies on layout and roles. 🤔  

What do you think?

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+    iconClear: props.readOnly ? styles['icon-clear-readonly'] : styles['icon-clear'],
+    iconsContainer: styles['icon-container'],
+    input: props.readOnly ? styles['input-readonly'] : styles.input,
```
juanca, line 50: Consider `comboxbox` to mirror the roles instead of the markup?

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+    noOptionsContainer: styles['no-options'],
+    popupContainer: styles['popup-container'],
+    tagList: styles['tag-list'],
+    ...props.classNames,
+  };
+
+  const dropdownContents = () => {
```
juanca, line 57: Consider extracting this into its own private component.

src/components/multi-select/multi-select.jsx
juanca: I think that's okay. Most of the time, the values will be IDs anyways. When a developer runs into a cyclical issue then they can just pass down an ID and then use the `onChange` to map the selected ID to their respective object.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Not sure I love this option. I think it unnecessarily complicates the flow by introducing more props passing, creates additional points of failure, and complicates the file by putting two component declarations in the same place. We still also need to use `eslint-disable`, so it also doesn't dig us out of that hole. 😬

src/components/multi-select/multi-select.css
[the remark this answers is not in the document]
Bloodyaugust: It depends on how the dialog is implemented, but yeah, usually it involves creating a new "stacking context", and like you said, there's no CSS way to make something in one context appear above another context. That said, we're safe in the case where the dropdown is a child of the new context, as the z-index will be correct in that set of children.

See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

src/components/multi-select/multi-select.jsx
Bloodyaugust: Cool, that makes sense. I'll fix both the error state and "idle" state separation.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Oooooh, this is interesting! I definitely meant to backfill a change to this when I implemented the object handling behavior and forgot. What makes it interesting is that I get no prop-type errors in the console for the object example in MDS site or the tests...

src/components/multi-select/multi-select.jsx
Bloodyaugust: Yeah, that's an interesting viewpoint. Do we want the class names to reflect what is in JSX, or what is in the accessibility tree? I think I've tended towards what is in JSX because not all sets of elements that need styling have any relevant accessibility. Not sure that's a great reason to not mix the two, but I think it does make it simpler to grok.

src/components/multi-select/multi-select.jsx
Bloodyaugust: I think this is the same discussion as above for `formControlChildrenContainer`.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Additionally, I think renaming `formControlChildrenContainer` to `input` may be out-and-out confusing for anyone reading the code when they see that the literal `input` is not the `input` _className`. 😬

src/components/multi-select/multi-select.jsx
Bloodyaugust: See this comment: https://github.com/mavenlink/design-system/pull/311#discussion_r635657618

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+  return (
+    <div className={classNames.container} id={props.id} ref={wrapperRef}>
+      <FormControl
+        label={props.label}
+        labelId={ids.label}
+        id={ids.textbox}
+        error={validationMessage}
+        onKeyDown={onKeyDown}
+      >
+        <div role="presentation" className={classNames.formControlChildrenContainer} onClick={onClick}>
```
juanca, line 215: If the `<FormControlIcon />` is absolutely positioned, should we consider removing this wrapping div in favor of customizing `<FormControl />` with a better class names API and perhaps an `onClick`?

src/components/multi-select/multi-select.jsx
juanca: No, I think we still need the event to propagate in case there are window click event handlers. But I do like the idea of we "prevent the default" of opening the popup.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+          <FormControlIcons validationMessage={props.validationMessage} className={classNames.iconsContainer}>
+            {value.length > 0 && (
+              <IconButton
+                icon={iconClear}
+                label={`Remove all selected options on ${props.label}`}
+                onPress={onOptionsClear}
+                className={classNames.iconClear}
+              />
+            )}
+            <Icon icon={iconCaretDown} label="Open multi select options" />
```
juanca, line 268: I know this wasn't implemented in the original multi-choice field, but I think we want an `IconButton` here for an interactive `MultiSelect` and an `Icon` for a read-only mode. 

In read-only mode, the label is incorrect as well. I wonder what's a good label for it? Maybe none since it's purely presentational?

src/components/multi-select/multi-select.jsx
cbacon91: > @cbacon91 I recall you fixed something like recently.

I didn't fix anything, I just removed some rules that weren't needed.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,316 @@
...
+  const wrapperRef = useRef(null);
+
+  const ids = {
+    emptyMessage: `${props.id}-empty`,
+    label: `${props.id}-label`,
+    listbox: `${props.id}-listbox`,
+    textbox: `${props.id}-autocomplete`,
+  };
+
+  const classNames = {
```
juanca, line 45: We can "streamline" this:

```js
const classNames = {
  ...MultiSelect.defaultProps.classNames,
  ...props.classNames
}
```

where

```js
MultiSelect.defaultProps = {
  classNames: {
    container: ....,
    formControlChildrenContainer: ...,
     ...
  {
};
```

src/components/multi-select/multi-select.jsx
Bloodyaugust: That's an interesting proposition. I think we should keep this in mind for the future, and if we see this kind of use-case keep coming back up, then consider extending `FormControl`. As it is, I don't think this is a bad thing to do here considering the complexity of our use-case, and keeping `FormControl` as simple as possible is a win in my book.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Made this change, but `Icon` requires a `label` attribute. LMK if you don't like the `Opening ${props.label} options disabled while read only` label I went with.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Note: while we discussed this in 1:1 review, we realized this approach doesn't work, because we need the component context to determine classNames. Further exploration of changes to this pattern to support a better `defaultProps` declaration is needed.

src/components/multi-select/multi-select.css
Bloodyaugust: See this comment: https://github.com/mavenlink/design-system/pull/311#discussion_r638531610

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,86 @@
+@import '../../styles/colors-v2.css';
+@import '../../styles/spacing.css';
+@import "../../styles/typography-v2.css";
+
+.container {
+  composes: tag-list from '../tag-list/tag-list.css';
+  height: auto;
+  margin: var(--spacing-small) 0;
```
juanca, line 8: AFAIK, this shouldn't have any margin. This is a screenshot of our MDS Input:

image


This is the screenshot of the MDS MultiSelect:

image

juanca: Actually, this raises a concern:

- Do we want to make the whole component's container the "tag list"? I would figure the "tag list" is actually the "form control children container" layer.

src/components/multi-select/multi-select.css
```diff
@@ -0,0 +1,86 @@
...
+  composes: popup-container;
+  composes: no-options from '../no-options/no-options.css';
+  color: var(--mds-grey-54);
+  display: flex;
+  justify-content: center;
+  overflow-y: unset;
+}
+
+.tag-list {
+  composes: tag-list from '../tag-list/tag-list.css';
```
juanca, line 80: Why do we have two tag-lists in this component? I think this might have been an oversight?

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,342 @@
...
+  }
+
+  function getVisibleOptions() {
+    return props.options
+      .filter(option => props.optionLabelGetter(option).toLowerCase().includes(autocompleteValue.toLowerCase()))
+      .filter(option => !value.some((val) => {
+        const optionID = props.optionIDGetter(option);
+
+        if (typeof optionID === 'string') {
+          return val.toLowerCase() === optionID.toLowerCase();
```
juanca, line 125: I would wager that the fact that these are not the same before lower casing implies they are different. 

If you happen to have two option tags,

- `<option value="test-1">test 1</option>`
- `<option value="Test-1">Test 1</option>`

Then you would assume those two are unique, different options, correct?

src/components/multi-select/multi-select.jsx
juanca: Ah, yeah, I think so. Let's remove the case-insensitivity if my comment below checks out.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,342 @@
...
+    setExpanded(true);
+    setAutocompleteValue(event.target.value);
+  }
+
+  function onDropdownClose() {
+    setExpanded(false);
+    setAutocompleteValue('');
+  }
+
+  function onOptionRemove(event) {
```
juanca, line 153: I feel like _something_ should be focused after the underlying DOM element is removed. But I don't know what that should be. And I think I found a bug in TagList. So I made a new story to represent that work: https://www.pivotaltracker.com/story/show/178305087

src/components/multi-select/multi-select.md
```diff
@@ -0,0 +1,163 @@
+`MultiSelect` allows for the selection of multiple values in an accessible, searchable interface. It also provides flexibility in how you render selected options and options in the dropdown, but the default is MDS compatible and assured.
```
juanca, line 1:
```suggestion
See [design specifications.](https://www.notion.so/MultiSelect-9da6ee3245d946699106231cf161f783)

`MultiSelect` allows for the selection of multiple values in an accessible, searchable interface. It also provides flexibility in how you render selected options and options in the dropdown, but the default is MDS compatible and assured.
```

juanca: Welcome to edit the document as well if you want to call out anything specific. Otherwise, a link is fine for now.

src/components/multi-select/multi-select.test.jsx
```diff
@@ -0,0 +1,516 @@
...
+        {...requiredProps}
+        classNames={{
+          container: 'unique-container',
+          formControlChildrenContainer: 'unique-form-control-children-container',
+          input: 'unique-input-class',
+          tagList: 'unique-tag-list-class',
+        }}
+      />);
+
+      expect(document.querySelector('.unique-container')).toBeInTheDocument();
```
juanca, line 49: Welcome to use a snapshot for these kinds of tests too!

src/components/multi-select/multi-select.test.jsx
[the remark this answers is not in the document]
juanca: I'm more of a fan of `document`. 😅  I dislike the `container` queries in general because it can miss element that are rendered outside the component (with things like a React Portal). 

As for this particular test, I think this is a smell that our role is on the wrong item. I've been looking at the ARIA implementation of comboboxes for a few weeks now and I think the `role="combobox"` is _suppose_ to be on the container. 

However, that is the W3 ARIA 1.1 specifications. What we have now is W3 ARIA 1.0 specifications. So either is fine but I think if we implement 1.1 in the future then the test would be more readable: 

```suggestion
      render(<MultiSelect {...requiredProps} id="unique-id" value={['Foo']} />);

      expect(screen.getByRole('combobox')).toHaveAttribute('id', 'unique-id');
```

src/components/multi-select/multi-select.test.jsx
```diff
@@ -0,0 +1,516 @@
...
+
+      await openOptions('test label');
+      userEvent.click(await await findAvailableOption('test label', 'Foo'));
+      expect(value).toContain('Foo');
+    });
+  });
+
+  describe('onInput API', () => {
+    it('fires the onInput event when the user types in the combobox input', async () => {
+      let value = null;
```
juanca, line 113: This is odd. The standard patterns for testing callbacks is to use spies.

e.g. 

```
const spy = jest.fn();
render(<Component onInput={spy} />)
userEvent...
expect(spy).toHaveBeenCalledWith({ target: { value: ... }});
```

src/components/multi-select/multi-select.test.jsx
```diff
@@ -0,0 +1,516 @@
...
+      await openOptions('test label');
+      expect(screen.getByText('Override Foo')).toBeInTheDocument();
+      userEvent.click(screen.getByText('Override Foo'));
+      expect(await findSelectedOption('test label', 'Foo')).toBeInTheDocument();
+    });
+  });
+
+  describe('onChange API', () => {
+    it('fires the onChange event when the value changes', async () => {
+      let value = [];
```
juanca, line 97: This is odd. The standard patterns for testing callbacks is to use spies.

e.g. 

```
const spy = jest.fn();
render(<Component onInput={spy} />)
userEvent...
expect(spy).toHaveBeenCalledWith({ target: { value: ... }});
```

src/components/multi-select/multi-select.test.jsx
juanca: Yeah, native HTML is probably the thing to mimic. If we do need a highly-tailored bigmaven interface then we can just make the wrapper that does all the nice to configure. 

I think the power of composition is probably better for us than the ease-of-use of bigmaven interfaces since we can always make the "Project autocompleter" that does the right thing.

src/components/multi-select/multi-select.jsx
```diff
@@ -0,0 +1,342 @@
...
+            )}
+          </FormControlIcons>
+        </div>
+      </FormControl>
+      {dropdownContents()}
+    </div>
+  );
+});
+
+MultiSelect.propTypes = {
```
juanca, line 301: I realized if we're going to be okay with missing test coverage on "pass through props", there's no point in double check the test coverage for our props interfaces? I kinda seems silly now.

juanca (approved):
Looking great! I made a few stories to punt on some a11y functionality. I also didn't really cross check the test coverage due to our new perspective of being okay with leverage existing tests in other files.

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: Ooooh, that would make it more straightforward. I'll do that!

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: I know we've used the pattern as written before, but I can use the spy pattern instead.

src/components/multi-select/multi-select.jsx
Bloodyaugust: Yeah, this test suite was written before our latest conversation on the subject, so it does indeed have full coverage. 🙃

src/components/multi-select/multi-select.test.jsx
cbacon91: One thing I'll comment on after working with our `<Select>` component...

I'm a little surprised our [Select component](https://mavenlink.github.io/design-system/master/#/Components/Select) doesn't have this in any fashion. Like, I was trying to make enums work in Custom forms and it pained me just how verbose I had to be 

https://github.com/mavenlink/mavenlink/blob/5d0017c0e974b35a99c5c28892d97016d3ca1e8e/frontend/features/custom-forms/components/field.jsx#L69-L84

I expected to just write something like `<Select options={[1,2,3,4]} />` like we talked about here, or `<Select options={[1,2,3,4].map(r => ({ label: r, value: r }))} />` if it needed an object, but as far as I could tell from the component, neither of those existed - the `select` just runs on children (and I guess `refs`).

I don't think that's worse (or better). On the one hand, it's closer to native by requiring children instead of relying on objects, but on the other hand it's ... just very verbose. It's nice that the option for custom stuff is _there_, but I feel like that's the exception, not the rule.

I kind of like that the multiselect you've made here supports both the verbose / composable case as well as the basic "I have an array of stuff, render as options pls" case. It might be ekeing into `react-select` / god component / 2manyprops territory, but I like the convenience personally

src/components/multi-select/multi-select.jsx
Bloodyaugust: Case insensitivity for filtering options was kept after discussion, when the new `filterOptions` prop is `true`.

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: Yeah, the `Select` is pretty indisputably more difficult to work with for the reasons you mention, even if it does feel more like a native HTML implementation.

I've also been concerned over the large props surface area `MultiSelect` has at this point. I've done my best to keep it down to as little as possible, but perhaps there is a possibility for even lower-level components that implement less customization capability. The problem then becomes maintaining that simplicity while still allowing for higher-level abstractions to use it compositionally _and_ still be capable of adding the functionality they need.

src/components/multi-select/multi-select.css
Bloodyaugust: I've removed the extra margin here.

Re your concern: I think it makes sense from a compositional standpoint (and also from a layout/styling standpoint) to keep the `FormControlIcons` and children out of the `TagList`. They're not tags, they don't represent state that is directly derived from specific options, and they need to stay in the top-right-hand corner of the "input", in a different flex container.

src/components/multi-select/multi-select.css
Bloodyaugust: I think you're referring to this `compose` appearing twice. If so, you're correct, the instance in `.container` was extraneous now after layout changes earlier in the dev cycle, and has been removed.

src/components/multi-select/multi-select.jsx
Bloodyaugust: This code was removed, and hopefully the intent clarified now that we don't support string literal options. For posterity, this was detecting options as string literals and providing case-insensitive filtering with autocompleter input for that case.

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: This is actually not super straightforward, because of React `SyntheticEvents`. Basically, `toHaveBeenCalledWith` is going to be checking a `SyntheticEvent` that has already been returned to the pool, and therefore doesn't look like `{ target: { refStuff }}` anymore.

We can ensure that it was _called_ with this approach, but not what it was called with, unless we use `event.persist` which has other implications for production. See this comment from Kent Dodds, and the following 2: https://github.com/testing-library/react-testing-library/issues/175#issuecomment-422189880

For now, I'm going to go with `toHaveBeenCalled`. If we want to guarantee the event contract looks like what we need, I suggest we switch back to my non-standard pattern.

src/components/multi-select/multi-select.test.jsx
Bloodyaugust: See this comment: https://github.com/mavenlink/design-system/pull/311#discussion_r640196456

PR #316, opened 2021-05-19
Upgrade styleguidist

juanca:
https://www.pivotaltracker.com/story/show/177913602

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/upgrade-styleguidist/
- [x] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/24802


styleguide/components/styleguide/styleguide.css
```diff
@@ -1,6 +1,10 @@
 @import "../../../src/styles/colors-v2.css";
 @import "../../../src/styles/spacing.css";
 
+:global html {
```
juanca, line 4: I couldn't figure out a way to do with in the `styleguide.config.js` file. Supposedly, the `styles` key-value should take something along the lines of:

```
'@global html': { scrollPaddingTop: ... }
```

But it didn't work. I think for now this is fine and maybe later we can find more info on how to do better global styling in react-styleguidist.

cbacon91 (approved):
neat

PR #330, opened 2021-06-23
Add help icon / tooltip to MDS Input

cbacon91:
https://www.pivotaltracker.com/story/show/176504997

input-tooltip

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-help-me-obi-wan-kenobi-176504997/#!/Components/Input
- [ ] Green Bigmaven CI check


src/components/input/input.jsx
```diff
@@ -117,7 +122,7 @@ Input.propTypes = {
 Input.defaultProps = {
   autoFocus: undefined,
   className: undefined,
-  cssContainer: styles.container,
+  cssContainer: undefined,
```
cbacon91, line 125: `styles.container` did not exist in the stylesheet, so this did nothing.

src/components/help-icon/help-icon.jsx
```diff
@@ -0,0 +1,35 @@
...
+}) {
+  return (
+    <Tooltip text={text} direction={direction} id={id}>
+      <Icon label={label} icon={helpSvg} />
+    </Tooltip>
+  );
+}
+
+HelpIcon.propTypes = {
+  direction: Tooltip.propTypes.direction,
```
juanca, line 27: It is prefer to inline the documentation for the props interface instead of a large block at the top of the file. Using styleguidist comments will generate documentation on the site.

Something like:

```suggestion
  /** Direction for the tooltip to fly from relative to the icon. */
  direction: Tooltip.propTypes.direction,
```

src/components/help-icon/help-icon.jsx
```diff
@@ -0,0 +1,35 @@
...
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+};
+
+HelpIcon.defaultProps = {
+  direction: Tooltip.defaultProps.direction,
```
juanca, line 34: 💯  Nice use of props composition!

src/components/help-icon/help-icon.jsx
cbacon91: It's a bit weird to me that styleguidist prefers inline documentation instead of jsdoc.. jsdoc is pretty standardized

src/components/help-icon/help-icon.md
```diff
@@ -0,0 +1,35 @@
...
+  label="title text for the SVG" 
+  id="example-tooltip"
+/>
+```
+
+#### In-depth example
+
+More realistically, the `HelpIcon` is going to be used in conjunction with other elements, like inputs or paragraphs.
+
+Much like with `Tooltip`, please remember to apply `aria-describedby` on the component the help text is related to.
```
juanca, line 19: I wonder if it's worthwhile to add a `useLayoutEffect` that effectively looks for the `aria-describedby` element. 🤔  Maybe a future iteration.

src/components/input/input.jsx
```diff
@@ -50,6 +50,9 @@ const Input = forwardRef(function Input(props, forwardedRef) {
     },
   }));
 
+  const describedBy = [`${props.id}Hint`];
```
juanca, line 53: A recent pattern we've been kinda liking for managing IDs, refs, etc.

Use a single object variable to keep track of all IDs with respect to the layout/roles. e.g.

```js
const ids = {
  container: props.id,
  label: `${props.id}-label`,
  tooltip: `${props.id}-tooltip`,
  validationMessage: `${props.id}Hint`,
};
```

Up to you if you want to also repeat the pattern.

src/components/input/input.jsx
```diff
@@ -50,6 +50,9 @@ const Input = forwardRef(function Input(props, forwardedRef) {
...
+  if (props.tooltip) describedBy.push(`${props.id}-tooltip`);
```
juanca, line 54: I think it's okay to have a `describedby` with a non-existent element. Since it does not find the element, I think it just skips it.

```suggestion
  describedBy.push(`${props.id}-tooltip`);
```

I would prefer the simpler code than the branching code. But first it's gotta work!

src/components/form-control/form-control.jsx
```diff
@@ -21,13 +22,18 @@ function isInvalid(error, readOnly) {
...
   return (
     <div className={props.className} onKeyDown={props.onKeyDown} role="presentation">
-      <label
-        className={getLabelClassName(props.error, props.readOnly)}
-        htmlFor={props.id}
-        id={props.labelId}
-      >
-        {props.label}
-      </label>
+      <div className={styles['label-wrapper']}>
```
juanca, line 25: This is a long shot buuuut bear with me:

What if instead of using intermediary `<div>` for layout purposes.... the container div is designated as a `display: grid` and we use a template?

```css
.container {
  display: grid;
  grid-template-areas:
    label help
    input input
    valid valid;
```

juanca (approved):

juanca:
Thanks for updating the tests too!

src/components/help-icon/help-icon.md
cbacon91: What would the intent of it to actually be? Just scan the dom to make sure the element is there and raise an error if not or something?

src/components/form-control/form-control.jsx
cbacon91: I started to go down this route but doesn't this mean that each child of `FormControl` needs to have `grid-area: input`? I'm new to grid so I could be misunderstanding but it seems like that'd be a bit more invasive than the extra div

src/components/input/input.jsx
cbacon91: 9a2c0f8d

src/components/help-icon/help-icon.jsx
cbacon91: Ok I dug a bit through how styleguidist prefers its documentation. These are still jsdoc, just not a blob at the top of the file, which is relieving. I think the IDE point might also still work, but I'm not positive about that. Regardless, I did this to make styleguidist happy.

2561428a

src/components/form-control/form-control.jsx
juanca: Right... 🤔  Yeah, let's keep it as-is!

src/components/help-icon/help-icon.md
juanca: Yeah, exactly. It's basically a run-time linter so the dev is setting things up correctly. 

Pros:
- Our stuff is accessible by a forcing linter

Cons:
- CPU cycles wasted for checking validity of app on customer's computers

cbacon91: I think the pros would outweigh the cons.. 

We would only have to do it once per instance of the tooltip, I imagine we would only do it on mount instead of on every render cycle, but there's definitely still a cost there.

I've got some refactors I want to make to the tooltip anyway (mostly investigating react portal), so I'll see if I can add this in that PR. Thanks for the feedback!

PR #333, opened 2021-06-30
fix(tooltip): resolve `overflow: hidden` issue

cbacon91:
## Motivation
Fix issue in tooltip where `overflow: hidden` makes it impossible for the tooltip to be seen in all of its glory.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-portal-just-the-tip/
- [x] Green Bigmaven CI check

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,9 @@ module.exports = {
     'import/extensions': ['error', 'always', {
       ignorePackages: true,
     }],
+    'no-mixed-operators': ['error', {
+      allowSamePrecedence: true,
+    }],
   },
   overrides: [{
     files: [
--- a/src/components/help-icon/help-icon.test.jsx
+++ b/src/components/help-icon/help-icon.test.jsx
@@ -10,6 +10,11 @@ describe('HelpIcon', () => {
     text: 'Help Text',
   };
 
+  // There needs to be something on the DOM that the tooltip is describing, so render a dummy div for each test.
+  beforeEach(() => {
+    render(<div aria-describedby={requiredProps.id} />);
+  });
+
   it('has defaults', () => {
     render(<HelpIcon {...requiredProps} />);
     expect(document.body).toMatchSnapshot();
--- a/src/components/help-icon/help-icon.test.jsx
+++ b/src/components/help-icon/help-icon.test.jsx
@@ -44,21 +44,25 @@ describe('HelpIcon', () => {
   describe('direction api', () => {
     it('allows top', () => {
       render(<HelpIcon {...requiredProps} direction="top" />);
+      user.hover(screen.getByTitle('Icon Label'));
       expect(document.body).toMatchSnapshot();
     });
 
     it('allows bottom', () => {
       render(<HelpIcon {...requiredProps} direction="bottom" />);
+      user.hover(screen.getByTitle('Icon Label'));
       expect(document.body).toMatchSnapshot();
     });
 
     it('allows left', () => {
       render(<HelpIcon {...requiredProps} direction="left" />);
+      user.hover(screen.getByTitle('Icon Label'));
       expect(document.body).toMatchSnapshot();
     });
 
     it('allows right', () => {
       render(<HelpIcon {...requiredProps} direction="right" />);
+      user.hover(screen.getByTitle('Icon Label'));
       expect(document.body).toMatchSnapshot();
     });
   });
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -29,7 +29,7 @@ export default function Tooltip({
   }, []);
 
   // keep the div so enabling/disabling doesn't affect the dom tree or styles
-  if (disabled) return <div>{children}</div>;
+  if (disabled) return <div className={styles.wrapper}>{children}</div>;
 
   const classNames = [className];
   if (direction === 'top') classNames.push(styles.top);
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -24,7 +24,7 @@ export default function Tooltip({
   const { position, visible, show, hide } = useTooltipPositioning({ tooltipRef, triangleHeight, direction });
 
   useLayoutEffect(() => {
-    if (disabled || document.querySelector(`[aria-describedby="${id}"]`)) return;
+    if (disabled || document.querySelector('[aria-describedby]', id)) return;
 
     throw new Error('<Tooltip> was used without an element on the DOM being described by it. ' +
       'Please add `aria-describedby` to the element this tooltip is being used to describe.');
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -1,16 +1,11 @@
-import React, { useState } from 'react';
+import React, { useRef } from 'react';
+import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import styles from './tooltip.css';
+import useTooltipPositioning from './use-tooltip-positioning.js';
 
 /**
  * A component used to render help or reminder text.
- * @param children The component that needs to be hovered or focused to display the tooltip. These children are where the tooltip "appears" from.
- * @param className An optional className for the tooltip to provide styling.
- * @param direction The direction the tooltip should appear from.
- * @param disabled Whether or not the tooltip should appear at all. If this is true, the tooltip will not appear even when the children receive focus.
- * @param id The id that the tooltip should be given. The element that this tooltip is describing <strong>MUST</strong> have <code>aria-describedby={id}</code>.
- * @param text The text inside of the tooltip. Tooltip text should be concise and kept to 1-2 short sentences.
- * @param truncate Whether or not to truncate the <code>text</code> provided. Usage of this prop is generally not recommended unless the text is user-defined.
  */
 export default function Tooltip({
   children,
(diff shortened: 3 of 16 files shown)


.eslintrc.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
     'import/extensions': ['error', 'always', {
       ignorePackages: true,
     }],
+    'no-mixed-operators': ['error', {
+      allowSamePrecedence: true,
+    }],
```
cbacon91, line 16: Happy to revert, but this lint rule made the following arithmetic throw a lint error:

```
const someValue = a + b - c;
```

You _could_ add parentheses to that statement, but adding parentheses there would change the order of operations. You can add them in such a way that it doesn't matter too, but then you're just adding parens for the sake of making lint happy which makes the code uglier. In the real-world example I had, I had something like

```
const value = (x / 2) + (y / 2) - constant;
```

Adding extra parens to this would make it way harder to read imo:

```
const value = ((x / 2) + (y / 2)) - constant;
```

src/components/tooltip/tooltip.jsx
```diff
@@ -21,56 +16,65 @@ export default function Tooltip({
   text,
   truncate,
 }) {
-  const [visible, setVisible] = useState(false);
+  // size of the triangle in pixels; this should be the same as teh value in the linked stylesheet
+  // TODO is there a way to share the variable so its only in one place
+  const triangleHeight = 8;
```
cbacon91, line 21: @juanca I know you told me there was a way to share variables between css and jsx but I couldn't figure out how to do it.

I thought it was just something like `<div styles={{ '--triangle-height': triangleHeight }} />` but when I did that I never saw the value actually used in my css, I might have done something wrong though.. thoughts?

src/components/tooltip/use-tooltip-positioning.test.js
```diff
@@ -0,0 +1,124 @@
...
+          },
+        },
+      }));
+
+      act(() => result.current.show({ target }));
+
+      expect(result.current.position).toEqual({
+        left: sourceRect.x + (sourceRect.width / 2) - (tooltipRect.width / 2),
+        top: sourceRect.y - tooltipRect.height - triangleHeight,
+      });
```
cbacon91, line 64: I didn't like using the actual calculations here, but I was worried about using hardcoded numbers as being unclear to future engineers. I can add them with comments though if that's preferred to the arithmetic

src/components/tooltip/tooltip.jsx
```diff
@@ -21,56 +16,72 @@ export default function Tooltip({
...
 
-  const show = () => setVisible(true);
-  const hide = () => setVisible(false);
+  useLayoutEffect(() => {
+    if (disabled || document.querySelector(`[aria-describedby="${id}"]`)) return;
 
-  const classNames = [className];
+    throw new Error('<Tooltip> was used without an element on the DOM being described by it. ' +
+      'Please add `aria-describedby` to the element this tooltip is being used to describe.');
+  }, []);
```
cbacon91, line 31: @juanca this more or less what you had in mind?

We could probably also make it `return` early for production as well if we don't want to hinder clients. It should fail tests for devs locally so I expect it won't realistically hit production, but I could be wrong on that.

I could also change it from `throw` to `console.error` or something.

.eslintrc.js
juanca: 🚀

juanca:
Generally, positioning absolutely against the document can have unintended and hard to decipher consequences. 

a) Positioning while scrolling
b) Positioning with "distorted" document
c) ... I think there are more but if I don't remember then they probably don't matter in the grand scheme of things

---


Aaaand, here are the results:

a) It scrolls with page. 💯 
b) It miscalculated the tooltip position when the document is scrolled _horizontally_ 

image

This might be fixable though!

src/components/tooltip/tooltip.jsx
```diff
@@ -21,56 +16,72 @@ export default function Tooltip({
   text,
   truncate,
 }) {
-  const [visible, setVisible] = useState(false);
+  // size of the triangle in pixels; this should be the same as teh value in the linked stylesheet
+  // TODO is there a way to share the variable so its only in one place
```
juanca, line 20: Shared across how many places?

CSS <=> CSS: use css variables
JS <=> JS: use exported variables
JS <=> CSS: use css variables, set variable in JS on DOM as `style={{ --le-var: '8px' }}`
JS <=> JS <=> CSS: export all the things!

src/components/tooltip/tooltip.jsx
juanca: Huh.... I can lend my 👀  cuz it should work.

src/components/tooltip/tooltip.jsx
juanca: This is 💯 

> won't realistically hit production

Yeah, it's okay. We should have caught it by deployment time. The ID lookup is constant so it should be blazingly fast. 

> I could also change it from throw to console.error or something.

I think this is fine. Throwing an error is good because we want the code to stop running and make the developer fix it ASAP. At some point, we will implement proper React error boundary handling so it won't be as intrusive -- i.e. blank page.

src/components/tooltip/tooltip.jsx
cbacon91: So it's basically what I mentioned here: https://github.com/mavenlink/design-system/pull/333#discussion_r661669634

JS and CSS are what need to share it - I tried using `style` but ran into issues. I might need your help in it.

cbacon91:
> b) Positioning with "distorted" document

Yeah, I did actually notice that, at one point when I zoomed in using the browser to 125 - 200%, but I just tried it again to take a screenshot and it seems to be working as expected. I'll keep an eye out for it otherwise - any ideas on how to reproduce?

//ss

> b) It miscalculated the tooltip position when the document is scrolled _horizontally_

Ah, crap. Yeah that's super easy to fix! - 8c823e76

src/components/tooltip/tooltip.jsx
cbacon91: Ok, figured it out - I just forgot the units 🤦 

I was using `<div styles={{ '--triangle-height': triangleHeight }} />` when what I should have had was `<div styles={{ '--triangle-height': `${triangleHeight}px` }} />`

src/components/help-icon/help-icon.test.jsx
```diff
@@ -10,6 +10,11 @@ describe('HelpIcon', () => {
     text: 'Help Text',
   };
 
+  // There needs to be something on the DOM that the tooltip is describing, so render a dummy div for each test.
+  beforeEach(() => {
```
juanca, line 14: A more standardized pattern for "setting up the env for a given test suite" is the `beforeAll`. And a similar clean up in a `afterAll`. I think it's more of a convention. Normally, it wouldn't work for Mavenlink (tm) tests because we are accustomed to cleaning up the whole env between each test due to global pollution.

src/components/tooltip/tooltip.jsx
```diff
@@ -19,58 +14,71 @@ export default function Tooltip({
...
 
-  switch (direction) {
-    case 'top': classNames.push(styles.top); break;
-    case 'bottom': classNames.push(styles.bottom); break;
-    case 'left': classNames.push(styles.left); break;
-    case 'right': classNames.push(styles.right); break;
-    default: break;
-  }
+  // keep the div so enabling/disabling doesn't affect the dom tree or styles
+  if (disabled) return <div>{children}</div>;
```
juanca, line 32: This is missing `styles.wrapper`. Not sure if that would affect the rendering of `children`.

src/components/tooltip/tooltip.test.jsx
```diff
@@ -18,6 +18,24 @@ describe('tooltip', () => {
     expect(document.body).toMatchSnapshot();
   });
 
+  it('errors if the tooltip is not describing any elements', () => {
+    /* eslint-disable no-console */
+    const originalConsoleError = console.error;
```
juanca, line 23: Pro-tip: access the window instead of the global console to bypass the linter! (This is a secret)

```suggestion
    const originalConsoleError = window.console.error;
```

src/components/help-icon/help-icon.test.jsx
cbacon91: I think in this case it doesn't work as well. Since `cleanup` is called `afterEach` test, and `beforeAll` only happens once, only the first test will pass. Subsequent tests will fail because this div has been unmounted by `cleanup`.

There's a case to be made that since we're calling `render` twice there's already some weirdness, but this is definitely an edge case which was caused by the `useLayoutEffect` added

image

juanca (approved):

src/components/tooltip/tooltip.test.jsx
cbacon91: 1d149d67

src/components/tooltip/tooltip.jsx
cbacon91: b46ecc0c

src/components/help-icon/help-icon.test.jsx
juanca: Oh, right, the auto-cleanup. 😓 

I guess the other route is injecting a native DOM node yourself. But then we don't get the auto-cleanup. 

/shrug I'd stick with as-is.

PR #334, opened 2021-07-01
MDS Site restructuring

juanca:
https://www.pivotaltracker.com/story/show/178733149
https://www.pivotaltracker.com/story/show/178700445

## PR upkeep checklist

- [ ] Change log entry N/A
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds_cells/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/25376


README.md
```diff
@@ -2,11 +2,13 @@
 
 A set of [React](https://reactjs.org/) components created by, and for, [Mavenlink](https://www.mavenlink.com/). See all the available components at our [Github Page](https://mavenlink.github.io/design-system/master).
 
+For a living style guide of design patterns, visit our [site on Notion](https://www.notion.so/Mavenlink-Design-System-d3118c3e2f5647d6adfc5b1599af993e).
```
Bloodyaugust, line 5: Nice to get this in there!

src/components/__internal__/abstract-custom-field/abstract-custom-field.jsx
```diff
@@ -5,6 +5,9 @@ import styles from './abstract-custom-field.css';
 import cautionSvg from '../../../svgs/caution.svg';
 import FormControl from '../../form-control/form-control.jsx';
 
+/**
+ * @deprecated
```
Bloodyaugust, line 9: Ooooh, good idea. 👍

src/components/no-options/no-options.jsx
```diff
@@ -2,6 +2,9 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styles from './no-options.css';
 
+/**
+ * @deprecated
```
Bloodyaugust, line 6: Wait, why is this one deprecated? Are we moving away from this component for Multi and Single Select?

styleguide.config.js
```diff
@@ -4,13 +4,14 @@ const webpackConfig = require('./styleguide/webpack.config.js');
 
 module.exports = {
   assetsDir: [
```
Bloodyaugust, line 6: Nice that we got to remove most of our assets 👍

styleguide.config.js
```diff
@@ -25,82 +26,69 @@ module.exports = {
...
+    name: 'Living Codebase',
+    href: 'https://github.com/mavenlink/design-system',
+  }, {
+    name: 'Living Style Guide',
+    href: 'https://www.notion.so/Mavenlink-Design-System-d3118c3e2f5647d6adfc5b1599af993e',
+  }, {
+    name: 'Form Controls',
+    components: [
+      path.join(__dirname, 'src/components/form/form.jsx'),
+      path.join(__dirname, 'src/components/form-control/form-control.jsx'), // Is this private? Should it be undocumented as implementation details?
```
Bloodyaugust, line 45: I think there are definitely use cases for composing this into bigmaven specific components. If a feature team needs a new input that conforms to some specific case, but still wants to maintain accessibility and form compatibility, I think this would be really useful.

styleguide.config.js
```diff
@@ -25,82 +26,69 @@ module.exports = {
...
+      path.join(__dirname, 'src/components/select/select.jsx'),
+      path.join(__dirname, 'src/components/textarea/textarea.jsx'),
+    ],
+  }, {
+    name: 'Widgets',
+    components: [
+      path.join(__dirname, 'src/components/button/button.jsx'),
+      path.join(__dirname, 'src/components/calendar/calendar.jsx'),
+      path.join(__dirname, 'src/components/help-icon/help-icon.jsx'),
+      path.join(__dirname, 'src/components/icon-button/icon-button.jsx'), // Should this be part of a Button interface?
```
Bloodyaugust, line 70: Good question. I think even if we go that route, we'd probably still want a separate component since the DOM structure is so different. Maybe `Button` could just implement a prop that switches to this implementation in that case?

Bloodyaugust (approved):
This is fantastic! I think it's going to be much easier to grok the side panel now. Also like the new `__site__` directory.

src/components/no-options/no-options.jsx
juanca: I think this is leftover from moving code around. I was thinking this was more of an private (but shared) component. At a glance, it's documentation page does not provide that much usefulness / utility. I was thinking of removing the documentation page.

styleguide.config.js
juanca: That's definitely true. My current headspace is: 

1. Do we want to document (with examples) for every extracted component on the MDS site? 
2. A "private component" has more utility when expressed / documented in the actual usage -- for example, `NoOptions` component has little value unless seen / interacted with in the Select component variants. 
3. I think perhaps having the source code is good enough for implementation details. If someone wanted to produce a new form control, they could look at the existing components, study their structure, and produce a new one. Whether it shares code or extracts a new shared code module is up to the developer.

juanca:
I'll revisit these comments in an upcoming story: https://www.pivotaltracker.com/story/show/178789716 I'd like to get the restructure into `master` before it becomes stale(r).

PR #335, opened 2021-07-01
fix(form-control) group label and required together

cbacon91:
## Motivation
fix(form-control): resolve an issue where tooltip text in a form control was making required move to a new line

## Acceptance Criteria
A `FormControl` with just required text should appear as before.
A `FormControl` with required text and a help icon should have them appear with the label and `Required` reminder text left-justified, with only the help icon right-justified, but all on the same line.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-required-and-tooltip-should-be-friends/
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/form-control/form-control.css
+++ b/src/components/form-control/form-control.css
@@ -14,13 +14,7 @@
 }
 
 .required {
-  composes: label;
   margin-left: var(--spacing-small);
-}
-
-.invalid-required {
-  composes: required;
-  color: var(--mds-red-text);
 }
 
 .control-container {
--- a/src/components/form-control/form-control.css
+++ b/src/components/form-control/form-control.css
@@ -4,17 +4,14 @@
 
 .label {
   color: var(--mds-grey-54);
-  display: inline;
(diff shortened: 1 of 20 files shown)


cbacon91:
Before:
image


After:
image

src/components/form-control/form-control.css
```diff
@@ -14,15 +14,9 @@
 }
 
 .required {
-  composes: label;
   margin-left: var(--spacing-small);
```
juanca, line 17: If the label container was `inline-flex` or whatever and we used a `gap` property on it... we wouldn't even need a required classname. 💥

juanca (approved):

src/components/form-control/form-control.css
cbacon91: Yep, you're 100% right! 

c48e1ec5

PR #345, opened 2021-07-14
Refactor CustomFieldInputMultipleChoice

Bloodyaugust:
- refactor(CustomFieldInputMultipleChoice): use `<MultiAutocompleter />` component internally
  - feat(MultiAutocompleter): add `extraParams` API for supporting arbitrary additional query params when loading choices

## Motivation
We want to re-use our MDS components

## Acceptance Criteria
Custom Field Input Multiple Choice API does not change!

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-custom-field-input-multi-choice-refactor-178013333/#/Form%20Controls/CustomFieldInputMultipleChoice
- [x] Green Bigmaven CI check


src/components/multi-autocompleter/multi-autocompleter.jsx
```diff
@@ -93,9 +94,10 @@ MultiAutocompleter.propTypes = {
 
 MultiAutocompleter.defaultProps = {
   className: multiSelectStyles.container,
+  extraParams: '',
   onChange: () => {},
   optionIDGetter: option => option.id,
-  optionLabelGetter: option => option.title || option.name || option.full_name || option.currency,
+  optionLabelGetter: option => option.title || option.name || option.full_name || option.currency || option.label,
```
juanca, line 100: It might be worthwhile to add an inline comment in the prop-type for `optionLabelGetter`.

```suggestion
  /** A function to get the label of an option. Defaults to accessing one of `title`, `name`, `full_name`, `currency`, or `label` keys.  */
  optionLabelGetter: PropTypes.func,```

Comment (juanca) [reply src/components/multi-autocompleter/multi-autocompleter.jsx:100]: I'm not sure if we wanted test coverage on the new `option.label` bit.

File: src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx [hunk 2]
```diff
@@ -213,96 +33,44 @@ const CustomFieldInputMultipleChoice = forwardRef(function CustomFieldInputMulti
...
     };
 
-    getChoices();
+    fetchChoices();
   }, []);
 
+  useEffect(() => {
+    setFullPropValue(props.value.map((choiceID) => {
+      return allChoices.find(choice => choice.id === choiceID);
+    }).filter(mappedChoice => mappedChoice !== undefined));
```
juanca, line 42: When is a mapped choice undefined? This might be indicative of a missing state for the component -- e.g. "if we haven't fully loaded the selected choices then we need to continue to be in a loading state". 

But that could be an iteration step... First, I would like to understand how it's `undefined`.

juanca (approved):

src/components/custom-field-input-multiple-choice/custom-field-input-multiple-choice.jsx
Bloodyaugust: Actually, this was intended to protect against choices saved in something like a custom form that no longer exist on the backend. The loading state is (kind of) already covered by the `MultiAutocompleter` which doesn't currently expose a `loading` prop. The assumption is that loading of an unfiltered set of choices will return more quickly than the filtered set the `MultiAutocompleter` loads, but that may turn out to be faulty.

I think the potential iteration step here is more along the lines of "allow `props.loading` to be set in such a way that the internal loading state is used _after_ the `props.loading` state becomes false".

src/components/multi-autocompleter/multi-autocompleter.jsx
Bloodyaugust: Test added, and also support for `model.label` was backfilled to `Autocompleter`.

PR #346, opened 2021-07-20
MDS Select fires `onChange` after clearing value

Bloodyaugust:
fix(Select): set value to `null` on clear and as default state so `onChange` is called and `dirty` is correct

## Motivation

Feature teams will find this useful if they can detect when the Select is cleared!

## Acceptance Criteria

- MDS Select invokes `onChange` when clearing the component ('x')
  - In the "Ref Example" for select, the `value` column changes to `null` when the selection is cleared
- MDS Select has a ref example in the documentation

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-select-fires-onchange-when-cleared-178953901/#/Form%20Controls/Select
- [x] Green Bigmaven CI check


src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -104,7 +104,7 @@ const CustomFieldInputSingleChoice = forwardRef(function CustomFieldInputSingleC
         readOnly={props.readOnly}
         required={props.required}
         tooltip={props.tooltip}
-        value={value[0] !== -1 ? choices.find(choice => choice.id === value[0]) : undefined}
+        value={value[0] !== -1 ? (choices.find(choice => choice.id === value[0]) || null) : null}
```
Bloodyaugust, line 107: This is probably something we want to come back and look for later. We were (I think unintentionally) hitting cases where `choices.find...` was coming back `undefined`.

src/components/select/select.jsx
```diff
@@ -105,7 +105,7 @@ const Select = forwardRef(function Select(props, ref) {
 
   useImperativeHandle(selfRef, () => ({
     get dirty() {
-      return props.value !== this.value;
+      return props.value === undefined ? value !== null : props.value !== value;
```
Bloodyaugust, line 108: Without this change, `dirty` starts in an incorrect state. It will still look incorrect if `undefined` is intentionally passed as the prop `value`. Not sure how to work around this... collections are so much easier in these cases lol

src/components/select/select.jsx
```diff
@@ -117,6 +117,8 @@ const Select = forwardRef(function Select(props, ref) {
   useEffect(() => {
     if (!mounted.current) return;
 
+    if (props.value === undefined) return;
```
Bloodyaugust, line 120: As above. Without this, our default state becomes undefined, which seems like a breaking of contract, but also means that if `undefined` is intentionally passed for `props.value`, `onChange` will not fire.

src/components/select/select.jsx
```diff
@@ -21,7 +21,7 @@ import useMounted from '../../hooks/use-mounted.js';
 
 const Select = forwardRef(function Select(props, ref) {
   const [showOptions, setShowOptions] = useState(false);
-  const [value, setValue] = useState(props.value);
+  const [value, setValue] = useState(props.value || null);
```
cbacon91, line 24: It might not matter, so feel free to tell me if it doesn't, but remember that an empty string is also falsey in javascript. Is it okay `props.value` changes from an empty string to `null`?

If not we can either use the nullish coalesce (`props.value ?? null`) or just a terenary here

src/components/select/select.jsx
```diff
@@ -47,7 +47,7 @@ const Select = forwardRef(function Select(props, ref) {
   useDropdownClose(wrapperRef, showOptions, handleDropdownClose);
 
   const clear = () => {
-    setValue(undefined);
+    setValue(null);
     setSearchValue(undefined);
```
cbacon91, line 51: Should this also go to null? As a reviewer, it's very weird to me that sometimes we use `undefined` and other times we use `null`

cbacon91 (approved):
it seems fine to me

I feel like there are some details I'm not fully grokking but I don't know that I need to understand it fully

src/components/select/select.jsx
Bloodyaugust: This is just text used internally in the component, so it doesn't actually matter which it is... and I opted to not change it. 😅

src/components/select/select.jsx
Bloodyaugust: Good point, I do think we need to worry about that. I'll update!

juanca (approved):
LGTM! Contingent on updating the imperative handle to avoid serializing `null` as a state in favor of `undefined`. This keeps our imperative handle interface the same so it should be backwards compatible ;)

src/components/select/select.jsx
Bloodyaugust: Fun fact, we can't currently use nullish coalescing in MDS. 🙃

Bloodyaugust:
A note for posterity: we're merging this branch without a current green Bigmaven integration due to ongoing feature work from another team. Bigmaven will be behind the latest release that this makes it into until the integration with this work is clean and green.

PR #353, opened 2021-07-28
Select cell

juanca:
https://www.pivotaltracker.com/story/show/178801453

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds_more_cells/
- [x] Green Bigmaven CI check


juanca:
@cbacon91 While this PR isn't "finished", I think this is a good point for a review. I essentially deprecated AbstractCustomField and that means modifying a bunch of files. 

The next step is to split out a Select control which can be composed into a Select form control and a Select cell control. But that should be a couple of days or so.

src/components/custom-field-input-single-choice/__snapshots__/custom-field-input-single-choice.test.jsx.snap
```diff
@@ -40,7 +40,7 @@ exports[`src/components/custom-field-input-single-choice/custom-field-input-sing
               id="test-id"
               name="field-id"
               role="combobox"
-              style="--numIcon: 1;"
+              style="--numIcon: 3;"
```
juanca, line 43: Instead of a dynamic number, it is hard-coded to 3 -- that is the max number of icons on the select component. We don't want to allow test to underflow into this padded area because a feature team might end up sizing the component too small to fit when all 3 icons are shown.

src/components/select/select.jsx
```diff
@@ -173,48 +173,84 @@ const Select = forwardRef(function Select(props, ref) {
 
   return (
     <div ref={wrapperRef} className={props.className}>
-      <AbstractCustomField
```
cbacon91, line 176: I'm really glad to see any mention of the words `custom field` ripped out of a component that has nothing to do with custom fields

src/components/select/select.jsx
```diff
@@ -135,7 +135,7 @@ const Select = forwardRef(function Select(props, ref) {
   }, [value]);
 
   function handleBlur(event) {
-    if (!(listBoxRef.current && listBoxRef.current.contains(event.relatedTarget))) {
+    if (!(refs.listbox.current && refs.listbox.current.contains(event.relatedTarget))) {
```
cbacon91, line 138: having a `!` outside of a wrapped `&` makes this hard to tell what is actually intended. I know this was already here mostly, but personally I'd take the time to invert the logic slightly:

```suggestion
    if (!refs.listbox.current || !refs.listbox.current.contains(event.relatedTarget))) {
```

cbacon91 (commented):
@juanca was `AbstractCustomField` only used in `Select`, or is that part of the half-complete work

src/components/select/select.jsx
juanca: Gotta get those parentheses matching bruda

src/components/select/select.jsx
```diff
@@ -135,7 +135,7 @@ const Select = forwardRef(function Select(props, ref) {
   }, [value]);
 
   function handleBlur(event) {
-    if (!(listBoxRef.current && listBoxRef.current.contains(event.relatedTarget))) {
+    if (!refs.listbox.current || !refs.listbox.current.contains(event.relatedTarget))) {
```
juanca, line 138:
```suggestion
    if (!refs.listbox.current || !refs.listbox.current.contains(event.relatedTarget)) {
```

juanca:
> @juanca was `AbstractCustomField` only used in `Select`, or is that part of the half-complete work?

Correct! It was the last usage.

src/components/select/select.jsx
cbacon91: psh, never take a suggestion without proofreading!

src/components/select/select.jsx
```diff
@@ -1,220 +1,71 @@
...
-        className={styles['clear-button']}
-        icon={iconClear}
-        label={'Remove selected choice'}
-        onPress={clear}
-      />);
-    }
-
-    return undefined;
+  const refs = {
+    container: useRef(), // ?????
```
cbacon91, line 25: what's the `???` about? Just the fact that a container uses a `ref`?

src/components/control/select.jsx
```diff
@@ -0,0 +1,295 @@
...
+    setSearchValue(defaultValue);
+  };
+  useDropdownClose(wrapperRef, showOptions, handleDropdownClose);
+
+  const refs = {
+    listbox: useRef(),
+    input: useRef(),
+  };
+
+  const [validationMessage, validate] = useValidation(props.validationMessage, refs.input);
```
cbacon91, line 46: I feel like I vaguely recall being told not to use this hook; are we okay with using it here since this is mostly copy/paste refactoring?

src/components/control/select.jsx
```diff
@@ -0,0 +1,295 @@
...
+        }
+      });
+    } else {
+      props.listOptionRefs.forEach((listOptionRef) => {
+        if (listOptionRef.current) {
+          listOptionRef.current.setVisible(true);
+        }
+      });
+    }
+  }, [searchValue, props.listOptionRefs]);
```
cbacon91, line 170: Since this is in a `forEach`, I think we could get rid of the `ifs` entirely and use `filter`. 

There is also some restructuring we could do to this entire `useEffect` to reduce some of the redundancy and nesting. Both the `if` and `else` only care about `props.listOptionRefs` where there is a `current`, so we could hoist that out of the nested logic.

I also made an early `return` for the `displayValueEvaluator` falsiness case to preserve existing logic.

```javascript
useEffect(() => {
  const listOptionRefs = props.listOptionRefs.filter(r => r.current);
  if(searchValue) {
    if(!displayValueEvaluator) return;
    
    listOptionRefs.forEach(listOptionRef => {
      const listOptionDisplayValue = props.displayValueEvaluator(listOptionRef.current.value);
      const matches = listOptionDisplayValue.toLowerCase().includes(searchValue.toLowerCase());
      listOptionRef.current.setVisible(matches);
    }
  } else {
    listOptionRefs.forEach(listOptionRef => listOptionRef.current.setVisible(true));
  }
})
```

Still a bit complex but it has less nesting. As usual this is untested so I might have missed something but I think that's green-green.

src/components/control/select.jsx
```diff
@@ -0,0 +1,295 @@
...
+
+  useEffect(function updateOptionSelected() {
+    props.listOptionRefs.forEach((listOptionRef) => {
+      if (listOptionRef.current) {
```
cbacon91, line 174: as above, would recommend a `filter`

cbacon91 (approved):
lgtm

src/components/select/select.jsx
juanca: Oh, it was my temporary TODO. It's like figure this out. Keep it or remove it. I _did_ figure it out but forgot to remove the comment.

src/components/control/select.jsx
juanca: Yeah, I do want to replace it soon too. Thanks for the reminder

src/components/control/select.jsx
juanca: I'll give it a go. Since we're going for a full round of QA on this component, it might be a good time to do this refactor.

PR #368, opened 2021-08-11
MDS interactive table improvements

### LINKED ISSUE #363: <IconButton readOnly />
The button should be disabled / read-only.

### LINKED ISSUE #364: Cell control improvements
The input cell control needs:

- one-off styling capabilities
- ref interface

The select cell control needs:

- one-off styling capabilities
- ref interface

The autocompleter cell control needs:

- one-off styling capabilities
- ref interface

### LINKED ISSUE #365: Form control labels
The label font size is being overridden by global CSS in bigmaven. 

1. We need to provide a `!important` to our label.

juanca:
## Motivation

As we develop more interactive tables, there's some missing functionality in our table compatible controls.

## Acceptance Criteria

- IconButton component has a disabled API (closes #363)
- Form controls have a MDS font (regardless of bigmaven pollution) (closes #365)
- Cell control improvements: ref API, className API (closes #364)

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds_interactive_table/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/25697


src/components/cell-control/input.test.jsx
```diff
@@ -29,7 +29,15 @@ describe('Input cell control', () => {
   };
 
   it('has defaults', () => {
```
cbacon91, line 31: I think in other components we usually have a `it('supports ref api')` type test

src/components/cell-control/autocompleter.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+  it('has defaults', () => {
+    const ref = createRef();
+    render(<Autocompleter {...requiredProps} ref={ref} />);
+    expect(document.body).toMatchSnapshot();
+    expect(document.body).toBe(document.activeElement);
+    expect(ref.current).toMatchSnapshot();
+  });
+
+  describe('className API', () => {
```
cbacon91, line 40: Do we want a test for the `ref` api? 

And also I know this is a conversation we've beaten to death, but how much of the underlying `autocompleter` do we want to test? Have we decided that since the control itself is tested we're okay with skimping on the cell vs form control version?

src/components/form-control/form-control.css
```diff
@@ -5,7 +5,7 @@
 .label {
   color: var(--mds-grey-54);
   display: inline-flex;
-  font: var(--mds-type-subtext);
+  font: var(--mds-type-subtext) !important;
```
cbacon91, line 8: I feel kind of weird about this. I understand we want to combat pollution, and admittedly if someone really wants a different style they can use their own `!important` and/or not compose this class, I just worry about the arms race that can develop when using `!important`

src/components/icon-button/icon-button.jsx
```diff
@@ -10,7 +10,14 @@ const IconButton = forwardRef(function IconButton(props, ref) {
   const height = parseInt(viewBox[3], 10);
   const tabIndex = props.active ? 0 : -1;
 
+  function onClick(event) {
+    if (props.disabled) return;
+    event.preventDefault();
+    props.onPress(event);
```
cbacon91, line 16: I know this wasn't added here, but I'm wondering why we call refer to `click` as `onPress`? 

I looked it up once before and read something about `press` being more-or-less the same but having some mobile-specific distinctions, but if we're not actually using an `onPress` event then this prop is just misleading

cbacon91 (commented):
Works for me.

Couple questions about `ref` tests and a question I've had for a while about `onPress` but nothing blocking!

src/components/cell-control/autocompleter.test.jsx
juanca: Yeah, I've got another story to revisit the test coverage. For now, I am definitely adding test coverage on the "broken bits" as I integrate with bigmaven. I'm hoping it offers some glimpse at what is reasonable test coverage.

For now, I am just using the fact that the default test has an assertion against the ref API already -- I don't think we need more than that. If I had that originally, I think I wouldn't have had to do this bugfix.

src/components/cell-control/input.test.jsx
juanca: I think it should just be part of the "default tests". I've found that we always want to test 3 things for any component:

- DOM markup
- Initial focus
- Initial ref structure

PR #370, opened 2021-08-13
add pointer-events: none; for disabled buttons

bennapp:
## Motivation
I had to add this css in bigmaven and I thought I shouldn't have to :)

## PR upkeep checklist

- [ ] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   - feat(Input): implement `className` API
   - feat(Select): implement `className` API
   - feat(Autocompleter): implement `className` API
+  - fix(Button): add pointer-events: none; for disabled button
 </details>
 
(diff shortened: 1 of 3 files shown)


juanca:
What's the use-case? I'd figure any pointer events would just not be handled in the JS while a button is disabled.

bennapp:
@juanca I think its better to prevent the js calls rather than force every dev who uses a disabled button to have to handle that case, right?

bennapp:
@juanca Ya, I think I misattributed some bigmaven css pollution to MDS. This was to fix an issue where on hover was highlighting, I realize now that must have been coming from bigmaven.

PR #374, opened 2021-09-01
Webinf duration input

bennapp:
## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check


src/components/duration-input/duration-input.css
```diff
@@ -0,0 +1,47 @@
+@import '../../styles/colors-v2.css';
```
bennapp, line 1: Needs to be consolidated with input

src/components/duration-input/duration-input.jsx
```diff
@@ -0,0 +1,131 @@
...
+    const duration = timeToMinutes(e.target.value);
+    const formattedTime = formatTime(duration);
+    if (formattedTime) {
+      setFormattedValue(formattedTime);
+    }
+    props.onBlur(e);
+  }
+
+  return (
+    <FormControl
```
bennapp, line 52: Should be consolidated with input (?)

src/components/input/input.jsx
```diff
@@ -67,6 +67,8 @@ Input.propTypes = {
   autoFocus: PropTypes.bool,
   className: PropTypes.string,
   cssContainer: PropTypes.string,
+  controlRef: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
bennapp, line 70: I think this is a good use case to pass refs as props since we want to pass two different refs through, but let me know if we have a better pattern for this.

PR #377, opened 2021-09-08
Single choice as an autocompleter

juanca:
https://www.pivotaltracker.com/story/show/179443358

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/autocompleter_single_choice/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/26211

### DIFF
--- a/src/components/control/select.jsx
+++ b/src/components/control/select.jsx
@@ -112,7 +112,7 @@ const Select = forwardRef(function Select(props, ref) {
 
   useImperativeHandle(selfRef, () => ({
     get dirty() {
-      return props.value === undefined ? value !== null : props.value !== value;
+      return props.value !== this.value;
     },
     /** The name of the control element which is used to reference the data after submitting the control. */
     name: props.name,
@@ -132,8 +132,6 @@ const Select = forwardRef(function Select(props, ref) {
 
   useEffect(() => {
     if (!mounted.current) return;
-
-    if (props.value === undefined) return;
 
     setValue(props.value);
   }, [props.value]);
@@ -203,7 +201,7 @@ const Select = forwardRef(function Select(props, ref) {
           required={props.required}
           style={{ '--numIcon': 3 }}
           type="text"
-          value={searchValue || defaultValue}
+          value={searchValue ?? defaultValue}
         />
         <div className={styles['icon-container']}>
           {validationMessage.length > 0 ? (<Icon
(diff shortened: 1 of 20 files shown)


src/components/control/select.jsx
```diff
@@ -112,7 +112,7 @@ const Select = forwardRef(function Select(props, ref) {
 
   useImperativeHandle(selfRef, () => ({
     get dirty() {
-      return props.value === undefined ? value !== null : props.value !== value;
+      return props.value !== this.value;
```
bennapp, line 115: Just curious, how does this work or what is going on here?

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
```diff
@@ -1,116 +1,62 @@
...
-          {item.label}
-        </ListOption>
-      ));
-  };
-
-  useImperativeHandle(selfRef, () => ({
+  useImperativeHandle(ref, () => ({
+    ...autocompleterRef.current,
     get dirty() {
-      const providedValue = props.value[0] !== defaultValue[0] ? props.value[0] : undefined;
```
bennapp, line 49: did this work?

bennapp (approved):
LGTM, much cleaner, nice job!

bennapp:
@juanca just curious, won't we need to do multi-choice as well?

src/components/control/select.jsx
juanca: `this.value` is the result from the `get value()` below. So it's basically what the component thinks it's current state is, like what you would send to the server, i.e. the non-persisted value

`props.value` is what is passed down from "somewhere" -- this somewhere is usually an XHR that is from the server, i.e. the persisted value. 

So we know the component is "dirty" when it does not match up!

src/components/custom-field-input-single-choice/custom-field-input-single-choice.jsx
juanca: It used to work because we had a "reserved default value" of `[-1]` to indicate no choice is selected. We now only need to keep the array empty to indicate that no choice is selected.

juanca:
> @juanca just curious, won't we need to do multi-choice as well?

I though so too but we already did this fix awhile ago! Multi-choice already sends XHRs to the server on user typing. :)

PR #394, opened 2021-10-29
fix(Calendar): set the "year button" type to "button" so we don't submit parent forms

Bloodyaugust:
## Motivation
Choosing a year from the year dropdown in the `Calendar` component submits a parent form, and we don't want that.

## Acceptance Criteria
- Button for year selection is not default type "submit"
- Button for year selection is type "button"

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-mds-datepicker-year-dropdown-179583869
- [x] Green Bigmaven CI check


bennapp (approved):
LGTM

PR #395, opened 2021-11-03
Webinf select bug

bennapp:
## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/control/select.jsx
+++ b/src/components/control/select.jsx
@@ -147,7 +147,7 @@ const Select = forwardRef(function Select(props, ref) {
       validate();
     }
 
-    if (props.value === value) return;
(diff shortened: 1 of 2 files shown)


src/components/control/select.jsx
```diff
@@ -147,7 +147,7 @@ const Select = forwardRef(function Select(props, ref) {
       validate();
     }
 
-    if (props.value === value) return;
+    if (JSON.stringify(props.value) === JSON.stringify(value)) return;
```
bennapp, line 150: This is pretty difficult to test. If we pass in objects as the option values, if this does regular equality, there will be an infinite loop :-/

cbacon91 (approved):

PR #396, opened 2021-11-05
adds new themed tooltip

BenHSmith13:
https://www.pivotaltracker.com/story/show/179806675

https://github.com/mavenlink/mavenlink/pull/26862

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/light-variant-tooltip/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -14,6 +14,7 @@ export default function Tooltip({
   disabled,
   id,
   text,
+  variant,
 }) {
   // size of the triangle in pixels
   const triangleHeight = 8;
@@ -31,7 +32,16 @@ export default function Tooltip({
   // keep the div so enabling/disabling doesn't affect the dom tree or styles
   if (disabled) return <div className={styles.wrapper}>{children}</div>;
 
-  const classNames = [className];
+  let baseClassName;
+  if (className) {
+    baseClassName = className;
+  } else if (variant === 'light') {
+    baseClassName = styles.tooltipLight;
+  } else {
+    baseClassName = styles.tooltip;
+  }
+
+  const classNames = [baseClassName];
   if (direction === 'top') classNames.push(styles.top);
   else if (direction === 'bottom') classNames.push(styles.bottom);
   else if (direction === 'left') classNames.push(styles.left);
@@ -46,7 +56,7 @@ export default function Tooltip({
       onMouseLeave={hide}
     >
       { children }
-      { visible && ReactDOM.createPortal((
+      {visible ? ReactDOM.createPortal((
         <span
           id={id}
           className={classNames.join(' ')}
@@ -58,7 +68,7 @@ export default function Tooltip({
         >
           {text}
         </span>
-      ), document.body)}
+      ), document.body) : null}
     </div>
   );
 }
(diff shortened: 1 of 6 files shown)


src/components/tooltip/tooltip.jsx
```diff
@@ -73,12 +83,15 @@ Tooltip.propTypes = {
   disabled: PropTypes.bool,
   /** The id that the tooltip should be given. The element that this tooltip is describing <strong>MUST</strong> have <code>aria-describedby={id}</code>. */
   id: PropTypes.string.isRequired,
-  /** The text inside of the tooltip. Tooltip text should be concise and kept to 1-2 short sentences. */
-  text: PropTypes.string.isRequired,
+  /** The text and/or graphic inside of the tooltip. Tooltip text should be concise and kept to 1-2 short sentences. */
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+  /** Show the default dark background tooltip or the light background tooltip */
```
cbacon91, line 88: In my opinion, we shouldn't be using nodes in tooltips, it seems like if you want rich content in a tooltip you should use a [popover](https://mavenlink.github.io/design-system/master/#!/Widgets/Popover) instead

BenHSmith13: It's in the design to be a tooltip with non-text content.
Screen Shot 2021-11-05 at 4 06 09 PM

BenHSmith13: Talking to Jason they have plans for other rich content in tooltips. Unordered lists, headers, and what not while still having the tooltip design and functionality.

BenHSmith13: I agree that not just anything should go in there, namely intractable content. I'd love to hear the basis of that opinion, or see some standard that says why that is bad, but images and rich text are in almost every case a better experience to users.

cbacon91: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/tooltip_role

Specifically:

> Because the tooltip itself never receives focus and is not in the tabbing order, a tooltip can not contain interactive elements like links, inputs, or buttons.

I think a slider would fall in that categorization. I think where it gets more opinionated is the discussion of bulleted lists, images, or anything more than 1-2 (short!) sentences of text.

This source is a bit less "official" (just a stackoverflow answer), but I feel that this user basically explains exactly how I feel as well: https://ux.stackexchange.com/a/94552.

> Since tooltips are only meant to tell the purpose of an element they should be short and to the point "Click X to do X" or "User post count"
>
> Popovers, on the other hand, can be much more verbose, they can include a header and many lines of text in the body.

It's opinionated because a lot of people consider a popover to just be a specific type of dialog or modal, where obviously you can use whatever html and focus you want, but others think a popover is a special subtype of tooltip that is "allowed" to have rich content. There isn't a specific aria role for popover, so it's all opinionated at this point.

I like the way this article words it as well (emphasis mine): https://techservicesillinois.github.io/accessibility/examples/popover.html

> A popover is a disclosure pattern (i.e. hide/show widget) **styled to appear as a tooltip**.

A popover is a modal/dialog that _looks_ like a tooltip. It's not a tooltip. It looks like one basically for the design and consistency reasons I think Jason is mentioning. 

> Popover widgets allow for more content than would be appropriate for a tooltip and for the inclusion of interactive elements

I think admittedly one of my fears with allowing `node` in a `tooltip` is that people will start jamming everything in them. It might be an unfounded fear, and I know you're not saying that you're going to add buttons to every tooltip, but keeping the prop type at `string` initially was my way of at least forcing the discussion

BenHSmith13: That's not a slider, it's an image.

BenHSmith13: I agree with every link you posted, Tooltips should not have intractable content, but nothing that you shared said anything about prohibiting rich text, icons, or images. The designs call for a hover over tooltip (and we will continue to see this pattern in the designs), not a popup that requires interaction to open and close.

I understand and agree with preventing intractable content inside a tooltip, but this is not how to do it, not at the cost of rich text and images.

TailorDone (approved):
LGTM!

src/components/tooltip/tooltip.jsx
juanca: I think the image is valid tooltip content. When an accessibility tool reads the contents of the tooltip, it basically reads the label/alt/caption of every nested element. In this case, it will read the text and then the image's alt (unless the image was pure presentational). 

I'm not really sure how to enable developers to drop images in the tooltip but forbid them to add interactive content. But my gut thinks it can be a solved problem so we should probably allow this change in.

juanca (approved):

PR #401, opened 2021-11-18
Interactive cells change value

juanca:
## Motivation

We need interactive cells! We don't really have usages of our Autocompleter in other parts of the app so I'm taking the liberty to do a breaking change and align the `value` prop with the same shape of data as our backend API (i.e. just an ID).

https://www.pivotaltracker.com/story/show/179729086

## Acceptance Criteria

- Autocompleter still works
- Select still works
- CustomFieldSingleChoice still works

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/interactive-cells-change-value/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/autocompleter/autocompleter.test.jsx
+++ b/src/components/autocompleter/autocompleter.test.jsx
@@ -2,6 +2,7 @@ import React, { createRef } from 'react';
 import {
   fireEvent,
   render, screen,
+  waitFor,
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import jestServer from '../../mocks/jest-server.js';
@@ -98,12 +99,10 @@ describe('src/components/autocompleter/autocompleter', () => {
 
   describe('value API', () => {
     it('accepts a value', async () => {
-      const value1 = { id: 22, name: 'cool dude' };
-      const value2 = { id: 68, name: 'coolio dude' };
-      const { rerender } = render(<Autocompleter {...requiredProps} value={value1} />);
-      expect(screen.getByLabelText('Test label')).toHaveValue('cool dude');
-      rerender(<Autocompleter {...requiredProps} value={value2} />);
-      expect(screen.getByLabelText('Test label')).toHaveValue('coolio dude');
+      const { rerender } = render(<Autocompleter {...requiredProps} value="731" />);
+      await waitFor(() => expect(screen.getByLabelText('Test label')).toHaveValue('Another page of models!'));
+      rerender(<Autocompleter {...requiredProps} value="732" />);
+      await waitFor(() => expect(screen.getByLabelText('Test label')).toHaveValue('Do not throw the object!'));
     });
   });
 
@@ -128,9 +127,9 @@ describe('src/components/autocompleter/autocompleter', () => {
   describe('forwardRef API', () => {
     it('can be used to get value as array of selected id', async () => {
       const ref = createRef();
-      render(<Autocompleter {...requiredProps} value={{ name: 'neat', id: 11 }} ref={ref} />);
+      render(<Autocompleter {...requiredProps} value={9} ref={ref} />);
 
-      expect(ref.current.value).toStrictEqual({ name: 'neat', id: 11 });
+      await waitFor(() => expect(ref.current.value).toEqual({ id: '9', label: 'Option 9' }));
     });
   });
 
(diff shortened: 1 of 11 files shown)


src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -146,46 +127,45 @@ describe('src/components/autocompleter/autocompleter', () => {
...
-    it('calls onChange when a new value is selected', () => {
+    it('calls onChange when a new value is selected', async () => {
       let changeValue = '';
       const onChange = (event) => {
         changeValue = event.target.value;
       };
 
-      const models = [{ id: 22, name: 'Foo' }, { id: 33, name: 'Bar' }];
-      render(<Autocompleter {...requiredProps} models={models} onChange={onChange} />);
+      // const models = [{ id: 22, name: 'Foo' }, { id: 33, name: 'Bar' }];
```
juanca, line 136:
```suggestion
```

bennapp (approved):
LGTM

juanca:
Looks like I need to update the imperative handle for an autocompleter to also return the `id` of the selected value. I need to consider what happens when an old value is already in the database. Do we run a backfiller? 🤔

src/components/control/autocompleter.jsx
```diff
@@ -1,28 +1,45 @@
...
 import useFetch from '@bloodyaugust/use-fetch';
 import useMounted from '../../hooks/use-mounted.js';
 import Select from './select.jsx';
 import ListOption from '../list-option/list-option.jsx';
 import { API_ROOT } from '../../mocks/mock-constants.js';
 
 const Autocompleter = forwardRef(function Autocompleter(props, ref) {
-  const { execute } = useFetch();
+  const { execute: executeModels } = useFetch();
+  const { execute: executeValue } = useFetch();
```
cbacon91, line 11: pedantic: `executeModels` sounds like murdering people that get paid to show clothing lines. I'd recommend `fetchModels` personally.

Also, I'm not 100% sure I understand why we need a separate hook for each `fetch` request - that seems super odd to me imo

src/components/control/autocompleter.jsx
```diff
@@ -1,28 +1,45 @@
...
-
-  useEffect(fetchModels, []);
+  useEffect(() => {
+    if (props.value) {
+      executeValue(apiEndpoint('only', props.value))
+        .then((respObj) => {
+          if (mounted.current) {
+            setModel(respObj.json.results.map(result => respObj.json[result.key][result.id])[0]);
+          }
+        })
```
cbacon91, line 28: The entire benefit of `useFetch` is that it has the `isMounted` behavior built-in right? As in the second arg of the promise returned from `executeValue` should be the same as `mounted`?

src/components/control/autocompleter.jsx
juanca: `fetchModels` is already defined in this scope :( I had to switch it. 

I think we double `useFetch` since that's part of its API. I don't recall the details but it needs to generate a function with some variables associated with it. Kinda like state but not really. 🤔

src/components/control/autocompleter.jsx
juanca: It's got a few other nifty things but yeah it does have its own `mounted` var. I can clean this up.

cbacon91 (approved):

src/components/autocompleter/autocompleter.test.jsx
Bloodyaugust: Don't forget to nuke this!

src/components/control/autocompleter.jsx
Bloodyaugust: Most of the state is hidden inside the native `fetch`, but there is some relevant state as well that's part of the hook. Overall I think it makes much more sense for a hook-based fetching API to handle a single request and not multiples. You can re-use the same hook, but basically we want to be able to manage them individually (like cancelling), and the interface for handling multiple gets really messy really fast. Maybe a good topic of discussion for arch meeting?

src/components/control/autocompleter.jsx
Bloodyaugust: The other benefits of `use-fetch` basically come down to moving it into a hook API which feels nicer in react land, abstracting away the need for `useEffect` from your component, automatically transforming the payload to `JSON`, and cancelling the request when the parent component un-mounts.

Bloodyaugust (approved):
LGTM!

PR #406, opened 2021-12-09
Webinf custom form builder fields restyling 180076999

Bloodyaugust:
## Motivation
Fix a bug in `use-tooltip-positioning` and add a new button color type

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-custom-form-builder-fields-restyling-180076999/
- [x] Green Bigmaven CI check

### DIFF
--- a/src/components/tooltip/use-tooltip-positioning.js
+++ b/src/components/tooltip/use-tooltip-positioning.js
@@ -12,7 +12,7 @@ export default function useTooltipPositioning({
   const hide = () => setTarget(null);
 
   useLayoutEffect(() => {
-    if (!target) {
+    if (!target || !tooltipRef.current) {
       setPosition(null);
       return;
     }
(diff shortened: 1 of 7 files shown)


src/components/tooltip/use-tooltip-positioning.js
```diff
@@ -12,7 +12,7 @@ export default function useTooltipPositioning({
   const hide = () => setTarget(null);
 
   useLayoutEffect(() => {
-    if (!target) {
+    if (!target || !tooltipRef.current) {
```
juanca, line 15: Interesting. I wonder why the ref is not set up by now. 🤔 The effect is ran after painting to the DOM so all DOM refs should be present. 

Can you add a bit more information about how you found this bug?

juanca (approved):

src/components/tooltip/use-tooltip-positioning.js
Bloodyaugust: It does work correctly in the initial case, but (and this is how I found it), if you pass `disabled={true}` to the `Tooltip` after it is opened, or otherwise remove it from the DOM while it is being shown, this case shows up.

PR #409, opened 2022-01-04
Update circleci docker image to the supported image

brandon099:
CircleCI has deprecated the image we use in this repo. See this [thread](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034/2).

This PR updates to the officially supported image from CircleCI.


diclophis:
@juanca you might be best suited to determine if there is anything off with the build ... it of course is green, so most of it should be working the same

juanca (approved):

juanca:
Thanks for the PR! This should be good.

PR #420, opened 2022-02-08
refactor(number): extract a number control to be reused with a number…

juanca:
https://www.pivotaltracker.com/story/show/180679137

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/control/number.jsx
+++ b/src/components/control/number.jsx
@@ -1,0 +1,147 @@
+import React, { useEffect, useImperativeHandle, useRef } from 'react';
+import PropTypes from 'prop-types';
+import useValidation from '../../hooks/use-validation.jsx';
+import useMounted from '../../hooks/use-mounted.js';
+import Icon from '../icon/icon.jsx';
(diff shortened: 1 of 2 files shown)


src/components/control/icons.css
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/spacing.css';
+
+.container {
+  align-items: center;
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: var(--spacing-small);
+  position: absolute;
+  right: var(--spacing-medium);
+  transform: translateY(-50%);
```
bennapp, line 10: Just curious, what is the transform for?

juanca: This `.icons` class is the "container". The container is absolutely position, and its top-border is placed in the "vertical midway point" of the parent container. 

image

This transform then "tweaks" the icons container by pushing it up by 50% of its height. This means the icons container is now vertically centered as an absolutely positioned element. 

image

PR #422, opened 2022-02-11
MDS Date cell

juanca:
https://www.pivotaltracker.com/story/show/180679135

## Note for reviewer(s)

I ended up upgrading a bunch of testing libraries to have a better label on the calendar widget. Unfortunately, it didn't work because there is a [bug with something inside testing-library](https://github.com/testing-library/dom-testing-library/issues/1103). For now, I changed the label on the buttons to be generic. 

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/mds-date-cell/
- [x] Green Bigmaven CI check: https://github.com/mavenlink/mavenlink/pull/27963

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "eslint-config-mavenlint-react": "^2.0.0",
     "eslint-plugin-cypress": "^2.2.0",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^27.0.6",
+    "jest": "^27",
     "msw": "^0.26.0",
     "postcss-loader": "^4.0.4",
     "postcss-preset-env": "^6.7.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@testing-library/jest-dom": "^5",
     "@testing-library/react": "^12",
     "@testing-library/react-hooks": "^7",
-    "@testing-library/user-event": "^13.1.8",
+    "@testing-library/user-event": "^13",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^27.0.6",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@testing-library/jest-dom": "^5",
     "@testing-library/react": "^12",
-    "@testing-library/react-hooks": "^3.4.1",
+    "@testing-library/react-hooks": "^3",
     "@testing-library/user-event": "^13.1.8",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@testing-library/jest-dom": "^5",
     "@testing-library/react": "^12",
-    "@testing-library/react-hooks": "^4",
+    "@testing-library/react-hooks": "^5",
     "@testing-library/user-event": "^13.1.8",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@testing-library/jest-dom": "^5",
     "@testing-library/react": "^12",
-    "@testing-library/react-hooks": "^6",
+    "@testing-library/react-hooks": "^7",
     "@testing-library/user-event": "^13.1.8",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-react": "^7.0.0",
     "@testing-library/jest-dom": "^5",
     "@testing-library/react": "^12",
-    "@testing-library/react-hooks": "^3",
+    "@testing-library/react-hooks": "^4",
     "@testing-library/user-event": "^13.1.8",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
     "@testing-library/jest-dom": "^5",
-    "@testing-library/react": "^10",
+    "@testing-library/react": "^11",
     "@testing-library/react-hooks": "^3.4.1",
     "@testing-library/user-event": "^13.1.8",
     "babel-core": "^7.0.0-bridge.0",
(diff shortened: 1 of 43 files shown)


src/components/autocompleter/autocompleter.test.jsx
```diff
@@ -45,7 +45,7 @@ describe('src/components/autocompleter/autocompleter', () => {
   describe('validationMessage', () => {
     it('sets the input to be invalid', () => {
       render(<Autocompleter {...requiredProps} validationMessage="not valid" />);
-      expect(screen.getByLabelText('Test label')).toHaveDescription('not valid');
+      expect(screen.getByLabelText('Test label')).toHaveAccessibleDescription('not valid');
```
juanca, line 48: This is needed for the latest testing-library libs. I did a global find and replace.

src/components/cell-control/autocompleter.md
```diff
@@ -6,22 +6,22 @@ import { Table, TableHeader, TableHeaderCell, TableBody, TableRow, TableCell } f
 
 const ids = {
   th: {
-    autocompleter: uuid.v4(),
+    example: uuid.v4(),
```
juanca, line 9: I fixed a bunch of bad IDs in my recent examples. I made it more copy-pastable for now. Ideally, we should have code that _verifies_ the existence of labels. But that's another problem to solve in the future.

src/components/custom-field-input-date/__snapshots__/custom-field-input-date.test.jsx.snap
```diff
@@ -4,58 +4,54 @@ exports[`src/components/custom-field-input-date/custom-field-input-date classNam
 <body>
   <div>
     <div
-      class="container"
```
juanca, line 7: In order to extract a "date control" to be shared between the field and cell components, I "simplified" the structure. Just moved some `<div>`s inside and moved the `<FormControl>` to the outside. We lost a little bit of padding but I think that's okay given the easy code re-use.

src/components/control/date.jsx
```diff
@@ -0,0 +1,240 @@
...
+    event.preventDefault();
+    setActive(true);
+    setEditing(true);
+    setExpanded(true);
+  }
+
+  function onInputKeyDown(event) {
+    setActive(true);
+
+    if (props.readOnly) return;
```
Bloodyaugust, line 86: Should we be able to `setActive` if the input is `readOnly`?

src/components/control/date.jsx
```diff
@@ -0,0 +1,240 @@
...
+      target: ref.current,
+      detail: {
+        validationMessage,
+      },
+    });
+  }, [validationMessage]);
+
+  useLayoutEffect(() => {
+    refs.input.current.value = editing ? toFullDateFormat(value) : toDateStringFormat(value);
+  }, [value]);
```
Bloodyaugust, line 141: Should this dep array also include `editing`? 🤔

Bloodyaugust (approved):
LGTM! There are some non-blocking questions and potential refactors.

package.json
```diff
@@ -34,10 +34,10 @@
     "@babel/plugin-proposal-class-properties": "^7.2.3",
     "@babel/preset-env": "^7.2.3",
     "@babel/preset-react": "^7.0.0",
-    "@testing-library/jest-dom": "^5.11.1",
-    "@testing-library/react": "^10.4.7",
-    "@testing-library/react-hooks": "^3.4.1",
-    "@testing-library/user-event": "^13.1.8",
+    "@testing-library/jest-dom": "^5",
+    "@testing-library/react": "^12",
```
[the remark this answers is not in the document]
juanca: Hmmmm.... I don't like the extra work. Theoretically, both suites of testes are independent of each other. I would prefer less work 😅  and or less friction for updating libs.

src/components/calendar/calendar.css
```diff
@@ -13,7 +13,6 @@
   padding-left: var(--spacing-medium);
   padding-right: var(--spacing-medium);
   width: calc((var(--spacing-x-large) * 7) + calc(var(--spacing-medium) * 2));
-  height: 293px;
```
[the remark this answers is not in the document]
juanca: I'm actually confused why this was hard-coded. It looks 👨‍🍳 👌  without it. 🤔

src/components/control/date.jsx
```diff
@@ -0,0 +1,240 @@
...
+import calendarSvg from '../../svgs/calendar.svg';
+import Calendar from '../calendar/calendar.jsx';
+import IconButton from '../icon-button/icon-button.jsx';
+import Icons from './icons.jsx';
+import useDropdownClose from '../../hooks/use-dropdown-close.js';
+import styles from './date.css';
+import useForwardedRef from '../../hooks/use-forwarded-ref.js';
+
+function toDateStringFormat(date) {
+  return date ? date.toLocaleDateString(undefined, { month: 'short', year: 'numeric', day: 'numeric' }) : '';
```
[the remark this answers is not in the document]
juanca: That probably makes sense but I don't know of a real use-case to let it be a prop. I would rather keep it good enough. Let me look into it.

src/components/control/date.jsx
```diff
@@ -0,0 +1,240 @@
...
+}
+
+function toFullDateFormat(date) {
+  return date ? date.toISOString().slice(0, 10) : undefined;
```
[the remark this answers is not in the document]
juanca: Nope! Pre-existing code and I think it is working?

src/components/control/date.jsx
juanca: 🤔 I'm not sure how `active` is used here but a read-only input can still be focusable which I think implies it is active?

src/components/control/date.jsx
juanca: I don't think it matters here:

- If the value changes then we re-format and update the DOM reference
- If editing changes, we are already re-formatting the in-memory value and passing it down as a `value` or `defaultValue` prop IIRC

src/components/control/date.jsx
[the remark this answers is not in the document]
juanca: `undefined` results in browser defaults for now. I think that's pretty good until we store the user's preference in our data model.

package.json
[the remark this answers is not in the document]
Bloodyaugust: Fair point! They are pretty well separated. And I'm all for less work! 😉

src/components/calendar/calendar.css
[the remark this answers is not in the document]
Bloodyaugust: I bet it was some edge-case that we'll run into again now that it's gone. 😉

src/components/control/date.jsx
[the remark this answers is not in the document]
Bloodyaugust: Ah, I was under the impression that either project or account did store that, but maybe it's just timezone? 🤔

src/components/control/date.jsx
Bloodyaugust: Fair enough. We can always tweak it later if needed.

src/components/control/date.jsx
Bloodyaugust: Cool, sounds good!

PR #434, opened 2022-03-09
feat(autocompleter control): Add singleModelParam to Autocompleter control, use for single model fetching

Bloodyaugust:
## Motivation
The `Currency` controller in BigMaven (and possibly others) does not use a classic BrainstemPresenter, and does not support the `only` `queryParam`. This means that `Autocompleter`, when provided with `params.value`, does not correctly return only one `Currency`. The correct `queryParam` for that endpoint is `matching`, so we need to support an override here for that. The default value for this param is `only`, to maintain existing functionality without need for code changes.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/webinf-currency-autocompleter-single-match-180877221/
- [ ] Green Bigmaven CI check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
 Prepend new items to make git merging easier.
+- feat(autocompleter control): Add singleModelParam to Autocompleter control, use for single model fetching
 
 ## 0.99.2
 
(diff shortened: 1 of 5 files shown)


Bloodyaugust:
Closing, as we ended up fixing this issue with changes to the API instead. This could still be a thing we want in the future, but 🤷

PR #435, opened 2022-03-15
Webinf form update changes

bennapp:
## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Deployment URL: https://mavenlink.github.io/design-system/webinf-form-update-changes/
- [ ] Green Bigmaven CI check


src/components/control/multi-autocompleter.jsx
```diff
@@ -13,10 +13,17 @@ const MultiAutocompleter = forwardRef(function MultiAutocompleter(props, ref) {
   const { execute: fetchSelectedChoices } = useFetch();
   const [loading, setLoading] = useState(true);
   const [options, setOptions] = useState([]);
-  const [value, setValue] = useState(props.value);
+  const mounted = useRef(false);
+
+  function arrayWrap(v) {
+    return Object.prototype.toString.call(v) !== '[object Array]' ? [v] : v;
```
juanca, line 19: I think this is more standard but I haven't looked it up in a hot minute. 

```suggestion
    return Array.isArray(v) ? v : [v];
```

src/components/multi-autocompleter/multi-autocompleter.test.jsx
```diff
@@ -257,6 +257,34 @@ describe('<MultiAutocompleter>', () => {
...
+      rerender(<MultiAutocompleter
+        {...requiredProps}
+        value={['55']}
+      />);
+
+      const foo = await findSelectedOption('test label', 'Foo');
+      expect(foo).toBeInTheDocument();
+    });
+
+    it('accepts a single id', async () => {
```
juanca, line 275: I'm not sure if we want to keep this. A multi-completer should always accept a list of things since it is suppose to select multiple things. 

What is the bigmaven use-case for this kind of API?

src/components/control/multi-autocompleter.jsx
```diff
@@ -64,12 +76,27 @@ const MultiAutocompleter = forwardRef(function MultiAutocompleter(props, ref) {
...
     }
 
-    if (props.value.length) fetchPropsValue();
-  }, [props.value.map(props.optionIDGetter).join(',')]);
+    if (value.length) {
+      fetchPropsValue();
+    } else {
+      setValueForSelect([]);
+    }
+  }, [listOfIdsString(value), options]);
```
juanca, line 84: Why does this re-fetch the selected options when it fetches the autocompleter dropdown choices?

src/components/multi-autocompleter/multi-autocompleter.test.jsx
bennapp: Yeah. Why not keep it?

src/components/control/multi-autocompleter.jsx
bennapp: If value changes, so say a new array of ids, then it needs to re-fetch with the only query including those models in order to set the correct data structure.

juanca: Given that statement, this effect is only dependent on the `value` changing:

```suggestion
  }, [listOfIdsString(value)]);
```

src/components/control/multi-autocompleter.jsx
```diff
@@ -64,12 +76,27 @@ const MultiAutocompleter = forwardRef(function MultiAutocompleter(props, ref) {
...
+
+  function safePropsValue(v) {
+    if (Object.prototype.toString.call(v) !== '[object Array]') {
```
juanca, line 87:
```suggestion
    if (!Array.isArray(v)) {
```

src/components/control/multi-autocompleter.jsx
```diff
@@ -13,10 +13,17 @@ const MultiAutocompleter = forwardRef(function MultiAutocompleter(props, ref) {
...
   const [options, setOptions] = useState([]);
-  const [value, setValue] = useState(props.value);
+  const mounted = useRef(false);
+
+  function arrayWrap(v) {
+    return Array.isArray(v) ? v : [v];
+  }
+
+  const [value, setValue] = useState(arrayWrap(props.value));
+  const [valueForSelect, setValueForSelect] = useState(value);
```
juanca, line 23: From my understanding between `value` and `valueForSelect`, this initial state might not be entirely correct.

`value` is always an array of objects or strings. `valueForSelect` is always an array of objects 
 -- which is populated after fetching the objects from the API. 

Does it make more sense to ensure `valueForSelect` is a list of objects in the initial state? Otherwise I think we're violating the prop-types for `value` on the MultiSelect component. Something like: 

```suggestion
  const [valueForSelect, setValueForSelect] = useState(value.map(v => typeof v === 'string' ? { id, label: '' } : v));
```

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Prepend new items to make git merging easier.
 
 ## 0.100.3
 - fix(multi-select): don't set invalid when blurring to click an option from the dropdown
+- feat(multi-select): allow setting value with types: single id as string or array of ids
```
juanca, line 7: I don't see any changes to `MultiSelect` in this PR. Is this more accurate? 

```suggestion
- feat(multi-autocompleter): allow setting value with types: single id as string or array of ids
```

src/components/multi-autocompleter/multi-autocompleter.test.jsx
juanca: It's a bit easier to reason about when we know the `value` prop is always an array (and preferably an array of _something_). 

It also seems like `*_ids` fields should always be arrays. What is an example network payload where it returns a single id vs a list of ids? I bet there might be a bug in the API responses (or API parsing).

bennapp: `approver_id` is one example

juanca (changes requested):
At least 1 improvement to be made. Other pending questions / discussions.

bennapp:
What pending improvement? @juanca

src/components/control/multi-autocompleter.jsx
```diff
@@ -123,15 +139,19 @@ MultiAutocompleter.propTypes = {
   tooltip: PropTypes.string,
   validationMessage: PropTypes.string,
   /** value is an array of objects matching the shape of options */
-  value: PropTypes.arrayOf(PropTypes.object),
+  value: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.object),
+    PropTypes.arrayOf(PropTypes.string),
+    PropTypes.string,
```
juanca, line 145:
```suggestion
```

src/components/multi-autocompleter/multi-autocompleter.jsx
```diff
@@ -102,13 +102,17 @@ MultiAutocompleter.propTypes = {
   tooltip: PropTypes.string,
   validationMessage: PropTypes.string,
   /** value is an array of objects matching the shape of options */
-  value: PropTypes.arrayOf(PropTypes.object),
+  value: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.object),
+    PropTypes.arrayOf(PropTypes.string),
+    PropTypes.string,
```
juanca, line 108:
```suggestion
```

juanca (approved):

PR #438, opened 2022-03-21
v0.100.2

Bloodyaugust:
## 0.100.2
- fix(multi-select): don't set invalid when blurring to click an option from the dropdown


bennapp (approved):
LGTM

PR #446, opened 2022-04-18
fix(MultiSelect): clicking within the component focuses the autocomplete input

juanca:
## Motivation

QA observed that you can click around the icons and it would not "focus" the component. After clicking away, client-side validations were not running.

## Acceptance Criteria

- On a required multi-select (or variants), you can click around the icons and click away 
- Client-side validations show up as appropriate (e.g. required)

## PR upkeep checklist

- [x] Change log entry
- [x] Deployment URL: https://mavenlink.github.io/design-system/fix-focus-on-multi-autocompleter/
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/control/multi-select.jsx
+++ b/src/components/control/multi-select.jsx
@@ -176,6 +176,7 @@ const MultiSelect = forwardRef(function MultiSelect(props, ref) {
 
   function onClick(event) {
     if (event.defaultPrevented) return;
+    if (event.target !== refs.autocomplete.current) refs.autocomplete.current.focus();
     if (props.readOnly) return;
 
     setExpanded(true);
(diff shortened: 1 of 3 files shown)


src/components/control/multi-select.jsx
```diff
@@ -176,6 +176,7 @@ const MultiSelect = forwardRef(function MultiSelect(props, ref) {
 
   function onClick(event) {
     if (event.defaultPrevented) return;
+    if (event.target !== refs.autocomplete.current) refs.autocomplete.current.focus();
```
juanca, line 179: I can't test this because you have to "click" the background div (which is not `screen.getBy`able). I am leaving this up to QA to verify this behavior.

Bloodyaugust: We could give it a `testID` or test it in Cypress, but tbh I think it's fine without an automated test.

Bloodyaugust (approved):
LGTM!

src/components/control/multi-select.jsx
juanca: Yeah, I'm leaning towards only interacting with accessible element in tests. But that's just my "all-in" mentality. 😬

PR #447, opened 2022-04-26
- fix(Select): clearing items triggers props.onChange

bennapp:
## Motivation
There was a minor bug where clearing a value wouldn't trigger onChange. This is because we do this check in the useEffect; `if (JSON.stringify(props.value) === JSON.stringify(value)) return;`.

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 Major/Minor/Patch: This is an example changelog item, usually can be the commit message.
 Prepend new items to make git merging easier.
+- fix(Select): clearing items triggers props.onChange
 
 ## 0.101.1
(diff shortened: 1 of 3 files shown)


juanca (approved):
This solutions looks good. We might be able to get rid of the value checker in the use effect.

```
if (JSON.stringify(props.value) === JSON.stringify(value)) return;
```

I _think_ we no longer need this because the `useEffect` runs only on component updates. The `value` should only update via user interactions. 

But that can be a refactor if we find problems with the existing solution -- as long as we have test coverage it is fine.

PR #452, opened 2022-05-05
Enhance Date Cell controls

JustinRyanH:
## Motivation

Developers should know when the user starts and stop interacting with the Date Control.

For example, if the developer needs to hold off making an HTTP request until the Date is complete when the 
end user manually types in the date into the `input type=date` cell.

### Notable Changes

- Activate the `Date Control` whenever the user tabs into the Date control component. 
- When the user reverse tabs into the `Date Control` the focus directly goes into the Input first.
- `IconButton` will now forward `onBlur` and `onFocus` (needed for the updated accessibility controls with `Date Control`)

**BEFORE**

https://user-images.githubusercontent.com/571433/166849163-d15ce5a4-9d56-41d3-84dd-70e6691fcb82.mp4

**AFTER**

https://user-images.githubusercontent.com/571433/166849034-06f9d927-27f8-4b5b-8fdb-bf46044118e5.mp4

## PR upkeep checklist

- [x] Change log entry
- [x] Label(s)
- [x] Assignee(s)
- [x] Deployment URL: https://mavenlink.github.io/design-system/enhance-cell-data-controls/
- [x] Green Bigmaven CI check https://github.com/mavenlink/mavenlink/pull/29249


src/components/control/date.jsx
```diff
@@ -63,13 +72,19 @@ const Date = forwardRef(function Date(props, forwardedRef) {
 
     validate();
     setActive(false);
-    if (!refs.input.current.validationMessage) setEditing(false);
+    if (!props.validationMessage) setEditing(false);
```
juanca, line 75: I need to check this once it is deployed. Date control should stay in edit mode when an entered date is invalid and the user leaves the control.

src/components/control/date.jsx
```diff
@@ -165,6 +190,7 @@ const Date = forwardRef(function Date(props, forwardedRef) {
         name={props.name}
         onChange={onInputChange}
         onClick={onInputClick}
+        onFocus={onFocusIntoDateControl}
```
juanca, line 193: I wonder if we can just hoist the `onFocus` handler to the root container since we really care about _any_ focus within.

src/components/control/date.jsx
```diff
@@ -119,6 +134,15 @@ const Date = forwardRef(function Date(props, forwardedRef) {
     if (active && props.onChange) props.onChange({ target: ref.current });
   }, [value]);
 
+  useEffect(() => {
+    if (!previousActive && active) {
```
juanca, line 138: Hmmm... would using a dependency on the active state work as well? I would prefer to stick to React conventions.

```
useEffect(() => {
  if (active) {
    props.onActive(value);
  } else { 
    props.onDeactive(value);
  }
}, [active]);

File: src/components/control/date.jsx [hunk 4]
```diff
@@ -155,6 +179,7 @@ const Date = forwardRef(function Date(props, forwardedRef) {
       style={{ height: '100%', position: 'relative' }}
     >
       <input
+        aria-labelledby={ids.label}
```
juanca, line 182: AFAIK, the `<label>`'s `htmlFor` attribute points to the `id` here. 

But, does this improve table cell accessibility? We might have been missing some better testing patterns and/or QA patterns in regards to a11y.

src/components/control/date.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+    it('toggles the date to active when user tabs into Date Control', () => {
+      const onActivate = jest.fn();
+      render(<>
+        <button aria-label="pre input button" />
+        <label id="test-id-label">Date</label>
+        <Date id="test-id" onActivate={onActivate} />
+      </>);
+
+      // User has Focused on Element Before Date Input
```
juanca, line 39: IIRC, I _think_ the "first" element focused on mount is the document body. You could just tab into the date.

src/components/control/date.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+      user.click(screen.getByRole('button', { name: 'pre input button' }));
+
+      //
+      // User Tabs to the Date Input
+      user.tab();
+      expect(screen.getByLabelText('Date')).toHaveFocus();
+      expect(onActivate).toHaveBeenCalledWith(undefined);
+    });
+
+    it('activates the date cell when user tabs out of the Input', () => {
```
juanca, line 49:
```suggestion
    it('activates the date cell when user tabs back into the Input', () => {
```

src/components/control/date.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+      // User Tabs to the Date Input
+      user.tab({ shift: true });
+      expect(screen.getByLabelText('Date')).toHaveFocus();
+      expect(onActivate).toHaveBeenCalled();
+    });
+  });
+
+  describe('onDeactivate API', () => {
+    it('toggles date to active when user clicks on date control', () => {
```
juanca, line 69:
```suggestion
    it('toggles active when user clicks away from the date control', () => {
```

juanca (approved):
Looking good. Non-blocking suggestions and questions. I would like to dev QA the component when it deploys to github pages

src/components/control/date.jsx
JustinRyanH: This was a previous bug, we can check into it. With the `refs.input.current.validationMessage` we actually got `null` as a String

src/components/control/date.jsx
JustinRyanH: Yeah pulling this up to the container div works. I'm going to do that

src/components/control/date.jsx
JustinRyanH: I trashed this

src/components/control/date.jsx
JustinRyanH: the problem with this implementation is is we immediately call `props.onDeactive ` when we render the component for the first time. Which I mean it is inactive, but it's not the same as changing from active to inactive, and from inactive to active. 

One option is we could make a hook like this. https://thewebdev.info/2021/03/13/how-to-make-the-react-useeffect-hook-not-run-on-initial-render/ and then we would have the cleaner version of the `useEffect`

JustinRyanH:
@juanca Each of my `fix: Spec` commit has a pretty lengthy description. Let me know if any of them are confusing

src/components/control/date.test.jsx
JustinRyanH: Yeah. I wanted to walk through the accessibility story here. Especially with Keyboard focused users.

src/components/control/date.jsx
juanca: ah, okay, that works. we also have a `useMountedEffect` which only runs the effect after it has been mounted. looks like both achieve the same thing.

src/components/control/date.jsx
```diff
@@ -188,7 +188,7 @@ const Date = forwardRef(function Date(props, forwardedRef) {
         min={props.min}
         name={props.name}
         onChange={onInputChange}
-        onClick={onInputClick}
+        onMouseDown={onInputClick}
```
juanca, line 193: I will to verify this. Mouse down could fire but the mouse up could occur on a different element (think click+hold). This could put the component in a weird state but it might not matter.

juanca:
@JustinRyanH Do you have an associated bigmaven PR? We need to verify the bigmaven test suite is green with these changes. If so, we can make a release in this PR and bump the bigmaven PR to the latest release.

PR #468, opened 2022-06-29
feat(Popover): implement className api

bennapp:
feat(Popover): implement hideHeading api

## Motivation
We want a fly over menu, which is really just a popover but with no X to close, no padding and no title. I felt it was better to compose existing components than create flyover menu, which is really just pop-over in combination with list options.

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/popover/popover.jsx
+++ b/src/components/popover/popover.jsx
@@ -28,7 +28,7 @@ const Popover = forwardRef(function Popover(props, ref) {
 
   useLayoutEffect(() => {
     if (open) {
-      closeIconRef.current.focus({ preventScroll: true });
+      closeIconRef.current?.focus({ preventScroll: true });
     }
   }, [open]);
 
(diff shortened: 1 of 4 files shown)


src/components/popover/popover.jsx
```diff
@@ -53,46 +53,55 @@ const Popover = forwardRef(function Popover(props, ref) {
 
   return (
     <section
-      aria-labelledby="popover-heading"
-      className={styles.container}
+      aria-labelledby={props.hideHeading ? undefined : 'popover-heading'}
```
juanca, line 56: If we're hiding the heading, we will want an `aria-label` for accessibility. i.e. what is the fly out menu's purpose? e.g. "Settings for Notifications page", "Actions on resource Foo row".

juanca (approved):
What a coincidence, I need something just like this for a Notifications story.

PR #471, opened 2022-07-21
Autocompleter render children

bennapp:
## Motivation
We need the Autocompleter to render children

## Acceptance Criteria
The AutoCompleter can render children

## PR upkeep checklist

- [ ] Change log entry
- [ ] Label(s)
- [ ] Assignee(s)
- [ ] Green Bigmaven CI check

### DIFF
--- a/src/components/control/autocompleter.jsx
+++ b/src/components/control/autocompleter.jsx
@@ -79,16 +79,6 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
     },
   }));
 
-  // function renderSelectItems(onSelect) {
-  //   return () => {
-  //     return models.map((modelInfo, index) => (
-  //       <ListOption key={modelInfo.id} onSelect={onSelect} ref={listOptionRefs[index]} value={modelInfo}>
-  //         { props.children ? props.children({ modelInfo }) : props.displayValueEvaluator(modelInfo) }
-  //       </ListOption>
-  //     ));
-  //   };
-  // }
-
   return (
     <Select
       classNames={props.classNames}
@@ -106,6 +96,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
       readOnly={props.readOnly}
       ref={selectRef}
       required={props.required}
+      specialChildrenRender={!!props.children}
       tooltip={props.tooltip}
       validationMessage={props.validationMessage}
       value={model}
@@ -114,7 +105,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
       {({ onSelect }) => (
         models.map((modelInfo, index) => (
           <ListOption key={modelInfo.id} onSelect={onSelect} ref={listOptionRefs[index]} value={modelInfo}>
-            { props.children({ modelInfo }) ?? props.displayValueEvaluator(modelInfo) }
+            { props.children ? props.children(modelInfo) : props.displayValueEvaluator(modelInfo) }
           </ListOption>
         ))
       )}
@@ -157,7 +148,7 @@ Autocompleter.propTypes = {
 
 Autocompleter.defaultProps = {
   apiEndpoint: undefined,
-  children: () => {},
+  children: undefined,
   classNames: {},
   displayValueEvaluator: displayName,
   label: undefined,
--- a/src/components/control/autocompleter.jsx
+++ b/src/components/control/autocompleter.jsx
@@ -10,7 +10,7 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
   const { execute: executeValue } = useFetch();
   const [models, setModels] = useState([]);
   const [searchModels, setSearchModels] = useState([]);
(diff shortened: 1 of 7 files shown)


src/components/control/autocompleter.jsx
```diff
@@ -41,7 +43,12 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
     executeModels(apiEndpoint(props.searchParam, searchString))
       .then(({ json, mounted }) => {
         if (mounted) {
-          setModels(json.results.map(result => json[result.key][result.id]));
+          const jsonResults = json.results.map(result => json[result.key][result.id]);
```
juanca, line 46: What is the difference between search models and models? They seem to both represent API data.

src/components/control/autocompleter.jsx
```diff
@@ -88,23 +102,24 @@ const Autocompleter = forwardRef(function Autocompleter(props, ref) {
...
       validationMessage={props.validationMessage}
       value={model}
       wrapperRef={props.wrapperRef}
     >
       {({ onSelect }) => (
-        models.map((modelInfo, index) => (
+        renderModels.map((modelInfo, index) => (
           <ListOption key={modelInfo.id} onSelect={onSelect} ref={listOptionRefs[index]} value={modelInfo}>
-            {props.displayValueEvaluator(modelInfo)}
+            { props.children ? props.children(modelInfo) : props.displayValueEvaluator(modelInfo) }
```
juanca, line 122: Is it preferable to use `displayValueEvaluator`?

src/components/control/select.jsx
```diff
@@ -195,6 +196,30 @@ const Select = forwardRef(function Select(props, ref) {
       .forEach(_ref => _ref.current.setSelected(JSON.stringify(value) === JSON.stringify(_ref.current.value)));
   }, [value, showOptions]);
 
+  useEffect(() => {
```
juanca, line 199: If you are manipulating the DOM (e.g. `focus`) then you _need_ to use `useLayoutEffect`

src/components/control/select.jsx
```diff
@@ -195,6 +196,30 @@ const Select = forwardRef(function Select(props, ref) {
...
+      refs.input.current.focus();
+    }
+
+    setBeenCleared(false);
+  }, [hasBeenCleared]);
+
+  const currentValue = searchValue ?? defaultValue ?? '';
+  const shouldRenderSpecialChild = props.specialChildrenRender && defaultValue;
+
+  function renderSpecialChild() {
```
juanca, line 210: What is the purpose of a special child? I'd like to understand the use-case. I'm concerned we're creating a one-off autocompleter.

src/components/control/select.jsx
```diff
@@ -309,6 +338,7 @@ Select.defaultProps = {
   children: () => {},
   classNames: {},
   displayValueEvaluator: value => value,
+  inputHeight: undefined,
```
bennapp, line 341: remove this

src/components/control/autocompleter.jsx
bennapp: This is to fix a bug where the dropdown list would only include previous search results. We need to manage two lists so that you can open the dropdown when you are searching vs. not searching and see the correct set.

src/components/control/autocompleter.jsx
bennapp: What is nice about this is that your children will be rendered when you select an option.

src/components/control/select.jsx
bennapp: I've had issues with `useLayoutEffect` but i'll give that a shot.

src/components/control/select.jsx
bennapp: It allows you to render a child as a selected component.

bennapp: So it will determine based on keys what is selected and render that instead of the input. Like this:
Screen Shot 2022-07-26 at 10 10 17 AM

bennapp: It's fairly generic, you just need id as keys in your children and values to be of the shape `{ id: <id> ... }`.

src/components/control/autocompleter.jsx
juanca: This seems like an indirect way to fix the bug. Usually, as a React pattern, the component will perform effects after state changes. I would prefer to see a solution that would re-fetch the correct data after the user is done searching. Keeping two lists just seems like an improper modeling of the data fetching. 

If the logic is simplified, I would prefer to do that change in this PR. Otherwise, let's keep the two lists.

src/components/control/autocompleter.jsx
juanca: Oh, I see. Just brainstorming here: Looks like we might be able to default `children` as just text value / `<input />` since it seems like a base-case. But let's consider that in a distant future when we understand more use-cases of custom rendering in autocompleters.

juanca (approved):
This looks cool!

